PR #1, opened 2021-12-06
Update 00_introduction.ipynb 

WANGSSSSSSS:
I feel like you wrote it wrong by mistake, optimization variables are optimized by forward pass, while auxiliary variables are fixed during forward pass and then optimized by pytorch.


luisenp:
Hi @WANGSSSSSSS. Thanks for the PR :) We actually meant to say that, when running a learning loop that includes a TheseusLayer, updating variables (of any kind) before running the inner loop optimization is best done via `forward()`, rather than using `variable.update()` directly. We will reword this to avoid confusion.

PR #21, opened 2021-12-09
update continuous integration

maurimo:
This diff is updating the continuous integration in preparation for cublas-based sparse LU solver.

Ubuntu image is updated to latest.
Cuda 10 and 11 are installed and used as test environments.


.circleci/config.yml
```diff
@@ -5,43 +5,119 @@ version: 2.1
...
+      working_directory: ~/project
+      command: |
+        pip install --progress-bar off -e .
+
+run_tests: &run_tests
+  - run:
+      name: Running tests
+      working_directory: ~/project
+      command: |
+        pytest -s theseus/tests/test_theseus_layer.py
```
mhmukadam, line 119: Shouldn't we be running all tests `pytest -s theseus`?

.circleci/config.yml
```diff
@@ -74,36 +150,35 @@ jobs:
...
 
 workflows:
   version: 2
   build:
     jobs:
       - py37_linux
       - py38_linux
-      - unittests_gpu17
+      - unittests_gpu17_cuda10
+      - unittests_gpu17_cuda11
```
mhmukadam, line 184: Missing new line.

mhmukadam (approved):
Left some minor comments. We can merge after resolving them.

.circleci/config.yml
mhmukadam: Yeah, I figured. We can merge once Luis confirms.

luisenp: Yes, we did this since the full test suite takes some time, and IMO there is not need for all these tests to be run both on CPU and GPU. That being said, we should probably run GPU tests on a large suit of unit tests, but we don't have to decide this in this PR. What's important for this PR is that the functionality provided here is also tested in CI, which I'm not sure it's the case.

maurimo: wait, this PR is only about the CI changes, and of course there are not new tests since there is no new functionality. The new tests are added in subsequent PRs, while the new functionalities are added incrementally.

luisenp: OK, that's good, I missed this, sorry. As I mentioned offline, I only responded to comments, haven't yet looked through the PRs again :)

mhmukadam: Makes sense. I'll flag this on the code coverage issue so we can tackle it there.

@maurimo, I think in your other PRs (if you haven't already) you will have to add a line 'pytest -s theseus/extlib/tests/' and maybe others to capture places where we do need to test GPU functionality.

PR #22, opened 2021-12-09
cusolver based batched LU solver

maurimo:
This PR adds as batched sparse cublas-based LU solver.


.circleci/config.yml
```diff
@@ -5,43 +5,120 @@ version: 2.1
 # Executors
 # -------------------------------------------------------------------------------------
 executors:
-  gpu:
+  gpu_cuda10:
```
mhmukadam, line 8: Maybe we should rebase on mau.update_ci so you don't have to repeat changes to this file?

maurimo: this branch is already based on top of `mau.update_ci`, but here it's showing the diff to `main` and that's unfortunate... it doesn't make sense to have this being an independent branch as the changes here require the ci, do you think there is a better way? should I target `mau.update_ci` as merge target? this doesn't look ideal either.

mhmukadam: Hmm, if there isn't an easy way to set this up with git then it makes sense to leave as is. Maybe once the other PR is merged this gets resolved.

maurimo: makes sense to me - so since now on the target with a set of subsequent diffs (as I have here) is setting as target the parent branch?

maurimo: Done, now it show as only change in the CI the new GPU tests added to the list of GPU tests.

luisenp: I still see diff for changes that were on the other brach and already merged. Did GitHub get confused here?

maurimo: probably! It wasn't showing those changes when I changed the target.
Probably it understood that the new target for merge is main (after update_ci was merged), but it's showing the diff to the old main! Bug report to GitHub?

maurimo: merged the new "main" into this branch, now it's showing the correct diff.

luisenp (approved):
LGTM! Only left a few minor comments and a question on the test code.

.circleci/config.yml
maurimo: Not sure what the correct behavior of github should be, should it show the diff merged-branch-with-main to current main, of current branch to common ancestor, where the branch branched? probabaly github is doing the right thing after all.

luisenp: I think it should show the diff with whatever the current state of main is, otherwise things can get really confusing. In the past when this happens, I change the target to some other branch, then back to main, and it solves it.

maurimo: so it cannot do a plain diff do main, or main could have included a ton of unrelated stuff from other sides. the automatic merge didn't work (there was a small conflict, no idea of why, considering that main was not identical to the previous target branch!), so it's justfied that it showed branch-to-branch-point instead of merged-branch-to-main diff.

mhmukadam: > I change the target to some other branch, then back to main, and it solves it
This has worked for me too.

theseus/extlib/test_cusolver_lu_solver.py
```diff
@@ -0,0 +1,99 @@
+import numpy as np
+import pytest  # noqa: F401
+import torch  # needed for import of Torch C++ extensions to work
+from scipy.sparse import csr_matrix
+
+from theseus.extlib.sparse_matrix_utils import generate_sparse_matrix
+
+
+def check_lu_solver(base_batch_size, batch_size, rows, cols, fill, verbose=False):
```
mhmukadam, line 9: We should have this file inside `extlib/tests/` dir following convention and for later as this module grows.

theseus/extlib/test_cusolver_lu_solver.py
```diff
@@ -0,0 +1,99 @@
+import numpy as np
+import pytest  # noqa: F401
+import torch  # needed for import of Torch C++ extensions to work
+from scipy.sparse import csr_matrix
+
+from theseus.extlib.sparse_matrix_utils import generate_sparse_matrix
```
mhmukadam, line 6: We should add most of these in the root init file so they are available with `import theseus as th`. Maybe you already do this in the next two PRs. We follow this import convention in examples/tutorials as well as tests.

theseus/extlib/cusolver_lu_solver.cpp
```diff
@@ -0,0 +1,334 @@
+
```
mhmukadam, line 1: Please add the license lines at the start of this and all new files created in this PR.
```
# Copyright (c) Meta Platforms, Inc. and affiliates.
#
```

theseus/extlib/test_cusolver_lu_solver.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def check_lu_solver(base_batch_size, batch_size, rows, cols, fill, verbose=False):
+    # this is necessary assumption, so that the hessian is full rank
+    assert rows > cols
```
mhmukadam, line 11: These could be equal too and would be full rank?
`assert rows >= cols`

theseus/extlib/test_cusolver_lu_solver.py
```diff
@@ -0,0 +1,99 @@
...
+    check_lu_solver(base_batch_size=5, batch_size=5, rows=150, cols=60, fill=0.1)
+
+
+def test_lu_solver_6():
+    check_lu_solver(base_batch_size=10, batch_size=10, rows=300, cols=90, fill=0.1)
+
+
+# not yet, the reason is most probably a cublas bug?!
+# def test_lu_solver_7():
+#     check_lu_solver(base_batch_size=10, batch_size=5, rows=150, cols=60, fill=0.2)
```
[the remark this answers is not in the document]
mhmukadam: Also, what's `base_batch_size`? Maybe add a comment explaining it.

theseus/extlib/sparse_matrix_utils.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+from scipy.sparse import lil_matrix
+
+
+def generate_sparse_matrix(rows, cols, fill, min_entries_per_col):
```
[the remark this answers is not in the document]
mhmukadam: +1, I got confused too, until I saw this was doing fill in with 1.0.

Also, should we move this stuff to the primary utils dir since it doesn't depends on the cpp code?

theseus/extlib/cusolver_lu_solver.cpp
```diff
@@ -0,0 +1,334 @@
...
+                                   torch::TensorOptions(torch::kDouble).device(A_rowPtr.device()));
+
+    CUSOLVER_CHECK(cusolverRfBatchSolve(cusolverRfH,
+                                        P.data_ptr<int>(), Q.data_ptr<int>(),
+                                        1, // nrhs
+                                        temp.data_ptr<double>(), rows,
+                                        (double**)b_array.data_ptr<uint8_t>(), rows));
+}
+
+PYBIND11_MODULE(cusolver_lu_solver, m) {
```
mhmukadam, line 299: Over time we will likely add more things to extlib. Does it make sense to keep the pybind stuff in a separate file?

theseus/extlib/cusolver_lu_solver.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+    P = P_cpu.cuda();
+    Q = Q_cpu.cuda();
+}
+
+std::vector<int> CusolverLUSolver::factor(const torch::Tensor& A_val) {
+
+    TORCH_CHECK(A_val.device().is_cuda());
+    TORCH_CHECK(A_val.dim() == 2);
+    TORCH_CHECK(A_val.size(0) <= batchSize); // currently failing unless "==", cublas bug?!
```
mhmukadam, line 241: Weird. Maybe a work around can be to check both <= and >= ?

mhmukadam (commented):
This is excellent! Left some comments.

theseus/extlib/test_cusolver_lu_solver.py
```diff
@@ -0,0 +1,99 @@
...
+def check_lu_solver(base_batch_size, batch_size, rows, cols, fill, verbose=False):
+    # this is necessary assumption, so that the hessian is full rank
+    assert rows > cols
+
+    if not torch.cuda.is_available():
+        return
+    from theseus.extlib.cusolver_lu_solver import CusolverLUSolver
+
+    A_skel = generate_sparse_matrix(rows, cols, fill, min_entries_per_col=3)
+    A_cols = cols
```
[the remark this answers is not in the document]
mhmukadam: +1

theseus/extlib/cusolver_lu_solver.cpp
maurimo: oh yeah will update the check - the comment is that it's sad that cublas forces the exact batch size.

theseus/extlib/test_cusolver_lu_solver.py
maurimo: true - well that's just a test, but still.

theseus/extlib/test_cusolver_lu_solver.py
maurimo: doing so in next rev.

theseus/extlib/sparse_matrix_utils.py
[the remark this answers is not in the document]
maurimo: moving. well the generated object is a matrix, with just ones as values, maybe it's better `generate_mock_sparse_matrix`?

theseus/extlib/test_cusolver_lu_solver.py
maurimo: Sorry I think I don't understand this comment! So, the point of adding things in the __init__ is just to avoid one more step right? so we can call `th.generate_sparse_matrix`. However most of these functions (test utils and C++ extensions) look quite internal and meant to be exported to me, right? The solver class is very low level, essentially the smallest thing that can be exported to Python.

theseus/extlib/cusolver_lu_solver.cpp
maurimo: so what I would like to say is that we would like to check "<=", but cuda functions fail unless "==" holds. Probably we should check "==" anyway, untill an improved cuda is released.

theseus/utils/sparse_matrix_utils.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import numpy as np
+from scipy.sparse import csr_matrix, lil_matrix
+
+
+def generate_mock_sparse_matrix(rows, cols, fill, min_entries_per_col) -> csr_matrix:
```
luisenp, line 10: How about `generate_random_binary_matrix`? This module is more general than unit tests, so having "mock" generating routines here seems somewhat out of place.

theseus/extlib/test_cusolver_lu_solver.py
mhmukadam: Yes, so we can call `th.something` and the user doesn't have to remember which function is available in which submodule scope.

Ultimately, it depends if any of these functions would be useful to a user directly? If not we can leave that subset as is. Also if any of them are in fact 'internal' then maybe we should append a `_` in front of them.

Any thoughts, @luisenp?

theseus/extlib/sparse_matrix_utils.py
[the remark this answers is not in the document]
luisenp: In another comment I proposed `generate_random_binary_matrix` which seems more descriptive and more appropriate for a generic sparse matrix utility function, intended to be used beyond unit tests (given the python module it's located in).

theseus/utils/sparse_matrix_utils.py
mhmukadam: `generate_random_sparse_binary_matrix`?

theseus/extlib/test_cusolver_lu_solver.py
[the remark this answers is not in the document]
maurimo: ideall we would like to have a setup base batch size, and a current batch size which should be possible at least to be irregular and smaller. however nothing works because of cublas, so those two have to identical, period.

theseus/utils/sparse_matrix_utils.py
luisenp: Hehe, yes, I was just realizing that this proposal leaves out the sparse property and came to comment on it. This name is a bit too long, though.

maurimo: that's good for me!

luisenp: Maybe `rand_sparse_binary_matrix`?

theseus/extlib/test_cusolver_lu_solver.py
maurimo: so, I would consider those functions to be hardly useful for the user. Of course the user may happen to be doing some low-level manipulation with batches sparse matrices and find those functions incidentally useful, but I would see adding those to the exported functionality more as a distraction than as something useful. Just my opinion!

theseus/utils/sparse_matrix_utils.py
mhmukadam: You could drop random since 'generate' implicitly implies that.

theseus/extlib/test_cusolver_lu_solver.py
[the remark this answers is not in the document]
mhmukadam: I see, is the base size referring to the size of the first symbolic structure you create?

theseus/extlib/test_cusolver_lu_solver.py
luisenp: The matrix generation routine seems OK to have at `th` level, but for the rest I don't think so. From what I can tell, we still need to wrap the CUDA solver into a subclass of `th.optimizer.linear.LinearSolver`, and that's the one that should be exponsed at the `th` level. @maurimo can you confirm?

theseus/extlib/test_cusolver_lu_solver.py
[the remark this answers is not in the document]
mhmukadam: Awesome, 'init_base_size' is much more clear without extra explanation needed.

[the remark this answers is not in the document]
luisenp: Got it. So, basically once we expose the sparse solver to Theseus optimizers, we will need to document that the batch size has to remain constant, and specified at construction time? Is this correct?

theseus/extlib/test_cusolver_lu_solver.py
maurimo: oh sure the subclass of LinearSolver (PR24) is perfectly fine to be exported, I was referring to what is in extlib, which is too low-level to be exported directly.

theseus/extlib/test_cusolver_lu_solver.py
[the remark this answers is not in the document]
maurimo: exactly!

theseus/extlib/cusolver_lu_solver.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+        CUSOLVER_CHECK(cusolverSpXcsrluAnalysisHost(cusolverSpH, rows, nnz,
+                                                    A_descr, pB_rowPtr_cpu, pB_colInd_cpu, 
+                                                    info));
+
+        torch::Tensor B_val_cpu = torch::zeros(nnz, torch::TensorOptions(torch::kDouble));
+        double *pB_val_cpu = B_val_cpu.data_ptr<double>();
+        // make our model B invertible
+        for(int r = 0; r < rows; r++) {
+            for(int i = pB_rowPtr_cpu[r]; i < pB_rowPtr_cpu[r+1]; i++) {
```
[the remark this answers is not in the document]
maurimo: yeah, but well, this is not going to be any bottleneck anyway, and a clever compiler may optimize it anyway!

theseus/extlib/test_cusolver_lu_solver.py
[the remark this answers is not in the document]
maurimo: yeah makes sense!

theseus/extlib/test_cusolver_lu_solver.py
mhmukadam: Sounds good, then let's leave as is. If we see a need later pull any of them up we can revisit.

theseus/__init__.py
```diff
@@ -43,5 +43,6 @@
     NonlinearOptimizerStatus,
 )
 from .theseus_layer import TheseusLayer
+from .utils import random_sparse_binary_matrix
```
mhmukadam, line 46: Given the discussion, just having this function in utils/__init__.py is okay.

mhmukadam (approved):

PR #23, opened 2021-12-09
CUDA batch matrix multiplication and ops

maurimo:
This PR adds batched CUDA-based matrix operations, on sparse matrices having the same structure:

Mt*M, M*v, Mt*v and diagonal damping. These are the operations needed to build the linear system to be solved from the Jacobian and to apply the forward/backward steps.


maurimo:
setting parent branch as merge target, to show the incremental diff.

theseus/extlib/mat_mult.cu
```diff
@@ -0,0 +1,393 @@
...
+	// matrices are in CSR format:
+	//   rowPtr determines begin/end of row data,
+	//   colInd determines the column index
+	int srcRow_offset = M_rowPtr[row];
+	int srcRow_len = M_rowPtr[row+1] - srcRow_offset;
+	const int* srcRow_colInd = M_colInd + srcRow_offset;
+	const double* srcRow_val = Ms_val + batchIndex * M_nnz + srcRow_offset;
+	double* MtMs_batch_val = MtMs_val + batchIndex * MtM_nnz;
+	for(int i = 0; i < srcRow_len; i++) {
+		//printf("i=%d\n", i);
```
luisenp, line 57: Lingering print.

luisenp (approved):
LGTM. Only one very minor comment.

theseus/extlib/mat_mult.cu
maurimo: gone!

theseus/extlib/test_mat_mult.py
```diff
@@ -0,0 +1,119 @@
+import pytest  # noqa: F401
+import torch  # needed for import of Torch C++ extensions to work
+from scipy.sparse import csr_matrix
+
+from theseus.extlib.sparse_matrix_utils import generate_sparse_matrix
+
+
+def check_mat_mult(batch_size, rows, cols, fill, verbose=False):
```
mhmukadam, line 8: Same as pervious PR about moving tests to a separate tests dir.

theseus/extlib/test_mat_mult.py
```diff
@@ -0,0 +1,119 @@
...
+    if not torch.cuda.is_available():
+        return
+    from theseus.extlib.mat_mult import apply_damping, mat_vec, mult_MtM, tmat_vec
```
mhmukadam, line 11: Same as pervious PR, adding import stuff to root init file.

theseus/extlib/mat_mult.cu
```diff
@@ -0,0 +1,392 @@
+
```
mhmukadam, line 1: Add license lines to all new files.

mhmukadam (approved):
Looks good! Just some minor comments similar to previous PR.

theseus/extlib/test_mat_mult.py
maurimo: done!

theseus/extlib/mat_mult.cu
maurimo: done!

theseus/extlib/test_mat_mult.py
maurimo: are we sure we want those on the root init file? those are really low level operations (and not putting them in the root file does not prevent an experienced user from importing them, they are just less advertised and the user should be aware of what she's doing)

mhmukadam: Makes sense. Do you or @luisenp think it would be worth pulling any of these to root init? We can always do this in the future after we see a constant need for any of them.

mhmukadam: Given the discussion in PR 22, we will leave as is for now.

luisenp: Agree with @maurimo here. Another possibility down the line is to have some sort of SparseMatrix class and have these be overridden operators.

mhmukadam (approved):

PR #24, opened 2021-12-09
CUDA-based solver class and autograd function

maurimo:
This PR adds solver and autograd function based on CUDA solver and operations.

Since CUDA "refactor" library has the limitation of sharing an opaque context for the symbolic decomposition and the numerical factor data, a number of context is used in order to allow (a limited amount of) optimization loop unrolling. If a context was overwritten when calling backward step an exception is raised.


maurimo:
Set parent branch as target.
Added damping parameters, now the solver supports the same damping types as the dense solver (ellipsoidal or not), the autograd is tested using "large" damping parameters (a correction in the autograd function may actually be overkill, and needed only in the test with very large values).

theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -0,0 +1,107 @@
...
+        ellipsoidal_damping: bool = True,
+        damping_eps: float = 1e-8,
+        **kwargs,
+    ) -> torch.Tensor:
+        if not isinstance(self.linearization, SparseLinearization):
+            raise RuntimeError(
+                "CholmodSparseSolver only works with theseus.optimizer.SparseLinearization."
+            )
+
+        self._last_solver_context = (
```
luisenp, line 85: Would it be safer to completely error out instead of doing FIFO? If I understand this correctly, if this solver is used more than `_num_solver_contexts`, the backward pass will start giving incorrect results. Or maybe this is already caught [here](https://github.com/facebookresearch/theseus/pull/24/files#diff-69f2100fbaed4dc21b13b12f5990f5b34eeb26adfec87de4a45385a21b0aa061R110)? Even if that's the case, is there a reason we would like to keep things mod `_num_solver_context` in this function?

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
```diff
@@ -0,0 +1,67 @@
...
+    for i in range(12):
+        start = max(0, i - 2)
+        end = min(i + 1, 10)
+        col_ind += all_cols[start:end]
+        row_ptr.append(len(col_ind))
+    data = torch.randn((batch_size, len(col_ind)), dtype=torch.double)
+    return 12, 10, data, col_ind, row_ptr
+
+
+def test_sparse_solver():
```
luisenp, line 27: Are we testing unrolling operations? Even w/o the implicit differentiation checks, we could add some basic test for this werein we do a series of `solve()` operations and check some tensor's `.grad` content numerically?

luisenp (commented):
Looks great! Left a couple of questions/comments for discussion.

theseus/__init__.py
```diff
@@ -45,3 +45,8 @@
 from .theseus_layer import TheseusLayer
 
 import theseus.embodied as eb
+
+import torch
```
mhmukadam, line 49: maybe just do `from torch.cuda import is_available`?

theseus/__init__.py
```diff
@@ -45,3 +45,8 @@
...
+
+if torch.cuda.is_available():
```
mhmukadam, line 51: Should we have an else here that raises an error/warning, if the user tries to call/use `LUCudaSparseSolver` without having cuda.

theseus/optimizer/autograd/__init__.py
```diff
@@ -8,3 +8,10 @@
 __all__ = [
     "CholmodSolveFunction",
 ]
+
+import torch
+
+if torch.cuda.is_available():
+    from .lu_cuda_sparse_autograd import LUCudaSolveFunction
```
mhmukadam, line 15: Same comment about torch import and use without cuda.

theseus/optimizer/autograd/tests/test_lu_cuda_sparse_backward.py
```diff
@@ -0,0 +1,69 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest  # noqa: F401
+import torch
+from torch.autograd import gradcheck
+
+import theseus.core as theseus
+import theseus.optimizer as thoptim
+import theseus.optimizer.linear as thlin
```
mhmukadam, line 12: Following convention from all other test files we should replace with `import theseus as th`.

theseus/optimizer/linear/__init__.py
```diff
@@ -3,7 +3,12 @@
...
 # LICENSE file in the root directory of this source tree.
 
+import torch
+
 from .dense_solver import CholeskyDenseSolver, DenseSolver, LUDenseSolver
 from .linear_optimizer import LinearOptimizer
 from .linear_solver import LinearSolver
 from .sparse_solver import CholmodSparseSolver
+
+if torch.cuda.is_available():
```
mhmukadam, line 13: Flagging here too so we remember to make it consistent with what we end up doing in other places.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -0,0 +1,107 @@
...
+
+from theseus.core import Objective
+from theseus.extlib.cusolver_lu_solver import CusolverLUSolver
+from theseus.optimizer import Linearization, SparseLinearization
+from theseus.optimizer.autograd import LUCudaSolveFunction
+
+from .linear_solver import LinearSolver
+
+
+class LUCudaSparseSolver(LinearSolver):
```
mhmukadam, line 18: We should move this to `linear/sparse_solver.py` so change that file to be Cholmod sparse solver since that's the only one in there currently. Same for the test file.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+
+        super().__init__(objective, linearization_cls, linearization_kwargs, **kwargs)
+        self.linearization: SparseLinearization = self.linearization
+
+        if self.linearization.structure().num_rows:
+            self.reset()
+
+        # the `damping` has the purpose of (optionally) improving conditioning
+        # self._damping: float = damping
```
mhmukadam, line 43: Is LM not supported at the moment?

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
```diff
@@ -0,0 +1,148 @@
...
+    def backward(ctx, grad_output):
+
+        # HACK: check if the context has been reused (and overwritten)
+        if ctx.factor_id is not None and ctx.factor_id != ctx.solver_context.factor_id:
+            raise RuntimeError(
+                "Factoring context was overwritten! Increase the number of contexts"
+            )
+
+        batch_size = grad_output.shape[0]
+        targs = {"dtype": grad_output.dtype, "device": "cuda"}  # grad_output.device}
```
mhmukadam, line 115: We don't have to do it in this PR, but aside from a few lines like this one, we might be able to pull in a lot of common code between LU sparse and Cholmod sparse.

This could also work for the solver class itself by having a class hierarchy
`LinearSolver` -> `SparseSolver` ->  `CholmodSparseSolver` and `LUCudaSparseSolver`

cc @luisenp

mhmukadam (commented):
Looks great! Only some minor comments and one comment about whether we should consolidate some of the contents in the sparse solvers with a common sparse solver parent class.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
mhmukadam: +1

theseus/__init__.py
luisenp: If cuda is not available, then `LUCudaSparseSolver` would not even be visible, so trying to do `import th.LUCudaSparseSolver` would throw an error. Maybe this will be confusing for users?

Is it possible to make all of these imports permanent in the relevant `__init__.py` files, but address the `cuda.is_available()` issue inside the corresponding `.py` files? Basically, allow users to import `LUCudaSparseSolver` and only throw an error if this class is attempted to be instantiated w/o cuda being available.

mhmukadam: Throwing an error seems okay but your other suggestion is also good. Not sure if there is a standard convention here that we can follow.

luisenp: Right, but the error will be something like:

```python
import th
solver = th.LUCudaSparseSolver(...)
...
...
AttributeError: module 'theseus' has no attribute 'LUCudaSparseSolver'
```

which seems misleading, no?

luisenp: What we want is for this to be a `RuntimeError` that says `"class `LUCudaSparseSolver` requires cuda to be available"`. This could be achieved, for example, by adding the check inside the class' constructor, or, if that's not possible due to some C++ import issues, making a stub class with no requirements whose job is only to throw an error, and that gets imported in place of the original one when no cuda is present.

mhmukadam: Yes, we would want an error message that indicates this functionality is only available with cuda. Whichever version get us there easily sounds good.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
maurimo: so, yes, there is where the error is caught. The mod-_num_solver_contexts logic is not here but in the Sparse solver, that instead of giving this class a sparse representation is providing this class a "new" context (new = least recently used, "mod" num contexts). Here (where you linked) it's just necessary to check that the context we have been provided has not been reused in the mean time, can do so comparing the "factor_id". If you didn't notice where that is done I think it was indeed because this fact was quite non-intrusive in the overall logic ;)

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: that can certainly be done!

theseus/__init__.py
maurimo: ok but at the level of the python class right? so no problems creating a mock python class which just throws error then I think? This may confuse python typing unless the mock class is mimicking all the methods with the same types.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
maurimo: I would prefer keeping files separate, it looks cleaner (even more so if the cuda class will have a mock version when cuda is not available). Anyway all is re-exported in the `__init__` file so it makes no difference to the user right?

theseus/optimizer/linear/lu_cuda_sparse_solver.py
maurimo: dead code, the parameter is set in solve, will be cleaned.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
maurimo: Talking about autograd function or the solver itself? the solver classes look to small to find any relevant common code... the autograd function I would say yes, even if the way to merge some functionality will certainly require some thoughts.

theseus/__init__.py
luisenp: Oh, but that's a good point about typing and the mock class. Is there a reason why the error cannot be thrown inside the original class' constructor? Can you elaborate a bit?

maurimo: so, the problem is that if cuda is not available I cannot import any C++ extension depending on cuda either (or an error occurs). Therefore the "import" has to be done inside the functions themselves. Do you see any issue with that?

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
mhmukadam: Autograd function does have a lot of shared code, but I think it's worth considering the solver too.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
mhmukadam: That sounds fine too. Then, after we decide what to do with the class hierarchy (from the other comment below) for the solver we should either rename sparse solver to cholmod sparse or make a new cholmod sparse with only the parent class in sparse solver to make things consistent.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
```diff
@@ -5,15 +5,26 @@
...
-from theseus.extlib.cusolver_lu_solver import CusolverLUSolver
-from theseus.extlib.mat_mult import apply_damping, mat_vec, mult_MtM, tmat_vec
-
 from ..linear_system import SparseStructure
 
 
 class LUCudaSolveFunction(torch.autograd.Function):
     @staticmethod
     def forward(ctx, *args, **kwargs):
+        if not torch.cuda.is_available():
```
luisenp, line 14: Any reason we cannot move all of these conditional imports to the top of the file?

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
```diff
@@ -105,6 +116,19 @@ def forward(ctx, *args, **kwargs):
     @staticmethod
     def backward(ctx, grad_output):
 
+        if not torch.cuda.is_available():
```
luisenp, line 119: Same question about moving to the top of the file.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -21,11 +20,12 @@ def __init__(
         objective: Objective,
         linearization_cls: Optional[Type[Linearization]] = None,
         linearization_kwargs: Optional[Dict[str, Any]] = None,
-        damping: float = 1e-6,
-        batch_size: int = 16,
         num_solver_contexts=1,
         **kwargs,
     ):
+        if not torch.cuda.is_available():
```
luisenp, line 26: Same question about moving to top of the file.

theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -39,12 +39,22 @@ def __init__(
         if self.linearization.structure().num_rows:
             self.reset()
 
-        # the `damping` has the purpose of (optionally) improving conditioning
-        # self._damping: float = damping
-
         self._num_solver_contexts: int = num_solver_contexts
 
     def reset(self, batch_size: int = 16):
+        if not torch.cuda.is_available():
```
luisenp, line 45: Same question about moving to top of the file.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
```diff
@@ -60,8 +56,108 @@ def test_sparse_solver():
...
+    ]
+    c = torch.randn((batch_size, num_cols), dtype=torch.double).cuda()
+
+    def batched_dot(a, b):
+        return torch.sum(a * b, dim=1)
+
+    def iterate_solver(As, bs):
+        accum = None
+        for A, b in zip(As, bs):
+            # print("it1...", linearization.b.shape)
```
luisenp, line 110: dead code?

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
```diff
@@ -6,9 +6,7 @@
 import pytest  # noqa: F401
```
luisenp, line 6: I tried to run this in my computer, before I installed the CUDA extensions, and got an error as shown below. Do we want to make the tests in this file conditional on the extension being installed? We could use [pytest.importskip](https://docs.pytest.org/en/6.2.x/reference.html#pytest-importorskip) for this. 

image

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
```diff
@@ -60,8 +56,108 @@ def test_sparse_solver():
...
+    # if insufficient contexts, assert exception is raised
+    if test_exception:
+        with pytest.raises(RuntimeError):
+            result.backward(torch.ones_like(result))
+        return
+
+    # otherwise, compute and check gradient
+    result.backward(torch.ones_like(result))
+
+    epsilon = 1e-4
```
luisenp, line 132: I was a bit worried about the high error tolerance, but changing epsilon to `1e-7` and both of the tolerances to `1e-4` works for me. Maybe better?

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
```diff
@@ -60,8 +56,108 @@ def test_sparse_solver():
...
+    num_checks = 10
+    for i in range(num_checks):
+        for perturb_A in [False, True]:
+            for step in range(num_steps):
+                perturbed_As = [A.detach().clone() for A in As]
+                perturbed_bs = [b.detach().clone() for b in bs]
+
+                if perturb_A:
```
luisenp, line 140: It would be useful to add some brief explanatory comments here in case someone else needs to look at this test in the future. It seems that this test checks against a directional derivative (in the direction of `perturb`), is this correct? Beforehand, I was looking for something like a partial derivative wrt to each of A's dimensions. Took me a bit to figure out what this test was doing.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
```diff
@@ -60,8 +56,108 @@ def test_sparse_solver():
...
+    bs = [
+        torch.randn((batch_size, num_rows), dtype=torch.double).cuda()
+        for _ in range(num_steps)
+    ]
+    c = torch.randn((batch_size, num_cols), dtype=torch.double).cuda()
+
+    def batched_dot(a, b):
+        return torch.sum(a * b, dim=1)
+
+    def iterate_solver(As, bs):
```
luisenp, line 107: Would be helpful to add some comments with a high level summary of what this does. If I understand correctly, this returns

```
y = dot(z,c)
z = sum(A_i \ b_i)
```

luisenp (changes requested):
Looks great! Left a bunch of minor comments, but they should all be quick to fix. After that merging is OK with me.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
maurimo: Well if so we are back to having to implement a mock-class for when cuda is not available...

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: removing in next rev!

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: well that's precisely what the ci currently does, they are only tested when the extension is installed.
However I would not make them run only if the extension is *available*, but rather on where the user tried to install with Cuda. In such a case they should fail and notify the user that something didn't work as expected and the extension she tried to compile is actually not available.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: makes sense, will change those in next rev, thanks for trying!
Anyway the check is not trivial with current values (eg as could be because if values verified were very close to zero), I had verified that too dumping the content of the derivatives.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: makes sense! I thought the fastest way to get a test working with with a random directional check (on several directions being sampled).

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: will add comments yeah

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
luisenp: I'm mostly wondering about any overheads from repeated `import` calls. I know python caches the module for future imports, but there might some additional non-negligible cost for calling `import`? (not sure). We could get around writing a mock class by doing the conditional import at the top while also setting a boolean flag, say `_EXTLIB_WAS_LOADED`, and (inside the class) throw the runtime error based on the status of the flag (rather than  checking the import).

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
luisenp: Just to be clear, I think the test makes a lot of sense, just mentioned the alternative to explain why it took me a bit to figure this out, and thus why a comment in the code would be helpful.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
maurimo: I see - yeah that would work. I would go for the cleaner solution, just not sure which is (ideas @mhmukadam ?) if there is an overhead doing import it will be certainly very small, and I feel we should not optimize around something that should already be optimized inside python.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: So, that's right that running tests should not fail because theseus was correctly installed with no cuda support!
Not sure how that can be achieved, I was just saying that the test should just not skip because if was unable to load the extension. Ideally there should be a command running tests that will skip cuda tests if the extension is not installed, or two separate commands for the user to check both scenarios.
I think we have to minimize the confusion for the user, considering the two scenarios: 1. the user willingly disabled cuda extensions 2. the user wants cuda extensions, but they may not have installed properly, or be nonfunctional, and tests should check that. Suggestions here too @mhmukadam ?

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
luisenp: I checked with other engineers, and importing does have some overhead, but perhaps it's negligible. Since we are doing benchmarking this half anyway, I vote for keeping this version, and revisit the issue if we see opportunities for improvement later. Sounds good?

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
luisenp: As discussed offline, let's make this test optional on depending on the extension being installed, and consider other mechanisms (e.g., a separate command) for testing that the installation was successful. Additionally, `pytest.importskip` has a `reason=` kwarg that we can use to print a warning message to alert those users that did in fact want to run the CUDA test that their install might be incorrect.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
mhmukadam: Sounds good. I would have proposed the same thing -- keep this and during benchmarking see if this is a bottle in performance to be prioritize and fixed.

mhmukadam (commented):
Looks good. We should be able to merge once the pending comments are resolved in the final rev.

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
maurimo: in new revision I'm adding the "cuda" mark to cuda tests, and a note on the readme on how to skip those tests

maurimo:
this new rev should address all the concerns in the comment, please give a look! @luisenp @mhmukadam

theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py
luisenp: This is great!

luisenp:
LGTM! Unless @mhmukadam has some more comments, I'm OK with merging now.

mhmukadam (approved):
Looks great, thanks for the final push!
(Flagged an issue for the consolidation of some of the autograd code which we can handle in a future PR)

PR #29, opened 2021-12-12
Initial implicit/truncated backward modes

bamos:
Migrating this from the private repo (where we have some more context. I've updated the example/test to use the quadratic cost fitting example as I like it's simplicity to the older example we were using with the cost weights. Here's what the current backward pass mode example looks like:

```
--- backward_mode=FULL
[ 0.09670612 -0.7900081   0.08778726  0.13497931  0.21781093 -0.20041543
  0.11018689 -1.5431342   0.13462512 -0.23546126]

--- backward_mode=IMPLICIT
[ 0.04835452 -0.3950162   0.04389496  0.0674917   0.10890876 -0.10021076
  0.05509511 -0.77159065  0.0673146  -0.11773425]

--- backward_mode=TRUNCATED, backward_num_iterations=5
[ 0.09368691 -0.7653437   0.08504649  0.1307652   0.21101077 -0.19415839
  0.10674679 -1.4949567   0.13042209 -0.22811009]

--- Numeric derivative
[ 0.09670693 -0.7900125   0.08778608  0.13497717  0.21781142 -0.20041636
  0.11019139 -1.54313637  0.13462317 -0.2354647 ]

=== Runtimes
Forward: 1.60e-02 s +/- 1.57e-03 s
Backward (FULL): 4.61e-03 s +/- 3.73e-04 s
Backward (IMPLICIT) 5.18e-04 s +/- 7.88e-05 s
Backward (TRUNCATED, 5 steps) 1.59e-03 s +/- 8.15e-05 s
```

This is almost ready to merge but still WIP as I'm still debugging some scaling issues in the implicit mode, which should be easier to debug with this simpler example.


theseus/core/cost_function.py
```diff
@@ -94,7 +94,8 @@ def __call__(
 class AutoDiffCostFunction(CostFunction):
     def __init__(
         self,
-        optim_vars: List[Variable],
+        # optim_vars: List[Variable],
+        optim_vars: list,  # TODO: List[Vector] raises mypy error?
```
bamos, line 98: @luisenp without this update the mypy test was giving an error since it doesn't seem to accept `Vector` even though it's a child of `Variable`, do you know if there's a better update to this?

```
mypy.....................................................................Failed                                        
- hook id: mypy                                                                                                        
- exit code: 1                                                                                                         
                                                           
theseus/optimizer/nonlinear/tests/test_backwards.py:43: error: Argument 1 to "AutoDiffCostFunction" has incompatible ty
pe "List[Vector]"; expected "List[Variable]"  [arg-type]                                                               
theseus/optimizer/nonlinear/tests/test_backwards.py:43: note: "List" is invariant -- see https://mypy.readthedocs.io/en
/stable/common_issues.html#variance                        
theseus/optimizer/nonlinear/tests/test_backwards.py:43: note: Consider using "Sequence" instead, which is covariant    
examples/backward_modes.py:51: error: Argument 1 to "AutoDiffCostFunction" has incompatible type "List[Vector]"; expect
ed "List[Variable]"  [arg-type]
examples/backward_modes.py:51: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.htm
l#variance
examples/backward_modes.py:51: note: Consider using "Sequence" instead, which is covariant
Found 2 errors in 2 files (checked 101 source files)
```

mhmukadam: Any reason we want to restrict to `Vector` only, here?

examples/backward_modes.py
```diff
@@ -0,0 +1,184 @@
...
+# on a problem fitting a quadratic to data.
+#
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import theseus as th
+import theseus.optimizer.nonlinear as thnl
```
mhmukadam, line 13: Should also add `BackwardMode` to `theseus/__init__.py` so it's available in `import theseus as th` directly.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -45,8 +45,18 @@ class NonlinearOptimizerStatus(Enum):
 @dataclass
 class NonlinearOptimizerInfo(OptimizerInfo):
     converged_iter: torch.Tensor
+    converged_indices: torch.Tensor
+    error_increase_indices: torch.Tensor
```
mhmukadam, line 49: Was the error increase stuff added for the backward support? If not, this might be left over from internal move and was deprecated (for now). cc @luisenp.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -191,43 +202,130 @@ def _optimize_impl(
...
+                verbose=verbose,
+                force_update=False,
+                **kwargs,
+            )
+        elif backward_mode in [BackwardMode.IMPLICIT, BackwardMode.TRUNCATED]:
+            if backward_mode == BackwardMode.IMPLICIT:
+                backward_num_iterations = 1
+            else:
+                if "backward_num_iterations" not in kwargs:
+                    raise ValueError()
```
mhmukadam, line 264: Maybe add some error message.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -191,43 +202,130 @@ def _optimize_impl(
...
+            )
+            info.status[M] = grad_loop_info.status[M]
+            info.converged_indices[M] = grad_loop_info.converged_indices[M]
+            info.converged_iter[M] = (
+                info.converged_iter[M] + grad_loop_info.converged_iter[M]
+            )
+
+            return info
+        else:
+            raise ValueError()
```
mhmukadam, line 308: Add error message along the lines of unknown backward mode.

theseus/optimizer/nonlinear/tests/test_backwards.py
```diff
@@ -0,0 +1,115 @@
...
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import numdifftools as nd
+import pytest  # noqa: F401
+import torch
+
+import theseus as th
+import theseus.optimizer.nonlinear as thnl
```
mhmukadam, line 11: Can remove this after adding `BackwardMode` to the root level init file.

theseus/optimizer/nonlinear/tests/test_backwards.py
```diff
@@ -0,0 +1,115 @@
...
+        track_best_solution=True,
+        verbose=False,
+        backward_mode=thnl.BackwardMode.FULL,
+    )
+    da_dx_full = torch.autograd.grad(updated_inputs["a"], data_x, retain_graph=True)[
+        0
+    ].squeeze()
+    assert torch.allclose(da_dx_numeric, da_dx_full, atol=1e-3)
+
+    # TODO: Debug a little more
```
mhmukadam, line 90: Let's flag an issue for this so we can keep track and prioritize.

examples/backward_modes.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python3
+#
```
mhmukadam, line 2: Move the license stuff up, to line 2.

examples/backward_modes.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def quad_error_fn(optim_vars, aux_vars):
+    a, b = optim_vars
+    x, y = aux_vars
+    est = a.data * x.data.square() + b.data
+    err = y.data - est
+    return err
+
+
```
mhmukadam, line 47: Adding some high level explanation in comments would be useful. Also remember to update the readme in the examples folder as needed. Amortizing some documentation during dev will pay off in the long run.

mhmukadam (approved):
Looks great! Left some minor comments, after which we should merge this and try to debug the implicit grad offset issue in a separate PR (unless you already fixed it).

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Yes, we removed this in [this](https://github.com/fairinternal/theseus/commit/1c74f6d578e64993119be5f7214a2eb5c049b89d) commit, due to some issues in the path planning code. The current version does the following:

* When a batch index passes `convergence_check`, the variables values at that index are not updated again.
* For all other indices, variables store the last value after `max_iterations`
* Info returns
  * For all indices (converged or not) the best solution seen (optional)
  * Error history for all iterations (optional)
  * The iteration at which convergence happened (if any)
  * The iteration at which the best solution was reached 

My main question here is whether having `converged_indices` and `error_increase_indices` is necessary for the backward logic introduced in this PR, or if it's some left over from the old parent, as @mhmukadam suggested.

theseus/core/cost_function.py
bamos: I've reverted this to `List[Variable]` and added ignores in the files causing this issue for now (`examples/backward_modes.py` and `theseus/optimizer/nonlinear/tests/test_backwards.py`). It seems like the issue is that it doesn't recognize a list of `Vector` objects as a subclass of a `Variable` list?

examples/backward_modes.py
bamos: Added, and updated the example/test to use this from `th` directly rather than `thnl`

theseus/optimizer/nonlinear/nonlinear_optimizer.py
bamos: Added

examples/backward_modes.py
bamos: Updated

examples/backward_modes.py
bamos: Added

theseus/optimizer/nonlinear/nonlinear_optimizer.py
bamos: Added

theseus/optimizer/nonlinear/nonlinear_optimizer.py
bamos: I removed `error_increase_indices` and moved `converged_indices` back into the main optimization loop

bamos:
Hi @mhmukadam and @luisenp, I just went through and addressed all the inline comments. Let me know if there's anything else! Otherwise this is ready to merge from my perspective, and I'll keep debugging some of the gradient scaling issues separately for #39.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -122,56 +135,48 @@ def _init_info(
...
-        info = self._init_info(last_err, track_best_solution, verbose)
-        for it_ in range(self.params.max_iterations):
+        converged_indices = self._check_convergence(err, info.last_err)
+        info.status[
+            np.array(converged_indices.detach().cpu())
+        ] = NonlinearOptimizerStatus.CONVERGED
+
+    # loop for the iterative optimizer
+    def _optimize_loop(
+        self,
```
luisenp, line 169: Missing type hints and also `track_best_solution` is not used.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -191,43 +196,132 @@ def _optimize_impl(
...
+                info=grad_loop_info,
+                track_best_solution=False,
+                verbose=verbose,
+                force_update=True,
+                **kwargs,
+            )
+
+            # Merge the converged status into the info from the detached loop,
+            # and for now, don't update the best err tracking or best solution.
+            M = info.status == NonlinearOptimizerStatus.MAX_ITERATIONS
```
luisenp, line 292: I remember this was the result of discussion during a meeting, but I'm having trouble remembering what the issue was. Can you remind me? (or @mhmukadam)

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -191,43 +196,132 @@ def _optimize_impl(
...
+        self,
+        delta: torch.Tensor,
+        converged_indices: torch.Tensor,
+        force_update: bool = False,
     ):
         var_idx = 0
         delta = self.params.step_size * delta
         for var in self.linear_solver.linearization.ordering:
             new_var = var.retract(delta[:, var_idx : var_idx + var.dof()])
             var.update(new_var.data, batch_ignore_mask=converged_indices)
```
luisenp, line 322: Should this line be removed? (since it's repeated inside the `else` clause)

theseus/optimizer/nonlinear/tests/test_backwards.py
```diff
@@ -0,0 +1,118 @@
...
+        theseus_inputs,
+        track_best_solution=True,
+        verbose=False,
+        backward_mode=th.BackwardMode.TRUNCATED,
+        backward_num_iterations=5,
+    )
+    da_dx_truncated = torch.autograd.grad(
+        updated_inputs["a"], data_x, retain_graph=True
+    )[0].squeeze()
+    assert torch.allclose(da_dx_numeric, da_dx_truncated, atol=1e-1)
```
luisenp, line 115: This `atol` seems high. However, since the gradient is actually a constant factor of the numeric gradient (I confirmed this), does it make sense to change this test? For example, something like

```python
diff_ratio = da_dx_numeric / da_dx_truncated
assert diff_ratio.max() - diff_ratio.min()  < TOLERANCE
```

BTW, for implicit, the ratio seems to be exactly = 2. Is this particular of this problem or have you observed this in other problems as well? Also, for truncated, the relative difference wrt reference seems to decrease exponentially in the number of iterations. You may had noticed this already, but just thought I mentioned it, since it could be useful/interesting.

luisenp (commented):
LGTM, some minor comments and a question!

theseus/optimizer/nonlinear/nonlinear_optimizer.py
mhmukadam: I think, it was to keep track of which batch elements had converged or not since implicit (or truncated) would add extra iteration(s).

mhmukadam (commented):
Looks good!

theseus/optimizer/nonlinear/nonlinear_optimizer.py
bamos: Added type hints and removed `track_best_solution` here

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: @bamos and I discussed this and decided to keep this, but we should open an issue to write a better spec for what the error info should look like in each case, and then implement accordingly.

bamos: @luisenp and I discussed this yesterday, it's the logic for merging the infos after the second loop here. I just filed https://github.com/facebookresearch/theseus/issues/44 so we can come up with some examples and decide what the final info should look like after merging this PR.

luisenp: Looks like Brandon just opened #44 for this

theseus/optimizer/nonlinear/tests/test_backwards.py
bamos: Fixed the scaling issue in https://github.com/facebookresearch/theseus/issues/39 and updated the tests to have `atol=1e-4` for both checks

theseus/optimizer/nonlinear/nonlinear_optimizer.py
bamos: Ah good catch, I removed it

bamos:
Thanks for the pass @luisenp! I just addressed your comments and pushed the latest that fixes the scaling issues and passes the backward tests with `atol=1e-4`. Can you let me know if there's anything else?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
mhmukadam: Sounds good!

PR #30, opened 2021-12-14
Adds support for energy based learning with NLL loss (LEO)

psodhi:
This PR introduces LEO, learning energy-based models in optimization (https://arxiv.org/abs/2108.02274). LEO is a method to learn models end-to-end within second-order optimizers like Gauss-Newton. The main difference is that instead of unrolling the optimizer and minimizing the MSE tracking loss, this introduces the NLL energy-based loss that does not backpropagate through the optimizer. It requires low-energy samples from the optimizer, pushing up the energy of optimizer samples and pushing down the energy of ground truth samples.

To execute it, run
`python examples/state_estimation_2d.py` with `learning_method="leo"`

This should update the learnable cost weights so that the optimizer trajectory (orange) matches the ground truth trajectory (green).

### DIFF
--- a/examples/state_estimation_2d.py
+++ b/examples/state_estimation_2d.py
@@ -10,17 +10,51 @@ import torch.nn as nn
 import torch.nn.functional as F
 
 import theseus as th
+import matplotlib.pyplot as plt
 
 device = "cpu"
 torch.manual_seed(0)
 path_length = 50
 state_size = 2
 batch_size = 4
+learning_method = "leo"  # "default", "leo"
+
+vis_flag = False
+plt.ion()
 
 
 # --------------------------------------------------- #
 # --------------------- Utilities ------------------- #
 # --------------------------------------------------- #
+def plot_path(optimizer_path, groundtruth_path):
+    plt.cla()
+    plt.gca().axis("equal")
+
+    plt.xlim(-250, 250)
+    plt.ylim(-100, 400)
+
+    batch_idx = 0
+    plt.plot(
+        optimizer_path[batch_idx, :, 0],
+        optimizer_path[batch_idx, :, 1],
+        linewidth=2,
+        linestyle="-",
+        color="tab:orange",
+        label="optimizer",
+    )
+    plt.plot(
+        groundtruth_path[batch_idx, :, 0],
+        groundtruth_path[batch_idx, :, 1],
+        linewidth=2,
+        linestyle="-",
+        color="tab:green",
+        label="groundtruth",
+    )
+
+    plt.show()
+    plt.pause(1e-12)
+
+
 def generate_path_data(
     batch_size_,
     num_measurements_,
@@ -120,6 +154,42 @@ def get_path_from_values(batch_size_, values_, path_length_):
     return path
 
 
+def get_values_from_path(path_):
+    """
+    :param path_: tensor of dim batch_size_ x path_length_ x 2
+    :return: values: dict of (x,y) pos values
+    """
+    [batch_size_, path_length_, dim] = path_.shape
+    values = {}
+    for i in range(path_length_):
+        values[f"pose_{i}"] = path_[:, i, :2]
+    return values
+
+
+def get_average_sample_cost(x_samples, cost_weights_model, objective, mode_):
+    cost_opt = None
+    n_samples = x_samples.shape[-1]
+    for sidx in range(0, n_samples):
+        x_sample_vals = get_values_from_path(
+            x_samples[:, :, sidx].reshape(x_samples.shape[0], -1, 2)
+        )
+        theseus_inputs = run_model(
+            mode_,
+            cost_weights_model,
+            x_sample_vals,
+            path_length,
+            print_stuff=False,
+        )
+        objective.update(theseus_inputs)
+        if cost_opt is not None:
+            cost_opt = cost_opt + torch.sum(objective.error(), dim=1)
+        else:
+            cost_opt = torch.sum(objective.error(), dim=1)
+    cost_opt = cost_opt / n_samples
+
+    return cost_opt
+
+
 # ------------------------------------------------------------- #
 # --------------------------- Learning ------------------------ #
 # ------------------------------------------------------------- #
@@ -132,7 +202,7 @@ def run_learning(mode_, path_data_, gps_targets_, measurements_):
         def cost_weights_model():
             return model_params * torch.ones(1)
 
-        model_optimizer = torch.optim.Adam([model_params], lr=3e-2)
+        model_optimizer = torch.optim.Adam([model_params], lr=5e-2)
     else:
         cost_weights_model = SimpleNN(state_size, 2, hid_size=100, use_offset=False).to(
             device
@@ -201,14 +271,14 @@ def run_learning(mode_, path_data_, gps_targets_, measurements_):
     state_estimator.to(device)
 
     # ## Learning loop
-    path_tensor = torch.stack(path_data_).permute(1, 0, 2)
     best_loss = 1000.0
+    inner_loop_iters = 3
+    groundtruth_path = torch.stack(path_data_).permute(1, 0, 2)
     best_solution = None
     losses = []
-    for epoch in range(200):
+    for epoch in range(500):
         model_optimizer.zero_grad()
 
-        inner_loop_iters = 3
         theseus_inputs = get_initial_inputs(gps_targets_)
         theseus_inputs = run_model(
             mode_,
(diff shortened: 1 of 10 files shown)


theseus/optimizer/nonlinear/gauss_newton.py
```diff
@@ -41,3 +41,29 @@ def __init__(
 
     def compute_delta(self, **kwargs) -> torch.Tensor:
         return self.linear_solver.solve()
+
+    def compute_samples(self, n_samples: int = 10, T: float = 1.0) -> torch.Tensor:
+        delta = self.linear_solver.solve()
+        AtA = self.linear_solver.precision() / T
```
mhmukadam, line 47: Any reason to have a separate `precision` function call since it just return `AtA` which you could directly access here?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -220,6 +220,9 @@ def _optimize_impl(
     def compute_delta(self, **kwargs) -> torch.Tensor:
         pass
 
+    def compute_samples(self, n_samples: int, T: float) -> torch.Tensor:
```
mhmukadam, line 223: Leave a comment here as discussed to say what's the intended use, in case it's confusing for a user to see a sampling function inside an optimizer.

theseus/optimizer/nonlinear/gauss_newton.py
```diff
@@ -41,3 +41,29 @@ def __init__(
 
     def compute_delta(self, **kwargs) -> torch.Tensor:
         return self.linear_solver.solve()
+
+    def compute_samples(self, n_samples: int = 10, T: float = 1.0) -> torch.Tensor:
```
mhmukadam, line 45: I am guessing `T` is temperature? Maybe pick a more descriptive name or describe in comments. Same for `B`, `N`, `R` below, where you can point to the equations from any reference.

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -65,3 +65,29 @@ def compute_delta(
             ellipsoidal_damping=ellipsoidal_damping,
             damping_eps=damping_eps,
         )
+
+    def compute_samples(self, n_samples: int = 10, T: float = 1.0) -> torch.Tensor:
```
mhmukadam, line 69: This is identical to the one in gauss newton. Maybe pull this up to nonlinear least squares class instead?

theseus/optimizer/nonlinear/gauss_newton.py
```diff
@@ -41,3 +41,29 @@ def __init__(
...
+        delta = self.linear_solver.solve()
+        AtA = self.linear_solver.precision() / T
+        R = torch.linalg.cholesky(AtA, upper=True)
+
+        B, N = delta.shape
+        y = torch.normal(
+            mean=torch.zeros((N, n_samples), device=delta.device),
+            std=torch.ones((N, n_samples), device=delta.device),
+        )
+        delta_samples = (torch.triangular_solve(y, R).solution) + (
```
mhmukadam, line 55: Add a comment for reference to the math.

theseus/tests/test_theseus_layer.py
```diff
@@ -280,11 +301,44 @@ def cost_weight_fn():
...
+        if learning_method == "leo":
+            # groundtruth cost
+            x_gt = target_vars["coefficients"]
+            input_values_gt = {
+                "coefficients": x_gt,
+                cost_weight_param_name: cost_weight_fn(),
+            }
+            layer_to_learn.objective.update(input_values_gt)
+            cost_gt = torch.sum(layer_to_learn.objective.error(), dim=1)  # B x 1
+            cost_gt = cost_gt.to(device)
```
mhmukadam, line 313: Might not be the best place to move between device in the learning loop. This line won't be needed if inputs, objective, layer etc are setup to be on the correct device at the beginning.

theseus/tests/test_theseus_layer.py
```diff
@@ -280,11 +301,44 @@ def cost_weight_fn():
...
+            loss = torch.mean(loss, dim=0)
+        else:
+            loss = F.mse_loss(pred_vars["coefficients"], target_vars["coefficients"])
+
         loss.backward()
-        print(i, loss.item(), loss.item() / loss0)
         optimizer.step()
-        if loss.item() / loss0 < 5e-3:
+
+        mse_loss = F.mse_loss(pred_vars["coefficients"], target_vars["coefficients"])
```
mhmukadam, line 339: Instead of calculating the mse loss again you can shuffle things around.

```
mse_loss = F.mse_loss(pred_vars["coefficients"], target_vars["coefficients"])
....
if learning_method == "leo":
...
else:
    loss = mse_loss

loss.backward()
optimizer.step()

print(i, mse_loss.item())        
if mse_loss.item() / loss0 < 5e-3:
    solved = True
    break
...
```

examples/state_estimation_2d.py
```diff
@@ -236,21 +306,60 @@ def cost_weights_model():
                 print_stuff=epoch % 10 == 0 and i == 0,
             )
 
-        solution_path = get_path_from_values(
-            objective.batch_size, theseus_inputs, path_length
-        )
+        if learning_method == "leo":
```
mhmukadam, line 309: Add comment as discussed about high level explanation of LEO and reference to the paper and any specific equations.

mhmukadam (commented):
Looks great! The API looks good -- just a sampling function in the optimizer class. Added some comments to address. The GPU test is failing due to cholesky upper not available.  We can add support for sparse solvers and optimize performance in later PRs.

luisenp (changes requested):
Great job on this PR! I see all the functionality is there, but I propose we shuffle some things around a bit. I'm OK leaving that as a separate PR, but we should definitely make issues for the comments so we don't forget. 

Regarding the broken test, `upper=` kwarg for cholesky was only introduced in `torch` 1.10.0. It used to be in `torch.cholesky`, but it was removed in  `torch.linalg.cholesky` in v1.9.0 (see screenshot). 

For now, can you update the call to match the one used in v1.9.0 so that the test passes?
image

theseus/optimizer/nonlinear/levenberg_marquardt.py
psodhi: Good point, I'll move this up to nonlinear least squares for now. The main consideration is if the same function is applicable for future least-squares solvers that would derive from this class.

mhmukadam: This proposal is much cleaner so if the logic works generally (say for sparse linear solvers and other possible future solvers) then let's go with this.

Do remember you will need to have run `TheseusLayer.forward` first before calling `TheseusLayer.compute_samples`.

examples/state_estimation_2d.py
```diff
@@ -120,6 +154,42 @@ def get_path_from_values(batch_size_, values_, path_length_):
...
+    :return: values: dict of (x,y) pos values
+    """
+    [batch_size_, path_length_, dim] = path_.shape
+    values = {}
+    for i in range(path_length_):
+        values[f"pose_{i}"] = path_[:, i, :2]
+    return values
+
+
+def get_average_sample_cost(x_samples, cost_weights_model, objective, mode_):
```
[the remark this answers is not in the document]
psodhi: Agreed, sounds good.

examples/state_estimation_2d.py
psodhi: Added comments.

theseus/tests/test_theseus_layer.py
psodhi: Agreed, removed the to(device)

theseus/tests/test_theseus_layer.py
```diff
@@ -280,11 +301,44 @@ def cost_weight_fn():
...
+            layer_to_learn.objective.update(input_values_gt)
+            cost_gt = torch.sum(layer_to_learn.objective.error(), dim=1)  # B x 1
+            cost_gt = cost_gt.to(device)
+
+            # optimizer cost
+            x_opt = pred_vars["coefficients"].detach()
+            x_samples = layer_to_learn.optimizer.compute_samples(
+                n_samples=10, T=1.0
+            )  # B x N x S
+            if x_samples is None:  # use mean solution
```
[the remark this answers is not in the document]
psodhi: It is possible when learning_method == "leo" but the underlying optimizer doesn't define a sampling function. In this case, it reverts to an MMP algorithm where it uses the mean trajectory.

theseus/tests/test_theseus_layer.py
psodhi: Agreed, moved for both test and state_estimation example.

theseus/optimizer/nonlinear/levenberg_marquardt.py
psodhi: On thinking about it more, it does make sense for `compute_samples()` to be independent of the optimizer. I have moved `compute_samples()` to theseus_layer where it uses the current linearization point. However, instead of `linearization`, it is using `linear_solver` as input because we need to compute delta as well within the function.

theseus/optimizer/nonlinear/gauss_newton.py
psodhi: Moved it to `Linearization` where it is called `hessian_approx()`.

theseus/optimizer/nonlinear/levenberg_marquardt.py
psodhi: Yes the final computation is a similar functionality as `retract_and_update_variables`.  One important issue to be flagged for the future is to broadcast the `retract_and_update_variables`. It would make sampling much faster.

theseus/tests/test_theseus_layer.py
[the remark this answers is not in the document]
mhmukadam: Should we add an or to cover x_samples <= 0?

theseus/theseus_layer.py
```diff
@@ -45,6 +46,39 @@ def forward(
...
         return values, info
 
+    def compute_samples(
+        self,
+        linear_solver: LinearSolver = None,
+        n_samples: int = 10,
+        temperature: float = 1.0,
+    ) -> torch.Tensor:
+        if linear_solver is None:
+            return None
```
mhmukadam, line 56: raise error?

theseus/theseus_layer.py
```diff
@@ -45,6 +46,39 @@ def forward(
...
+        delta = linear_solver.solve()
```
mhmukadam, line 57: Is there a way to avoid solving the linear system again? This should be computed in the last iteration in forward.

theseus/optimizer/nonlinear/levenberg_marquardt.py
mhmukadam: Sounds good, I'll flag an issue for this.

mhmukadam (approved):

theseus/optimizer/nonlinear/gauss_newton.py
psodhi: Added a reference.

theseus/theseus_layer.py
psodhi: Returned None here since if learning_method is set to "leo" and there isn't a valid sampling function, this would revert to a default MMP algorithm where it uses the mean trajectory. I can change it to a RaiseError if that seems better.

mhmukadam: Actually this is better. We should just document this in the comments.

psodhi: Sounds good, documented in the comments.

mhmukadam (approved):

PR #33, opened 2021-12-21
Update contrib and add gitattributes

mhmukadam:
- Update contributing file to the branching model we follow.
- Added gitattribute file (to ignore .pynb files) to display correct language on landing page.


luisenp (approved):
LGTM. And thanks for adding the linguist-documentation stuff, that Jupyter % is annoying me! :)

PR #47, opened 2022-01-25
add SO3.hat() and SO3.vee()

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/theseus/geometry/so3.py
+++ b/theseus/geometry/so3.py
@@ -52,7 +52,7 @@ class SO3(LieGroup):
             return f"SO3(matrix={self.data}), name={self.name})"
 
     def _adjoint_impl(self) -> torch.Tensor:
-        raise self.data
+        return self.data
 
     @staticmethod
     def _SO3_matrix_check(matrix: torch.Tensor):
--- a/theseus/geometry/so3.py
+++ b/theseus/geometry/so3.py
@@ -23,18 +23,23 @@ class SO3(LieGroup):
     ):
         if quaternion is not None and data is not None:
             raise ValueError("Please provide only one of theta or data.")
-
         if quaternion is not None:
             dtype = quaternion.dtype
+        if data is not None:
+            self._SO3_matrix_check(data)
         super().__init__(data=data, name=name, dtype=dtype)
-
         if quaternion is not None:
             if quaternion.ndim == 1:
                 quaternion = quaternion.unsqueeze(0)
+            self.update_from_unit_quaternion(quaternion)
 
     @staticmethod
     def _init_data() -> torch.Tensor:  # type: ignore
-        raise torch.eye(3, 3).veiw(1, 3, 3)
+        return torch.eye(3, 3).view(1, 3, 3)
+
+    def update_from_unit_quaternion(self, quaternion: torch.Tensor):
+        self._unit_quaternion_check(quaternion)
+        raise NotImplementedError
 
     def dof(self) -> int:
         return 3
--- a/theseus/geometry/so3.py
+++ b/theseus/geometry/so3.py
@@ -28,7 +28,6 @@ class SO3(LieGroup):
         if quaternion is not None:
             dtype = quaternion.dtype
         if data is not None:
-            dtype = data.dtype
             self._SO3_matrix_check(data)
         super().__init__(data=data, name=name, dtype=dtype)
         if quaternion is not None:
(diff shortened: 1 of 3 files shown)


theseus/geometry/so3.py
```diff
@@ -3,24 +3,42 @@
...
     def __init__(
         self,
+        quaternion: Optional[torch.Tensor] = None,
         data: Optional[torch.Tensor] = None,
         name: Optional[str] = None,
         dtype: Optional[torch.dtype] = None,
     ):
-        raise NotImplementedError
+        if quaternion is not None and data is not None:
+            raise ValueError("Please provide only one of theta or data.")
```
luisenp, line 25: theta --> quaternion

luisenp (changes requested):
Thanks, Taosha! This looks good to me (but I still haven't checked the math carefully). We should add a corresponding test file for SO3 (see SO2 [example](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/tests/test_so2.py)). For now we can add a `test_adjoint()` method for SO3, similar to the one in SO2. This will also cover the implementation of `vee` and `hat`, according to [this](https://github.com/facebookresearch/theseus/blob/58d3c6ee6aac256677a4932a75c5ec4de7e079c3/theseus/geometry/tests/common.py#L52).

Regarding the broken lint, you should install [precommit](https://pre-commit.com/), which will run the lint and mypy before every commit, fix format issues automatically, or otherwise tell you if something else needs to change in the code to pass our tests.

theseus/geometry/so3.py
fantaosha: Thanks, Luis! This is very helpful.

fantaosha: Hi Luis, [test_so3.py](https://github.com/facebookresearch/theseus/blob/taoshaf.add_so3_class/theseus/geometry/tests/test_so3.py) is added. Thanks!

theseus/geometry/so3.py
```diff
@@ -52,15 +54,16 @@ def __str__(self) -> str:
             return f"SO3(matrix={self.data}), name={self.name})"
 
     def _adjoint_impl(self) -> torch.Tensor:
-        return self.data
+        return self.data.clone()
```
luisenp, line 57: I was thinking whether this was necessary when I did the first past to the previous version of this PR, but I wasn't sure. Did you find that not cloning was giving errors (maybe in the test?).

theseus/geometry/so3.py
```diff
@@ -84,10 +87,10 @@ def _compose_impl(self, so3_2: LieGroup) -> "SO3":
         raise NotImplementedError
 
     def _inverse_impl(self, get_jacobian: bool = False) -> "SO3":
-        raise NotImplementedError
+        return SO3(data=self.data.transpose(1, 2).clone())
 
     def to_matrix(self) -> torch.Tensor:
-        return self.data
+        return self.data.clone()
```
luisenp, line 93: Same question regarding `.clone()`

theseus/geometry/tests/test_so3.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest  # noqa: F401
+import torch
+
+import theseus as th
+
+from .common import (  # check_inverse,; check_compose,; check_exp_map,; check_log_map,
```
luisenp, line 11: Remove commented code.

theseus/geometry/so3.py
```diff
@@ -31,10 +51,30 @@ def __repr__(self) -> str:
...
+        if not _check:
+            raise ValueError("Not valid 3D rotations.")
+
+    @staticmethod
+    def _unit_quaternion_check(quaternion: torch.Tensor):
+        if quaternion.ndim != 2 or quaternion.shape[1] != 4:
+            raise ValueError("Quaternions can only be 4-D vectors.")
+
+        if (torch.linalg.norm(quaternion, dim=1) - 1).abs().max().item() >= SO3.SO3_EPS:
+            raise ValueError("Not unit quaternions.")
```
luisenp, line 77: How about `f"SO3 only accepts unit quaternions as inputs but norm(quaternion) = {norm:.3f}."`?

luisenp (approved):
Except for minor comments, this LGTM. Thanks, Taosha! Let's wait to see if @mhmukadam has any comments.

theseus/geometry/so3.py
fantaosha: I added clone() mainly because of encapsulation. I'm a little concerned that users might have access to and modify self.data through adjoint() without clone().

Of course, I can remove clone() and there will be no errors.

theseus/geometry/tests/test_so3.py
fantaosha: I will remove these commented code at the end. This is just used to remind me future tests :)

theseus/geometry/so3.py
fantaosha: Agreed. Will replace the text.

theseus/geometry/so3.py
fantaosha: Please refer to the last [reply](https://github.com/facebookresearch/theseus/pull/47#discussion_r793064803).

theseus/geometry/so3.py
luisenp: This is reasonable. Let's keep as is.

PR #51, opened 2022-01-28
Encapsulated data loading functions in tactile pushing example into a new class

luisenp:
## Motivation and Context
This is one of several coming PRs In preparation for the backward pass experiments. I'm cleaning up the tactile pushing example to make it easier to run the resulting optimization problems in scripts outside of the examples folder.

## How Has This Been Tested
Ran 1 epoch of the example and the results match the original.

image

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/examples/tactile_pose_estimation.py
+++ b/examples/tactile_pose_estimation.py
@@ -63,12 +63,11 @@ plt.ion()
 
 
 def run_learning_loop(cfg):
-    dataset_file = EXP_PATH / "datasets" / f"{cfg.dataset_name}.json"
-    dataset = theg.load_tactile_dataset_from_file(
-        filename=dataset_file, device=device, episode=cfg.episode
-    )
+    dataset_path = EXP_PATH / "datasets" / f"{cfg.dataset_name}.json"
+    sdf_path = EXP_PATH / "sdfs" / f"{cfg.sdf_name}.json"
+    dataset = theg.TactilePushingDataset(dataset_path, sdf_path, cfg.episode, device)
 
-    time_steps = np.minimum(cfg.max_steps, len(dataset["obj_poses"]))
+    time_steps = np.minimum(cfg.max_steps, len(dataset.obj_poses))
 
     min_win_mf, max_win_mf, step_win_mf = (
         cfg.tactile_cost.min_win_mf,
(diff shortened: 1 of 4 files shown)


examples/tactile_pose_estimation.py
```diff
@@ -281,18 +275,11 @@ def run_learning_loop(cfg):
 
     outer_optim = optim.Adam(learnable_params, lr=learning_rate)
     batch_size = cfg.train.batch_size
-    num_batches = cfg.train.num_batches
-    measurements = [
-        (
-            dataset["img_feats"][0:time_steps].unsqueeze(0).repeat(batch_size, 1, 1),
```
luisenp, line 287: @psodhi Maybe you had already explained this and I forgot, but I'm curious about why this repeats the same trajectory data for each batch element and batch. Is this a placeholder? 

If I wanted to change this to select different trajectory data, what changes should I make?

examples/tactile_pose_estimation.py
```diff
@@ -281,18 +275,11 @@ def run_learning_loop(cfg):
...
-        )
-        for _ in range(num_batches)
-    ]
-
-    obj_poses_gt = dataset["obj_poses"][0:time_steps, :].clone().requires_grad_(True)
-    eff_poses_gt = dataset["eff_poses"][0:time_steps, :].clone().requires_grad_(True)
+    measurements = dataset.get_measurements(
+        cfg.train.batch_size, cfg.train.num_batches, time_steps
+    )
+    obj_poses_gt = dataset.obj_poses[0:time_steps, :].clone().requires_grad_(True)
```
luisenp, line 281: @psodhi Another question. Is setting `requires_grad_(True)` necessary here?

mhmukadam (approved):
Looks great, nicely streamlined data loading!

examples/tactile_pose_estimation.py
psodhi: I don't think it is necessary. We can remove it.

examples/tactile_pose_estimation.py
psodhi: Yes, it is a placeholder. It is currently repeating a single trajectory. To select different trajectories, you would have to pass in different episodes indexes (`cfg.episode`) in `dataset = theg.TactilePushingDataset(dataset_path, sdf_path, cfg.episode, device).` And concatenate the outputs to get a batch dataset. Note that different `cfg.episode` trajectories can have different lengths, so it may require some min length trimming / padding.

PR #57, opened 2022-01-31
Add SO3 rotate and unrotate

fantaosha:
## Types of changes

## Checklist


fantaosha:
I installed [precomit](https://pre-commit.com) and passed on my laptop. However, it looks like I still got errors `ci/circleci: py37_linux` and `ci/circleci: py38_linux` about the formats of python files. Do you know why @luisenp? Thanks!

Screen Shot 2022-01-31 at 10 24 48 AM

fantaosha:
SO3 is almost done. I will start working on SE3 if no new features are requested.

theseus/geometry/tests/test_so3.py
```diff
@@ -148,3 +149,52 @@ def test_adjoint():
...
+        check_compose(so3_1, so3_2)
+
+
+def test_rotate_and_unrotate():
+    rng = torch.Generator()
+    rng.manual_seed(0)
+    for _ in range(10):  # repeat a few times
+        for batch_size in [1, 20, 100]:
+            so3 = _create_random_so3(batch_size, rng)
+            # Tests that rotate works from tensor. unrotate() would work similarly), but
```
luisenp, line 169: This comment can be removed, correct?

theseus/geometry/so3.py
```diff
@@ -12,6 +7,11 @@
 from .lie_group import LieGroup
 from .point_types import Point3
 
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
luisenp, line 10: This should be at the top of the file.

luisenp (approved):
LGTM. Only two minor comments.

luisenp:
For the lint, I think we need to update the precommit config, due to an update in black version. I'll push a fix to main and you should rebase this on top.

luisenp:
I pushed a fix to the precommit tests. @fantaosha can you check if rebasing on top of `main` resolves the CI issues?

theseus/geometry/so3.py
fantaosha: This issue is fixed in the newest commit.

theseus/geometry/tests/test_so3.py
fantaosha: Fixed in the newest commit.

theseus/geometry/so3.py
```diff
@@ -274,3 +277,57 @@ def _copy_impl(self, new_name: Optional[str] = None) -> "SO3":
...
+        self,
+        point: Union[Point3, torch.Tensor],
+        jacobians: Optional[List[torch.Tensor]] = None,
+    ) -> Point3:
+        self._rotate_shape_check(point)
+        batch_size = max(self.shape[0], point.shape[0])
+        if isinstance(point, torch.Tensor):
+            p = point.view(-1, 3, 1)
+        else:
+            p = point.data.view(-1, 3, 1)
```
mhmukadam, line 291: Do you need a check here to make sure `point` is of the correct size if passed as a `torch.Tensor`?

mhmukadam (approved):

theseus/geometry/so3.py
fantaosha: > Do you need a check here to make sure `point` is of the correct size if passed as a `torch.Tensor`? Same for unrotate.

I don't think so. `Union[Point3, torch.Tensor]` has already guaranteed `point` is either `Point3` or `torch.Tensor`.

luisenp: I think @mhmukadam meant that when the tensor case is hit (line 289), check that `point.shape` is consistent with what's expected. This is already done in line 286 with the call to `self._rotate_shape_check(point)`.

BTW, note that typing hints are not enforced by Python, it's mostly a help for developers. If the user is not using `mypy`, they won't see any issue with passing arbitrary objects here, until the code crashes somewhere.

PR #68, opened 2022-02-01
add adjoint, hat and vee for SE3

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/theseus/geometry/se3.py
+++ b/theseus/geometry/se3.py
@@ -7,6 +7,9 @@ from typing import List, Optional, Union, cast
 
 import torch
 
+import theseus
+import theseus.constants
+
 from .lie_group import LieGroup
 from .point_types import Point3
 from .so3 import SO3
@@ -58,7 +61,18 @@ class SE3(LieGroup):
 
     @staticmethod
     def _hat_matrix_check(matrix: torch.Tensor):
-        raise NotImplementedError
+        if matrix.ndim != 3 or matrix.shape[1:] != (4, 4):
+            raise ValueError("Hat matrices of SE(3) can only be 4x4 matrices")
+
+        if matrix[:, 3].abs().max().item() > theseus.constants.EPS:
+            raise ValueError("The last row of hat matrices of SE(3) can only be zero.")
+
+        if (
+            matrix[:, :3, :3].transpose(1, 2) + matrix[:, :3, :3]
+        ).abs().max().item() > theseus.constants.EPS:
+            raise ValueError(
+                "The 3x3 top-left corner of hat matrices of SE(3) can only be skew-symmetric."
+            )
 
     @staticmethod
     def exp_map(tangent_vector: torch.Tensor) -> LieGroup:
@@ -78,7 +92,7 @@ class SE3(LieGroup):
 
     def update_from_x_y_z_quaternion(self, x_y_z_quaternion: torch.Tensor):
         if x_y_z_quaternion.ndim != 2 and x_y_z_quaternion.shape[1] != 7:
-            raise ValueError("x_y_z_quaternion can only be 4-D vectors.")
+            raise ValueError("x_y_z_quaternion can only be 7-D vectors.")
 
         batch_size = x_y_z_quaternion.shape[0]
         self.data = torch.empty(batch_size, 3, 4).to(
@@ -87,34 +101,39 @@ class SE3(LieGroup):
         self[:, :3, :3] = SO3.unit_quaternion_to_matrix(x_y_z_quaternion[3:])
         self[:, :, 3] = x_y_z_quaternion[:, :3]
 
-    def update_from_rot_and_trans(
-        self, rotation: SO3, translation: Union[Point3, torch.Tensor]
-    ):
-        raise NotImplementedError
+    def update_from_rot_and_trans(self, rotation: SO3, translation: Point3):
+        if rotation.shape[0] != translation.shape[0]:
+            raise ValueError("rotation and translation must have the same size.")
+
+        if rotation.dtype != translation.dtype:
+            raise ValueError("rotation and translation must be of the same type.")
+
+        if rotation.device != translation.device:
+            raise ValueError("rotation and translation must be on the same device.")
+
+        self.data = torch.cat((rotation.data, translation.data.unsqueeze(2)), dim=2)
 
     @staticmethod
     def hat(tangent_vector: torch.Tensor) -> torch.Tensor:
-        raise NotImplementedError
+        _check = tangent_vector.ndim == 3 and tangent_vector.shape[1:] == (6, 1)
+        _check |= tangent_vector.ndim == 2 and tangent_vector.shape[1] == 3
+        if not _check:
+            raise ValueError("Invalid vee matrix for SO3.")
+        matrix = torch.zeros(tangent_vector.shape[0], 4, 4).to(
+            dtype=tangent_vector.dtype, device=tangent_vector.device
+        )
+        matrix[:, :3, :3] = SO3.hat(tangent_vector[3:])
+        matrix[:, :3, 3] = tangent_vector[:3]
+
+        return matrix
 
     @staticmethod
     def vee(matrix: torch.Tensor) -> torch.Tensor:
-        raise NotImplementedError
+        SE3._hat_matrix_check(matrix)
+        return torch.stack((matrix[:, :3, 3], SO3.vee(matrix[:, :3, :3])), dim=1)
 
     def _transform_shape_check(self, point: Union[Point3, torch.Tensor]):
-        err_msg = "SE(3) can only transform 3-D vectors."
-        if isinstance(point, torch.Tensor):
-            if not point.ndim == 2 or point.shape[1] != 3:
-                raise ValueError(err_msg)
-        elif point.dof() != 3:
-            raise ValueError(err_msg)
-        if (
-            point.shape[0] != self.shape[0]
-            and point.shape[0] != 1
-            and self.shape[0] != 1
-        ):
-            raise ValueError(
-                "Input point batch size is not broadcastable with group batch size."
-            )
+        raise NotImplementedError
 
     def _copy_impl(self, new_name: Optional[str] = None) -> "SE3":
         return SE3(data=self.data.clone(), name=new_name)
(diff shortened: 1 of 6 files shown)


theseus/geometry/se3.py
```diff
@@ -0,0 +1,167 @@
...
+
+from .lie_group import LieGroup
+from .point_types import Point3
+from .so3 import SO3
+
+
+class SE3(LieGroup):
+    def __init__(
+        self,
+        x_y_z_quaternion: Optional[torch.Tensor] = None,
```
mhmukadam, line 21: Given that we follow _rotation-then-translation_ convention (like gtsam), should we make this input consistent with the convention.

theseus/geometry/se3.py
```diff
@@ -0,0 +1,167 @@
...
+        data: Optional[torch.Tensor] = None,
+        name: Optional[str] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        if x_y_z_quaternion is not None and data is not None:
+            raise ValueError("Please provide only one of x_y_z_quaternion or data.")
```
mhmukadam, line 27: Should we be more explicit here and for other Lie groups what we mean by data? For instance here it is a `bx4x4` transformation matrix.

theseus/geometry/se3.py
```diff
@@ -0,0 +1,167 @@
...
+        return SE3(data=ret)
+
+    def to_matrix(self) -> torch.Tensor:
+        ret = torch.zeros(self.shape[0], 4, 4).to(dtype=self.dtype, device=self.device)
+        ret[:, :3] = self.data
+        ret[:, 3, 3] = 1
+        return ret
+
+    def update_from_x_y_z_quaternion(self, x_y_z_quaternion: torch.Tensor):
+        if x_y_z_quaternion.ndim != 2 and x_y_z_quaternion.shape[1] != 7:
```
mhmukadam, line 105: We want to handle the case where the batch dimension is not passed and therefore batch size is assumed to be 1 (possibly also check in other places if we account for this).

theseus/geometry/se3.py
```diff
@@ -0,0 +1,167 @@
...
+        batch_size = x_y_z_quaternion.shape[0]
+        self.data = torch.empty(batch_size, 3, 4).to(
+            device=x_y_z_quaternion.device, dtype=x_y_z_quaternion.dtype
+        )
+        self[:, :3, :3] = SO3.unit_quaternion_to_matrix(x_y_z_quaternion[:, 3:])
+        self[:, :3, 3] = x_y_z_quaternion[:, :3]
+
+    def update_from_rot_and_trans(self, rotation: SO3, translation: Point3):
+        if rotation.shape[0] != translation.shape[0]:
+            raise ValueError("rotation and translation must have the same size.")
```
mhmukadam, line 117: same size -> same batch size

theseus/geometry/se3.py
```diff
@@ -0,0 +1,167 @@
...
+        return cast(SE3, super().copy(new_name=new_name))
+
+    def transform_to(
+        self,
+        point: Union[Point3, torch.Tensor],
+        jacobians: Optional[List[torch.Tensor]] = None,
+    ) -> Point3:
+        raise NotImplementedError
+
+    def untransform_to(
```
mhmukadam, line 162: Do you mean `transform_from` or is this different?

mhmukadam (approved):
Looks good!

theseus/geometry/se3.py
fantaosha: This is consistent with the convention in [`SE2`](https://github.com/facebookresearch/theseus/blob/6a92aff8a924e1bf036118d305a5afb3ff195a86/theseus/geometry/se2.py#L24), which uses `x_y_theta` for initialization. I'm OK with either.

Let me know if you would like to change the convention.

theseus/geometry/se3.py
fantaosha: This line is only used to resolve the conflict where both `x_y_z_quaternion` and `data` are provided.

There is another [function](https://github.com/facebookresearch/theseus/blob/f8cf872dc3d3acffc3232eace5826c6ef486babb/theseus/geometry/se3.py#L62) that explicitly defines what `data` should look like.

theseus/geometry/se3.py
fantaosha: OK. I will add this feature.

theseus/geometry/se3.py
fantaosha: Yeah, I mean `transform_from` and will change the name. 

Also, it seems that no `transform_from` is defined in [`SE2`](https://github.com/facebookresearch/theseus/blob/taoshaf.add_se3/theseus/geometry/se2.py). Should we add it to `SE2`.

theseus/geometry/se3.py
```diff
@@ -0,0 +1,167 @@
...
+    ):
+        if x_y_z_quaternion is not None and data is not None:
+            raise ValueError("Please provide only one of x_y_z_quaternion or data.")
+        if x_y_z_quaternion is not None:
+            dtype = x_y_z_quaternion.dtype
+        if data is not None:
+            self._SE3_matrix_check(data)
+        super().__init__(data=data, name=name, dtype=dtype)
+        if x_y_z_quaternion is not None:
+            if x_y_z_quaternion is not None:
```
luisenp, line 34: Repeated if statement.

theseus/geometry/se3.py
luisenp: TBH, I'm not sure if we support this in 2D functions. Should we open an issue to remember to check for this?

theseus/geometry/se3.py
luisenp: Created issue #69 for this to address in a separate PR. I'd say this is lower priority, so it can wait until the 3D groups are finished.

luisenp (approved):
LGTM!

theseus/geometry/se3.py
fantaosha: Fixed.

theseus/geometry/se3.py
fantaosha: Fixed.

theseus/geometry/se3.py
fantaosha: There is `SO2.update_from_angle` but `SE2` seems to have no similar functions.

theseus/geometry/se3.py
mhmukadam: SE3 data is rotation-then-translation. @luisenp, @dongjing3309 any thoughts?

theseus/geometry/se3.py
mhmukadam: That makes sense. I meant, when this error statement is printed the user will have an opaque prompt for `data`.

theseus/geometry/se3.py
mhmukadam: > TBH, I'm not sure if we support this in 2D functions. Should we open an issue to remember to check for this?

Okay with either of these but we should be consistent:
a. We clarify in docs that batch dimension is always present and users need to add a stub dim for a batch size of one.
b. We let users define things without batch and append the extra dim internally everywhere.

theseus/geometry/se3.py
```diff
@@ -123,7 +122,7 @@ def update_from_x_y_z_quaternion(self, x_y_z_quaternion: torch.Tensor):
 
     def update_from_rot_and_trans(self, rotation: SO3, translation: Point3):
         if rotation.shape[0] != translation.shape[0]:
-            raise ValueError("rotation and translation must have the same size.")
+            raise ValueError("rotation and translation must have the same batch size.")
```
mhmukadam, line 125: Flagging this for the scenario when a user might pass rotation and translation without a batch dim.

theseus/geometry/se3.py
luisenp: Actually, for some reason I was thinking of something else. In Lie group classes constructors we do automatically append batch dimensions if not present, and @fantaosha also added this here in [line 35](https://github.com/facebookresearch/theseus/pull/68/files/f8cf872dc3d3acffc3232eace5826c6ef486babb#diff-c5e3f8d4504d1054716f8e6877b315b545ff52580232d8433d837c74e0ef4461R35). I guess the question is if it makes sense to move the `unsqueeze` inside the `updated_from_x_y_z_quaternion` function, which I think it does.

theseus/geometry/se3.py
luisenp: I don't have a strong preference here. Aesthetically it seems more pleasing to me to be consistent between `SE2` and `SE3` (even if it means refactoring `SE2` to use `theta_x_y`). But we should def give priority to any well-established conventions for working with se2 and se3 objects over this.

theseus/geometry/se3.py
luisenp: I think this is fine as is. The format for data will be documented and there are also other exceptions thrown if the user provides the wrong format for data.

theseus/geometry/se3.py
luisenp: Created #76 for this, since it also involves SE2 changes.

theseus/geometry/se3.py
luisenp: Let's move the `unsqueeze` for adding a batch dimension (if needed) inside `updated_from_x_y_z_quaternion` before merging.

theseus/geometry/se3.py
luisenp: Since inputs are `SO3` and `Point3`, they would already have a batch dim here, but we should add some type checks.

theseus/geometry/se3.py
mhmukadam: @fantaosha @luisenp was this addressed before merging? If not maybe add to an active PR.

theseus/geometry/se3.py
mhmukadam: @fantaosha @luisenp was this addressed before merging? If not, maybe add to an active PR.

luisenp: Yes, I meant a type check. But I don't think it's super important to have this, it'd be mostly for the user to get a less confusing error message, but I don't expect this function to be used a lot directly.

Issue #70, opened 2022-02-02
Add all the EPS to `constants.py`


fantaosha:
Should this go to `constants.py`? Was this arbitrarily set or comes from another source?

_Originally posted by @mhmukadam in https://github.com/facebookresearch/theseus/pull/46#discussion_r795569723_

PR #71, opened 2022-02-02
Add SE3.exp_map() and SE3.log_map()

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/theseus/geometry/se3.py
+++ b/theseus/geometry/se3.py
@@ -7,6 +7,9 @@ from typing import List, Optional, Union, cast
 
 import torch
 
+import theseus
+import theseus.constants
+
 from .lie_group import LieGroup
 from .point_types import Point3
 from .so3 import SO3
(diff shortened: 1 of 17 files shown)


theseus/geometry/se3.py
```diff
@@ -97,10 +97,138 @@ def _hat_matrix_check(matrix: torch.Tensor):
 
     @staticmethod
     def exp_map(tangent_vector: torch.Tensor) -> LieGroup:
-        raise NotImplementedError
+        if tangent_vector.ndim != 2 or tangent_vector.shape[1] != 6:
+            raise ValueError("Tangent vectors of SE(3) can only be 6-D vectors.")
+
+        NEAR_ZERO_EPS = 5e-3
```
mhmukadam, line 103: We should consolidate all the EPS, otherwise these spread around could make debugging messy.

theseus/geometry/tests/test_se3.py
```diff
@@ -3,20 +3,122 @@
...
+        tangent_vector_ang *= torch.rand(batch_size, 1).double() * 2 * np.pi - np.pi
+        tangent_vector_lin = torch.randn(batch_size, 3).double()
+        tangent_vector = torch.cat([tangent_vector_lin, tangent_vector_ang], dim=1)
+
+        check_exp_map(tangent_vector, th.SE3)
+
+    for batch_size in [1, 20, 100]:
+        tangent_vector_ang = torch.rand(batch_size, 3).double() - 0.5
+        tangent_vector_ang /= tangent_vector_ang.norm(dim=1, keepdim=True)
+        tangent_vector_ang *= 1e-5
```
mhmukadam, line 43: Since all these for loops differ only on this line, maybe leave comments about why we are testing on these sets of values.

mhmukadam (approved):

theseus/geometry/se3.py
luisenp: This is part of #70

theseus/geometry/tests/test_se3.py
luisenp: If I understand correctly, this is to test different corner cases (e.g., near zero tangent vectors). If so, we should also add similar tests for SE2 and SO2 (in a new PR). But agree that adding some small comments here would be useful.

fantaosha: Yeah, as @luisenp said,  `1e-5` and `3e-3` will implement different methods to calculation the `exp_map` and `log_map`.

fantaosha: I will add some comments about these tests.

fantaosha: Comments added in the new commit.

PR #74, opened 2022-02-04
Add automatic differentiation on the Lie group tangent space

fantaosha:
## Motivation and Context

Add support for the automatic Jacobian and gradient computation on the Lie group tangent space.

## Types of changes

## Checklist

### DIFF
--- a/examples/se2_inverse.py
+++ b/examples/se2_inverse.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
+# This example will use torch to solve min_x1 || x1^-1 - x2 ||^2,
+# where x1 and x2 are SE2 objects
+
+import torch
+import theseus as th
+from theseus import LieGroupTensor
+from theseus.geometry.lie_group import LieGroup
+
+from theseus.geometry.tests.test_se2 import create_random_se2
+
+rng = torch.Generator()
+rng.manual_seed(0)
+x1 = create_random_se2(1, rng)
+x2 = create_random_se2(1, rng)
+
+
+def run(x1: LieGroup, x2: LieGroup, num_iters=10, use_lie_tangent=True):
+    x1.data = LieGroupTensor(x1)
+    x1.data.requires_grad = True
+
+    optim = torch.optim.Adam([x1.data], lr=1e-1)
+    scheduler = torch.optim.lr_scheduler.MultiStepLR(
+        optim, milestones=[250, 600], gamma=0.01
+    )
+    for i in range(num_iters):
+        optim.zero_grad()
+        cf = th.eb.VariableDifference(x1.inverse(), th.ScaleCostWeight(1.0), x2)
+        loss = cf.error().norm()
+        if i % 100 == 0:
+            print(
+                "iter {:0>4d}: loss is {:.10f}, cos(theta)^2 + sin(theta)^2 is {:.10f}".format(
+                    i, loss.item(), x1[0, 2:].norm().item() ** 2
+                )
+            )
+        loss.backward()
+
+        with th.set_lie_tangent_enabled(use_lie_tangent):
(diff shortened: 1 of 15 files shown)


fantaosha:
The automatic differentiation on the Lie group tangent space works for all the Lie groups in `theseus`. Maybe we can test some examples without providing any `Jacobians` manually.

mhmukadam (approved):
Looks great, nice work!

theseus/geometry/tests/test_so2.py
```diff
@@ -108,3 +108,44 @@ def test_copy():
     rng.manual_seed(0)
     so2 = _create_random_so2(1, rng)
     check_copy_var(so2)
+
+
+def test_projection():
```
fantaosha, line 113: This is used to test the correctness of the automatic Jacobean calculation.

theseus/geometry/tests/test_se3.py
```diff
@@ -185,3 +185,46 @@ def test_transform_from_and_to():
             )
             assert torch.allclose(jacobians_from[0], expected_jac[0])
             assert torch.allclose(jacobians_from[1], expected_jac[1])
+
+
+def test_projection():
```
fantaosha, line 190: This is used to test automatic jacobian calculation for `SE(3)`

examples/se2_inverse.py
```diff
@@ -0,0 +1,67 @@
...
+
+from theseus.geometry.tests.test_se2 import create_random_se2
+
+rng = torch.Generator()
+rng.manual_seed(0)
+x1 = create_random_se2(1, rng)
+x2 = create_random_se2(1, rng)
+
+
+def run(x1: LieGroup, x2: LieGroup, num_iters=10, use_lie_tangent=True):
```
luisenp, line 23: Given that this is in the examples folder, let's add some text at the beginning explaining what this example is showing.

examples/se2_inverse.py
```diff
@@ -0,0 +1,67 @@
...
+    x1.data = LieGroupTensor(x1)
```
luisenp, line 24: Is the change that is going to make `LieGroup.data` be of type `LieGroupTensor` going to be part of this PR?

theseus/geometry/tests/test_se2.py
```diff
@@ -120,3 +120,29 @@ def test_theta_jacobian():
...
+                return th.SE2(data=g).transform_to(p).data.sum(dim=0)
+
+            jac = torch.autograd.functional.jacobian(
+                transform_to_sum, (se2.data, point.data)
+            )
+
+            actual = [
+                se2.project(jac[0]).transpose(0, 1),
+                point.project(jac[1]).transpose(0, 1),
+            ]
```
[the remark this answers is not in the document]
luisenp: I think this is just needed for the test to compare that `project(euclidean_grad)` gives the same result as `tangent_grad`. BTW, @fantaosha, why is the `transpose` needed? Isn't dim=0 a batch dimension? 

Also, a few days ago we mentioned that `AutodiffCostFunction` will always call `project` on the computed jacobians before [returning](https://github.com/facebookresearch/theseus/blob/main/theseus/core/cost_function.py#L178). Should this be part of this PR? (this is related to Mustafa's question)

luisenp (commented):
Left some questions to address before merging.

examples/se2_inverse.py
mhmukadam: Also update the readme in that folder accordingly.

fantaosha:
Please refer to #98 for enhanced `LieGroup.autograd`.

examples/se2_inverse.py
fantaosha: No. This will cause more issues than I previously thought.

examples/se2_inverse.py
fantaosha: Will do.

examples/se2_inverse.py
luisenp: Let's merge without this. I'll look at this part next week and follow up with you if I run into some problems.

mhmukadam (approved):

PR #75, opened 2022-02-07
Add Lie Group Support to PyTorch Optimizers

fantaosha:
## Motivation and Context

Apply `torch.optim` to run optimization and train neural networks on Lie groups.

## Types of changes

## Checklist


examples/se2_inverse.py
```diff
@@ -0,0 +1,65 @@
+# This example will use torch to solve min_x1 || x1^-1 - x2 ||^2,
```
mhmukadam, line 1: Add license add the top and remember to update the readme in examples folder as necessary.

theseus/geometry/utils/utils.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+
+class LieGroupTensor(torch.Tensor):
```
mhmukadam, line 9: I am wondering if we need this class to be in a separate utils folder (i.e. do we anticipate more stuff) or we can just have all this in theseus/geometry.

examples/se2_inverse.py
```diff
@@ -0,0 +1,65 @@
...
+# where x1 and x2 are SE2 objects
+
+import torch
+import theseus as th
+from theseus import LieGroupTensor
+from theseus.geometry.lie_group import LieGroup
+
+from theseus.geometry.tests.test_se2 import create_random_se2
```
mhmukadam, line 9: We could move more generally useful stuff like create_random_se2 from tests to theseus/utils.

examples/se2_inverse.py
```diff
@@ -0,0 +1,65 @@
...
+
+def run(x1: LieGroup, x2: LieGroup, num_iters=10, use_lie_tangent=True):
+    if use_lie_tangent:
+        # update x1 with the Lie group tangent gradient
+        x1.data.__class__ = LieGroupTensor
+        x1.data.__dict__["group_cls"] = type(x1)
+    else:
+        # update x1 with the Euclidean gradient
+        x1.data.__class__ = torch.Tensor
+        x1.data.__dict__.pop("group_cls", None)
```
mhmukadam, line 25: I think if we can find a way to add this logic to the LieGroup class itself and use a variable (like dtype or device) to specific which grad mode to use then the overall api would be a lot more user friendly.

mhmukadam (commented):
Nice work! We should also add an example/test showing usage with a TheseusLayer objective defined on Lie group variables and an inner nonlinear optimizer and an outer pytorch optimizer.

examples/se2_inverse.py
```diff
@@ -0,0 +1,67 @@
...
+        loss = cf.error().norm()
+        if i % 100 == 0:
+            print(
+                "iter {:0>4d}: loss is {:.10f}, cos(theta)^2 + sin(theta)^2 is {:.10f}".format(
+                    i, loss.item(), x1[0, 2:].norm().item() ** 2
+                )
+            )
+        loss.backward()
+
+        with th.set_lie_tangent_enabled(use_lie_tangent):
```
fantaosha, line 43: The users need this line to update `LieGroup` on its tangent space not the Euclidean space.

examples/se2_inverse.py
```diff
@@ -0,0 +1,67 @@
...
+from theseus.geometry.tests.test_se2 import create_random_se2
+
+rng = torch.Generator()
+rng.manual_seed(0)
+x1 = create_random_se2(1, rng)
+x2 = create_random_se2(1, rng)
+
+
+def run(x1: LieGroup, x2: LieGroup, num_iters=10, use_lie_tangent=True):
+    x1.data = LieGroupTensor(x1)
```
fantaosha, line 24: We don't need this line if `theseus.LieGroup.data` is changed from `torch.Tensor` to `theseus.LieGroupTensor`.

examples/se2_inverse.py
fantaosha: > I think if we can find a way to add this logic to the LieGroup class itself and use a variable (like dtype or device) to specific which grad mode to use then the overall api would be a lot more user friendly.

I'm OK for either.

examples/se2_inverse.py
fantaosha: Updated.

theseus/geometry/utils/utils.py
fantaosha: I have no preference and I'm OK for this.

theseus/geometry/utils/utils.py
```diff
@@ -0,0 +1,92 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import threading
+from typing import Any
+
+import torch
+
+
+class LieGroupContext(object):
+    contexts = threading.local()
```
fantaosha, line 13: Someone said that `threading.local()` is safe for multi-thread processing. Are there any other concerns with this? @luisenp @mhmukadam

examples/se2_inverse.py
luisenp: Created #78 for this.

theseus/geometry/utils/utils.py
```diff
@@ -0,0 +1,92 @@
...
+        self.prev = super().get_context()
+        LieGroupContext.set_context(False)
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        LieGroupContext.set_context(self.prev)
+
+
+class LieGroupTensor(torch.Tensor):
+    from torch._C import _disabled_torch_function_impl
```
luisenp, line 58: Is this related to [this](https://github.com/pytorch/pytorch/issues/46159#issuecomment-707817037) issue? Can you explain a bit why this part is necessary?

theseus/geometry/utils/utils.py
luisenp: I think we can keep these in the same folder, as @mhmukadam suggests. On the other hand, we want to discourage users from using these classes directly, so we should rename as `_LieGroupTensor` and `_LieGroupContext`.

theseus/geometry/utils/utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def __new__(cls, group):
+        return torch.Tensor._make_subclass(cls, group.data)
+
+    def __init__(self, group):
+        self.group_cls = type(group)
+
+    def add_(self, update, alpha=1):
+        if LieGroupContext.get_context():
+            group = self.group_cls(data=self.data)
```
luisenp, line 70: One concern here and for the functions below is that this hard codes the assumption that a call to `add_` wants to add the gradient, which might not be the case for all of the optimization algorithms in Python. 

Is there a way to change this implementation so that within the context `lie_tensor.grad` value is replaced by its projection to tangent space, and `add_` is just a call to `retract`?

theseus/geometry/utils/utils.py
luisenp: Nothing comes to mind right now, but we should keep an eye open and tests this thoroughly.

theseus/geometry/utils/utils.py
```diff
@@ -0,0 +1,92 @@
...
+        if not hasattr(cls.contexts, "use_lie_tangent"):
+            cls.contexts.use_lie_tangent = False
+        return cls.contexts.use_lie_tangent
+
+    @classmethod
+    def set_context(cls, use_lie_tangent: bool):
+        cls.contexts.use_lie_tangent = use_lie_tangent
+
+
+class set_lie_tangent_enabled(object):
```
luisenp, line 26: Could we rename these as `set_tangent_enabled`, `enable_tangent`, `no_tangent`? (dropping the "lie"). Is that unambiguous enough? Since these context will be used a lot, I'd prefer shorter names.

cc: @mhmukadam

examples/se2_inverse.py
```diff
@@ -0,0 +1,67 @@
...
+    x1.data.requires_grad = True
+
+    optim = torch.optim.Adam([x1.data], lr=1e-1)
```
luisenp, line 27: Once we converge on a solution we are satisfied with, we should test this with all of the optimizer under `torch.optim`.

examples/se2_inverse.py
luisenp: I think this should be fine, but let's wait until we finalize the implementation of `LieGroupTensor`.

luisenp (changes requested):
This is a clever workaround! I left some important questions, and we need to test this carefully to avoid unwanted behavior, but this is promising.

theseus/geometry/utils/utils.py
fantaosha: Agreed. Will make this change.

theseus/geometry/utils/utils.py
fantaosha: That's what I previously tried. However, we might might implement `torch.optim` algorithms in that case. This is something similar to the notion of `virtual function` and `retract()` will be called only if`LieGroupContext.get_context()` is `True`.

examples/se2_inverse.py
fantaosha: I created a new issue #79.

theseus/geometry/utils/utils.py
fantaosha: My concern is that the Euclidean gradient is mathematically on the Euclidean tangent space, while the Lie tangent gradient is on the Lie group tangent space. Both of them are the tangent spaces.

theseus/geometry/utils/utils.py
fantaosha: BTW, note that `LieGroupContext.get_context()` is `False` by default and the users need something like `with theseus.lie_tangent_enabled` to trigger `retract`.

theseus/geometry/utils/utils.py
luisenp: Ah, OK, that makes sense. Then it's probably safest to keep the "lie" prefix.

theseus/geometry/utils/utils.py
luisenp: Yes, I'm talking about the case where the `LieGroupContext.get_context()` is `True`. Even in that scenario, we don't know if some torch optimizer step will try to do an in-place addition whose argument is not the gradient, but some other vector (e.g. ,a different descent direction), and in this case doing the projection would be wrong, no? Or is this projection correct for any vector in eucildean space, not just the exact gradient?

examples/se2_inverse.py
luisenp: I'd suggest doing these tests as part of this PR (but once we agree on the final solution).

fantaosha: Agreed.

theseus/geometry/utils/utils.py
fantaosha: See #79

theseus/geometry/utils/utils.py
fantaosha: See #93 and I will create a new PR for it.

theseus/geometry/utils.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
fantaosha, line 1: I move `theseus/geometry/utils/utils.py` to `theseus/geometry/utils.py`

theseus/geometry/utils/utils.py
mhmukadam: Captured in #92.

PR #81, opened 2022-02-14
Forced Gauss-Newton step for last iterations of truncated backward.

luisenp:
## Motivation and Context
Our current derivation of implicit/truncated backward modes hols only for Gauss-Newton. However, regardless of the optimization method used, we can first find the fixed point, then force the optimizer to apply a GN step. Preliminary experiments with Levenberg-Marquardt show that this improves the error in gradient computation, compared with numerical derivatives. 

Marking as bugfix, since this means that previous version of LM was incorrect. 

## Types of changes

## Checklist

### DIFF
--- a/theseus/optimizer/nonlinear/nonlinear_optimizer.py
+++ b/theseus/optimizer/nonlinear/nonlinear_optimizer.py
@@ -180,6 +180,11 @@ class NonlinearOptimizer(Optimizer, abc.ABC):
             self.linear_solver.linearization.linearize()
             try:
                 if truncated_grad_loop:
+                    # The derivation for implicit differentiation states that
+                    # the autograd-enabled loop (which `truncated_grad_loop` signals)
+                    # must be done using Gauss-Newton steps. Well, technically,
+                    # full Newton, but it seems less stable numerically and
+                    # GN is working well so far.
                     delta = self.linear_solver.solve()
                 else:
                     delta = self.compute_delta(**kwargs)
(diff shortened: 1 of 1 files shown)


bamos (approved):

mhmukadam (approved):

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -179,7 +179,10 @@ def _optimize_loop(
             # do optimizer step
             self.linear_solver.linearization.linearize()
             try:
-                delta = self.compute_delta(**kwargs)
+                if truncated_grad_loop:
```
mhmukadam, line 182: how about
```
if backward_mode == BackwardMode.IMPLICIT or backward_mode == BackwardMode.TRUNCATED
```
so we have less redundant variables?

luisenp: Agreed.

luisenp: Made #91 so we make this in a new PR, because it will create conflicts with several currently opened PRs.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -179,7 +179,10 @@ def _optimize_loop(
...
+                    delta = self.linear_solver.solve()
+                else:
+                    delta = self.compute_delta(**kwargs)
```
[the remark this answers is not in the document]
luisenp: Considering the offline conversation around #85, I don't think this PR addresses the original issue, so I'm inclined to close it without merging (cc @bamos).

mhmukadam:
@luisenp, should we merge this now (after comments are resolved) given that we are not going with the full Hessian route?

luisenp:
@mhmukadam Yes, but I'm planning to make a few changes before merging. In particular, I'm not sure we actually need this for truncated grad, in the sense that (AFAIK) there is no theory specifying that the grad enabled loops need to have some specific form. @bamos ?

Issue #76, opened 2022-02-09
Refactor SE2 and SE3 to use rotation-then-translation format


luisenp:
Sophus uses rotation-then-translation everywhere, gtsam is the same except for SE2 where it is flipped.

We should consider refactoring SE2 in that case, unless there are any consequences (and why gtsam uses the flipped convention for it). Aside from the function signature, the data order would also likely change then (x, y, c, s -> c, s, x, y).

_Originally posted by @mhmukadam in https://github.com/facebookresearch/theseus/pull/68#discussion_r801450549_

PR #84, opened 2022-02-16
Add forward kinematics using an URDF to theseus.embodied.kinematics.

exhaustin:
## Types of changes

## Checklist


theseus/embodied/kinematics/tests/test_urdf_model.py
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+def test_forward_kinematics_seq(robot_model, dataset):
+    ee_name = dataset["ee_name"]
+
+    for joint_state, ee_pose_target in zip(dataset["joint_states"], dataset["ee_poses"]):
+        ee_se3_target = SE3(x_y_z_quaternion=ee_pose_target)
+        ee_se3_computed = robot_model.forward_kinematics(joint_state)[ee_name]
+
+        assert torch.allclose(ee_se3_target.between(ee_se3_computed).log_map(), torch.zeros(6))
```
luisenp, line 38: We can replace the first argument by `ee_se3_target.local(ee_se3_computed)`.

theseus/embodied/kinematics/tests/test_urdf_model.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_forward_kinematics_batched(robot_model, dataset):
+    ee_name = dataset["ee_name"]
+
+    ee_se3_target = SE3(x_y_z_quaternion=dataset["ee_poses"])
+    ee_se3_computed = robot_model.forward_kinematics(dataset["joint_states"])[ee_name]
+
+    assert torch.allclose(ee_se3_target.between(ee_se3_computed).log_map(), torch.zeros(dataset["num_data"], 6))
```
luisenp, line 46: Same comment about simplifying with `local()`.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -39,3 +41,27 @@ def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
 
     def dim(self) -> int:
         return 1
+
+class UrdfRobotModel(RobotModel):
+    def __init__(self, urdf_path: str):
+        self.robot_model = drm.DifferentiableRobotModel(urdf_path)
+    
+    def forward_kinematics(self, joint_states: torch.Tensor) -> Dict[str, SE3]:
```
luisenp, line 49: We should consolidate the API of all `forward_kinematics` classes to match the parent class. To avoid using dicts when it's not practical (e.g., `IdentityModel`), we could maybe define a new type, for example

```python
RobotModelOutput = Union[LieGroup, Dict[str, LieGroup]]
```
which would be used as the output `RobotModel.forward_kinematics`, and let subclasses override with one of the subtypes. Alternatively, we could just pick one data type and use this for all subclasses (i.e., always a dict, or tuple, etc.).

Any thoughts?

luisenp (commented):
Nice job! This LGTM with some minor comments. @mhmukadam, any thoughts on the API for the `UrdfRobotModel.forward_kinematics()` method?

BTW, @exhaustin looks like you haven't installed the precommit hooks. You can follow the instructions [here](https://github.com/facebookresearch/theseus/blob/main/CONTRIBUTING.md#code-style).

exhaustin:
I installed the pre-commit hooks which indicated that the code wasn't passing due to the interfaces defined by the `RobotModel` parent class. The current definition feels off to me and I would also need understanding of how it's being used in [collision.py](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/collision/collision.py).
Can we can discuss in person when Mustafa's back?

luisenp:
> I installed the pre-commit hooks which indicated that the code wasn't passing due to the interfaces defined by the `RobotModel` parent class. The current definition feels off to me and I would also need understanding of how it's being used in [collision.py](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/collision/collision.py). Can we can discuss in person when Mustafa's back?

That's OK. Once we agree on the interface for `RobotModel`, as I mentioned in my comment above, then I can make the necessary changes to `RobotModel`, 2D collision, etc. in a separate PR, and you can then rebase this one of top of that. Let's wait for @mhmukadam to comment on the discussion above.

theseus/embodied/kinematics/robot_model.py
exhaustin: Suggested interface for RobotModel.forward_kinematics:
```py
forward_kinematics(Union[torch.Tensor, theseus.Vector]) -> RobotModelOutput
```

Will confirm after constructing a working example of an optimization problem using the robot model.

luisenp: Sounds good. I'll try to have a PR ready today with the old code adapted to this change. I'll start with `RobotModelInput` type which I'll set to `Union[torch.Tensor, theseus.Vector]`, and we can revisit this later if this seems overkill.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -4,10 +4,12 @@
...
+from typing import Dict
 
 import torch
+import differentiable_robot_model as drm
 
-from theseus.geometry import SE2, LieGroup, Point2, Vector
+from theseus.geometry import SE2, SE3, LieGroup, Point2, Vector
 
 
 class RobotModel(abc.ABC):
```
mhmukadam, line 15: We could rename this to something more general so we can derive from it urdf robot models and additionally later, human models, hand models, etc.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -39,3 +41,27 @@ def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
...
+        link_poses = {}
+        for link_name in self.robot_model.get_link_names():
+            pos, quat = self.robot_model.compute_forward_kinematics(joint_states, link_name)
+            link_poses[link_name] = SE3(x_y_z_quaternion=torch.cat([pos, quat[..., 3:], quat[..., :3]], dim=-1))
+
+        return link_poses
+    
+    def dim(self) -> int:
+        return len(self.robot_model.get_joint_limits())
+
```
mhmukadam, line 67: These missing new lines should get fixed by black with the pre-commit hooks.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -39,3 +41,27 @@ def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
...
+        Args:
+            joint_states: Vector of all joint angles
+        Outputs:
+             Dictionary that maps link name to link pose
+
+        (Note: differentiable robot model uses the xyzw quaternion convention, while theseus uses wxyz, hence the conversion.)
+        """
+        link_poses = {}
+        for link_name in self.robot_model.get_link_names():
+            pos, quat = self.robot_model.compute_forward_kinematics(joint_states, link_name)
```
mhmukadam, line 60: Note that for consistency we will be using rotation then translation everywhere #76, so this will get modified a bit. cc @luisenp @fantaosha

theseus/embodied/kinematics/robot_model.py
```diff
@@ -39,3 +41,27 @@ def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
 
     def dim(self) -> int:
         return 1
+
+class UrdfRobotModel(RobotModel):
+    def __init__(self, urdf_path: str):
+        self.robot_model = drm.DifferentiableRobotModel(urdf_path)
```
mhmukadam, line 47: Field name and class name (the parent class name might change) being similar might lead to some confusion.

mhmukadam (commented):
Looks good to me so far. I agree, we'll be able to ground the discussion more one we reach a full example, then we can work our way backwards from there to streamline the api.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -40,3 +41,34 @@ def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]
...
+        Outputs:
+             Dictionary that maps link name to link pose
+
+        (Note: differentiable robot model uses the xyzw quaternion convention,
+        while theseus uses wxyz, hence the conversion.)
+        """
+        link_poses: Dict[str, LieGroup] = {}
+        for link_name in self.drm_model.get_link_names():
+            pos, quat = self.drm_model.compute_forward_kinematics(
+                joint_states, link_name
```
mhmukadam, line 63: If joint states passed are torch.Tensor then we would need a size check (including appending batch dim if not passed?). Vector initialization should handle check externally so we are good on that.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -40,3 +41,34 @@ def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]
...
+        Args:
+            joint_states: Vector of all joint angles
+        Outputs:
+             Dictionary that maps link name to link pose
+
+        (Note: differentiable robot model uses the xyzw quaternion convention,
+        while theseus uses wxyz, hence the conversion.)
+        """
+        link_poses: Dict[str, LieGroup] = {}
+        for link_name in self.drm_model.get_link_names():
```
mhmukadam, line 61: Not sure if this needs to happen on DRM side but the intermediate manipulator jacobians can be reused/cached for efficiency when many/all links are being evaluated (does this already happen under the hood?).

theseus/embodied/kinematics/tests/test_urdf_model.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_forward_kinematics_seq(robot_model, dataset):
+    ee_name = dataset["ee_name"]
+
+    for joint_state, ee_pose_target in zip(
+        dataset["joint_states"], dataset["ee_poses"]
+    ):
+        ee_se3_target = SE3(x_y_z_quaternion=ee_pose_target)
+        ee_se3_computed = robot_model.forward_kinematics(joint_state)[ee_name]
```
mhmukadam, line 43: Maybe also have a version where you test this with `th.Vector` input.

theseus/embodied/kinematics/tests/test_urdf_model.py
```diff
@@ -0,0 +1,59 @@
...
+        assert torch.allclose(
+            ee_se3_target.local(ee_se3_computed), torch.zeros(6)
+        )
+
+
+def test_forward_kinematics_batched(robot_model, dataset):
+    ee_name = dataset["ee_name"]
+
+    ee_se3_target = SE3(x_y_z_quaternion=dataset["ee_poses"])
+    ee_se3_computed = robot_model.forward_kinematics(dataset["joint_states"])[ee_name]
```
mhmukadam, line 54: Maybe also have a version where you test this with th.Vector input.

Simply, `th.Vector(dataset["joint_states"])`.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -0,0 +1,105 @@
...
+        step_size=0.5,
+    )
+    theseus_optim = th.TheseusLayer(optimizer)
+
+    # Optimize
+    theseus_inputs = {
+        "ee_pose_target": ee_pose_target.data,
+        "theta": torch.Tensor(HOME_POSE).unsqueeze(0),
+    }
+    with torch.no_grad():
```
mhmukadam, line 93: We should add another test to check backwards as well.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -0,0 +1,105 @@
...
+    return SE3(
+        x_y_z_quaternion=torch.cat([ee_pos_target, ee_quat_target]),
+        name="ee_pose_target",
+    )
+
+
+def test_ik_optimization(robot_model, ee_pose_target):
+    """Sets up inverse kinematics as an optimization problem that uses forward kinematics"""
+    # Initialize variables (joint states)
+    theta = th.Vector(NUM_DOFS, name="theta")
```
mhmukadam, line 60: Possibly switch this to a batched version?

mhmukadam (commented):
Looks great so far!

theseus/embodied/kinematics/robot_model.py
```diff
@@ -40,3 +41,34 @@ def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]
...
+        self.drm_model = drm.DifferentiableRobotModel(urdf_path)
+
+    def forward_kinematics(self, joint_states: RobotModelInput) -> Dict[str, LieGroup]:
+        """Computes forward kinematics
+        Args:
+            joint_states: Vector of all joint angles
+        Outputs:
+             Dictionary that maps link name to link pose
+
+        (Note: differentiable robot model uses the xyzw quaternion convention,
```
luisenp, line 57: Maybe this comment about xyzw --> wxyz should be in the code (around line 67)? Note that from the point of view of a user reading this docstring in the documentation, this conversion happens under the hood.

theseus/embodied/kinematics/tests/data/panda_fk_dataset.json
```diff
@@ -0,0 +1,13 @@
...
+        [2.3248593412410172, 1.6994167497883723, -0.6916813389840697, -0.5851990245348841, 0.17146823279768597, 2.1335270160528377, 2.5255755709563092],
+        [-2.545557823442099, -1.513253690733913, -2.847119985909619, -0.453153780710263, 1.6506377914866026, 3.314099494785764, 2.840216966477668]
+    ],
+    "ee_poses": [
+        [[0.5618914365768433, 0.5131404399871826, 0.34427565336227417], [0.7712864279747009, 0.6026210188865662, 0.19384866952896118, -0.06624171137809753]],
+        [[-0.3478390574455261, 0.671889066696167, -0.023775160312652588], [0.952156126499176, -0.1261112540960312, 0.22165635228157043, -0.16841386258602142]],
+        [[0.6701399683952332, 0.41195011138916016, 0.1345270723104477], [0.4187312126159668, 0.7321641445159912, 0.47301992774009705, 0.2546604871749878]]
+    ],
+    "ee_name": "panda_virtual_ee_link"
+}
```
luisenp, line 13: nit: add missing newline at end of file.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -0,0 +1,105 @@
...
+            optimizer_kwargs={
+                "track_best_solution": True,
+                "verbose": True,
+                "track_error_history": True,
+                "damping": 0.1,
+            },
+        )
+
+    # Check result
+    assert info.last_err < 1e-2
```
luisenp, line 105: What's the initial error for this test?

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -0,0 +1,105 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import os
+
+import numpy as np
+import pytest
+import torch
+
+import theseus as th
+from theseus.embodied.kinematics import UrdfRobotModel
```
luisenp, line 13: By convention, in our unit tests we only import `theseus as th` and access all classes/functions in `embodied`  via `th.eb.XXXX`. See [example](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/collision/tests/test_signed_distance_field.py#L24). 

We should add `UrdfRobotModel` to [this](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/__init__.py) file and do the same here.

luisenp (changes requested):
LGTM, great job! Only left a few comments that should be easy to address.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
exhaustin: 0.29 at iteration 0
4.632548734662123e-05 after 15 iterations

Maybe I should decrease the error tolerance to something lower like 1e-4?

theseus/embodied/kinematics/robot_model.py
exhaustin: I agree that this is quite inefficient for now especially for models with a lot of links, as this is the only API for forward kinematics DRM provides. I will prefer to refactor this on the DRM side, and create another PR in Theseus to adapt.
Let me know if there are any timeline dependencies for this task!

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
exhaustin: By a batched version, do you mean optimizing for multiple output pose targets at the same time (and thus ee_pose_target would have shape `batch_size x num_dofs`)?

theseus/embodied/kinematics/robot_model.py
exhaustin: Added issue #123 to track this.

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -73,6 +75,7 @@ def forward_kinematics(self, joint_states: RobotModelInput) -> Dict[str, LieGrou
                 joint_states, link_name
             )
             quat_processed = self._postprocess_quaternion(quat)
+            print("QUAT: ", quat_processed.shape, torch.linalg.norm(quat_processed))
```
mhmukadam: Remove debugging print?

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -10,7 +10,6 @@
 import torch
 
 import theseus as th
-from theseus.embodied.kinematics import UrdfRobotModel
 from theseus.geometry import SE3, SO3
```
mhmukadam, line 13: You won't need to import these; they should be available as th.SE3 and th.SO3.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
mhmukadam: I mean optimizing with say N inputs to go to N corresponding output pose targets.

mhmukadam (commented):
Only thing remaining is testing batching and backwards.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -10,7 +10,6 @@
 import torch
 
 import theseus as th
-from theseus.geometry import SE3, SO3
```
mhmukadam: Removing this was okay given our convention.

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -47,25 +47,35 @@ class UrdfRobotModel(KinematicsModel):
...
             joint_states: Vector of all joint angles
         Outputs:
-             Dictionary that maps link name to link pose
-
-        (Note: differentiable robot model uses the xyzw quaternion convention,
-        while theseus uses wxyz, hence the conversion.)
+            Dictionary that maps link name to link pose
         """
+        # Check input dimensions
+        assert joint_states.shape[-1] == len(self.drm_model.get_joint_limits())
```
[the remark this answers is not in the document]
exhaustin: We will have to convert the input to torch.Tensor to query DRM anyways, it feels weird to be converting an input torch.Tensor to th.Vector, then converting it back again.
Also the current check only checks the dimension of the joint space, instead of checking for batch dims.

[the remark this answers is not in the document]
mhmukadam: Ah makes sense then to not convert back and forth. In that case does it make sense to check for batch dim or is that handled inside DRM? Also instead of just assert probably raise an error with a message.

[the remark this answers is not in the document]
exhaustin: Batch dim is checked and converted within DRM :)

luisenp (commented):
Getting there! Left some comments, the major ones being about the jacobians API (to consider for once all unit test pass).

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -47,25 +48,50 @@ class UrdfRobotModel(KinematicsModel):
...
+        assert joint_states.shape[-1] == robot_model_dofs, (
+            f"Robot model dofs ({robot_model_dofs}) incompatible with "
+            f"input joint state dimensions ({joint_states.shape[-1]})."
+        )
+
+        # Parse input
+        print(type(joint_states))
+        if isinstance(joint_states, torch.Tensor):
+            joint_states_input = joint_states
+        elif isinstance(joint_states, Variable):
```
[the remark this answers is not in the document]
exhaustin: I ran into an issue where during `theseus_layer.forward()`, the optimizer passes in vectors in the form of `th.Variable` primitive type sometimes, thus triggering the invalid input exception.
Maybe this is worth investigating in more depth, but checking for th.Variable alleviates the issue.

theseus/embodied/kinematics/tests/test_urdf_model.py
```diff
@@ -39,21 +56,52 @@ def test_forward_kinematics_seq(robot_model, dataset):
...
+            return ee_se3_target.local(ee_se3_output)
+
+        # Compute autograd jacobian (autograd joint-to-cost jacobian)
+        jacobian_autograd = torch.autograd.functional.jacobian(
+            fk_func, joint_state
+        ).squeeze()
+
+        # Compute analytical jacobian
+        # (analytical manipulator jacobian * autograd ee-to-cost jacobian)
+        jacobian_m_analytical = torch.cat(
```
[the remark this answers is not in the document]
exhaustin: Directly computing the jacobian of `local` yields the jacobian in the local frame instead of the world frame. The current implementation is from a discussion with Taosha, but is still not working out currently. I am still looking into ways to debug this and make the math work.

The jacobian implementation shouldn't be too hard to do.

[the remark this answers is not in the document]
mhmukadam: Ah good idea on adding native (optional) Jacobain call to `forward_kinematics` like we have for other functions.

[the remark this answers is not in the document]
exhaustin: Jacobian tests now working

requirements/main.txt
```diff
@@ -1,3 +1,4 @@
+differentiable-robot-model>=0.2.1
```
mhmukadam, line 1: Is this the latest version where you added the efficient computation of manipulator jacobian?

theseus/embodied/__init__.py
```diff
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from .collision import Collision2D, EffectorObjectContactPlanar, SignedDistanceField2D
+from .kinematics import UrdfRobotModel
```
mhmukadam, line 7: -> `from .kinematics import KinematicsModel, IdentityModel, UrdfRobotModel`

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -0,0 +1,112 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import abc
+from typing import Dict, Optional, Union
+
+import differentiable_robot_model as drm
+import torch
+
+from theseus import Variable
```
mhmukadam, line 12: -> `from theseus.core import Variable`

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -0,0 +1,112 @@
...
+    def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]:
+        if isinstance(robot_pose, Point2) or isinstance(robot_pose, Vector):
+            assert robot_pose.dof() == 2
+            return {"state": robot_pose}
+        raise NotImplementedError(
+            f"IdentityModel not implemented for pose with type {type(robot_pose)}."
+        )
+
+    def dim(self) -> int:
+        return 1
```
mhmukadam, line 44: 1 -> 2?

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -0,0 +1,112 @@
...
+        quat_normalized = quat_converted / torch.linalg.norm(
+            quat_converted, dim=-1, keepdim=True
+        )
+
+        return quat_normalized
+
+    def forward_kinematics(
+        self,
+        joint_states: RobotModelInput,
+        jacobians: Dict[str, Optional[torch.Tensor]] = {},
```
mhmukadam, line 65: The `Optional` on tensor means we just pass a string, but then where is the computed jacobian stored?

We might want to add `Optional[Dict[...` at the start so we can then avoid computing jacobians when not needed and be more efficient.

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -0,0 +1,112 @@
...
+                joint_states_input, link_name
+            )
+            quat_processed = self._postprocess_quaternion(quat)
+
+            link_poses[link_name] = SE3(
+                x_y_z_quaternion=torch.cat([pos, quat_processed], dim=-1)
+            )
+
+        # Compute jacobians
+        for link_name in jacobians.keys():
```
mhmukadam, line 103: We should add a `None` check and skip computing jacobians when not needed to be more efficient.

mhmukadam: For consistency we could add an identity jacobian to be returned for the `IdentityModel` and add the jacobians input signature to the base class.

requirements/main.txt
exhaustin: I only added a more efficient version of computing forward kinematics for all joints. Once that PR is merged into differentiable-robot-model I will make a separate PR to adopt it on the Theseus side.

theseus/embodied/kinematics/kinematics_model.py
exhaustin: My understanding of the example Luis provided is that we'll be passing in a dictionary of link names with values being None, and the `forward_kinematics` function will then populate those values with the actual jacobian. 
Note that I also plan to update DRM to compute jacobians for all links in a more efficient way, but I feel like that is lower pri because 1) not all usecases of forward kinematics uses jacobian computation; 2) I imagine it will be a rare case where users will need access to the Jacobians of all links.

theseus/embodied/kinematics/kinematics_model.py
exhaustin: The idea is to only pass in a dictionary containing only keys (link_names) whose jacobians are desired.

theseus/embodied/kinematics/kinematics_model.py
exhaustin: This object was implemented by @luisenp, can you confirm that the return value should be 2 insteaf?

theseus/embodied/kinematics/kinematics_model.py
exhaustin: @luisenp Can you confirm/clarify what the exact interface should be? Thanks!

requirements/main.txt
exhaustin: So to explicitly answer your question, this is still the inefficient version.

theseus/embodied/kinematics/tests/test_urdf_model.py
```diff
@@ -0,0 +1,107 @@
...
+    ee_se3_target = th.SE3(x_y_z_quaternion=dataset["ee_poses"])
+    ee_se3_computed = robot_model.forward_kinematics(dataset["joint_states"])[ee_name]
+
+    assert torch.allclose(
+        ee_se3_target.local(ee_se3_computed),
+        torch.zeros(dataset["num_data"], 6),
+    )
+
+
+def test_jacobian(robot_model, dataset):
```
mhmukadam, line 77: Should we try a batched version of this?

mhmukadam (commented):
Looks great. Nice work on the batched IK test! Just some minor comments and then we should be able to merge.

theseus/embodied/kinematics/kinematics_model.py
mhmukadam: That makes sense. In other places [like here](https://github.com/facebookresearch/theseus/pull/84/files#diff-0579a4f9834772c89a164d4e1f36069e7eab1168c2ce477c109be11c31c024cdR86-R98) we optionally pass a tensor for a jacobian to be computed. Luis can confirm if we want to follow a similar API here.

mhmukadam: On a closer look, I think the current version is similar to api elsewhere the difference is the Dict.

luisenp: What I had in mind was `jacobians: Optional[Dict[str, torch.Tensor]] = None`, and the function will only populate jacobians if the dict is not `None` and only for any keys that are present in the dict. 

PS: BTW, it's good to avoid passing mutable objects as defaults (e.g., `{}`) because these defaults are actually associated to the class, not the object.

theseus/embodied/kinematics/kinematics_model.py
mhmukadam: I see then this is implicitly handled by passing only the required link names.

theseus/embodied/kinematics/kinematics_model.py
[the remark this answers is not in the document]
luisenp: Is it easy for you to paste a screen shot of the error message you were getting?

theseus/embodied/kinematics/kinematics_model.py
luisenp: I think it should be 2, but this was set to 1 quite a long time ago, it seems. I don't remember why, but maybe it was to do [this](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/collision/collision.py#L73), which seems incorrect? As in, collision dim should just be hardcoded to 1?

BTW, do we actually need this dim function?

mhmukadam: Ah for now hardcoding collision dim to one is okay when it is identity model. Otherwise it should be the number of body points you are checking collision for.

theseus/embodied/kinematics/kinematics_model.py
mhmukadam: Good catch on passing mutable objects, this can actually introduce some hard to find bugs.

theseus/embodied/kinematics/kinematics_model.py
luisenp: Then in this case it seems that `dim()` for kinematics models is actually representing the number of body points? Should we rename this method?

theseus/embodied/kinematics/tests/test_urdf_model.py
exhaustin: Added

theseus/embodied/kinematics/kinematics_model.py
exhaustin: Should the values also be `Optional` since we might be passing in {ee_name: None}?
We could also just do `jacobians: Optional[Dict[str, Any]] = None` since the values are going to be overwritten anyways.

theseus/embodied/kinematics/kinematics_model.py
mhmukadam: Not sure yet how we want to add body points, inside kinematics class or as separate struct/class. Either way it should be a property related to that. If that was the only use for dim then maybe we can remove it for now.

theseus/embodied/kinematics/kinematics_model.py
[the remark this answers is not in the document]
exhaustin: image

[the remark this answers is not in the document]
exhaustin: I never ran the script involving `IdentityModel`, do you think you'll be able to test it out within the collision stuff?

theseus/embodied/kinematics/kinematics_model.py
exhaustin: Should I simply remove the dim method from `IdentityModel` and `UrdfRobotModel`, and also downstream methods in `Collision2D` and `EffectorObjectContactPlanar` that uses it?

theseus/embodied/kinematics/kinematics_model.py
luisenp: Ah, good point! It feels a bit strange to me to pass `None` values. Would it be better to do the following?

```python
def forward_kinematics(
        self,
        joint_states: RobotModelInput,
        jacobians: Optional[Dict[str, torch.Tensor]] = None,
        jacobians_links: Optional[List[str]] = None, 
)
```

If the `jacobians_links` list is not passed, then jacobians for all the links are computed. Any thoughts?

mhmukadam: You would still need to then make a Dict with all the link names for jacobians, which seems redundant.

`Optional[Dict[str, Optional[torch.Tensor]]] = None` seems like the cleanest option to me. Then a user can also pass `jacobians = dict.fromkeys(['link1', 'link2'])` with `None` in the values.

theseus/embodied/kinematics/kinematics_model.py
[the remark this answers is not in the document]
luisenp: Added a fix in #155.

theseus/embodied/kinematics/kinematics_model.py
luisenp: Removed and tested in #156

PR #87, opened 2022-02-18
Enabled back ellipsoidal damping in LM with linear solvers support checks

### LINKED ISSUE #35: Ellipsoidal damping for SparseSolver
`A' = A + damping * diag(A + eps)`

luisenp:
## Motivation and Context
Closes #35 

## Types of changes

## Checklist

### DIFF
--- a/theseus/optimizer/nonlinear/levenberg_marquardt.py
+++ b/theseus/optimizer/nonlinear/levenberg_marquardt.py
@@ -13,12 +13,12 @@ from theseus.optimizer.linear import DenseSolver, LinearSolver, LUCudaSparseSolv
 
 from .nonlinear_least_squares import NonlinearLeastSquares
 
-_LM_ALLOWED_SOLVERS = [DenseSolver, LUCudaSparseSolver]
+_LM_ALLOWED_ELLIPS_DAMP_SOLVERS = [DenseSolver, LUCudaSparseSolver]
 
 
 def _check_ellipsoidal_damping_cls(linear_solver: LinearSolver):
     good = False
-    for lsc in _LM_ALLOWED_SOLVERS:
+    for lsc in _LM_ALLOWED_ELLIPS_DAMP_SOLVERS:
         if isinstance(linear_solver, lsc):
             good = True
     return good
@@ -60,7 +60,7 @@ class LevenbergMarquardt(NonlinearLeastSquares):
         **kwargs,
     ) -> torch.Tensor:
 
-        solvers_str = ",".join(c.__name__ for c in _LM_ALLOWED_SOLVERS)
+        solvers_str = ",".join(c.__name__ for c in _LM_ALLOWED_ELLIPS_DAMP_SOLVERS)
         if ellipsoidal_damping and not self._allows_ellipsoidal:
             raise NotImplementedError(
                 f"Ellipsoidal damping is only supported by solvers with type "
--- a/theseus/optimizer/nonlinear/levenberg_marquardt.py
+++ b/theseus/optimizer/nonlinear/levenberg_marquardt.py
@@ -9,9 +9,19 @@ import torch
 
 from theseus.core import Objective
 from theseus.optimizer import Linearization
-from theseus.optimizer.linear import DenseSolver, LinearSolver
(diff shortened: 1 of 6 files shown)


theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -9,10 +9,20 @@
 
 from theseus.core import Objective
 from theseus.optimizer import Linearization
-from theseus.optimizer.linear import DenseSolver, LinearSolver
+from theseus.optimizer.linear import DenseSolver, LinearSolver, LUCudaSparseSolver
 
 from .nonlinear_least_squares import NonlinearLeastSquares
 
+_LM_ALLOWED_SOLVERS = [DenseSolver, LUCudaSparseSolver]
```
mhmukadam, line 16: Might be confusing name if you intend to use this for ellipsoidal damping and the list is different for default/spherical damping.

mhmukadam (approved):
Looks good, CI is failing though.

theseus/optimizer/nonlinear/levenberg_marquardt.py
luisenp: Good point!

luisenp:
CI is failing due to #88, which I assigned to @maurimo. Forgot to cross link here.

theseus/optimizer/nonlinear/levenberg_marquardt.py
luisenp: Fixed.

theseus/optimizer/nonlinear/tests/test_levenberg_marquardt.py
```diff
@@ -22,3 +32,34 @@ def test_levenberg_marquartd():
...
+    mock_objective.to(device="cuda", dtype=torch.double)
+    mock_objective.update(
+        {
+            "v1": torch.ones(1, 1, device="cuda", dtype=torch.double),
+            "v2": torch.zeros(1, 1, device="cuda", dtype=torch.double),
+        }
+    )
+    for lsc in [th.LUCudaSparseSolver]:
+        optimizer = th.LevenbergMarquardt(mock_objective, lsc)
+        optimizer.linear_solver.reset(batch_size=1)
```
luisenp, line 63: Somehow I had missed that we need to call `solver.reset(batch_size=)` on `LUCudaSparseSolver` before running optimization. I think we want to hide this under the hood, but not sure what the best place to put this would be. @maurimo, Is this a function that can be called with different `batch_size` values every time? Would it make sense to have `NonlinearOptimizer.optimize()` always call this  function at the beginning (likewise for `LinearOptimizer.optimize()`.

mhmukadam (approved):
Look good, can merge after the one open comment is resolved.

theseus/optimizer/nonlinear/tests/test_levenberg_marquardt.py
maurimo: "reset is not necessary if the Solver is created with a valid linearization (if so reset is called automatically in the constructor, reset is available for tests and because of the limitations of Cuda api).
Keep in mind that every time reset is called the symbolic decomposition is recomputed. This is consequence of the limitations of the CuSolver-Refactor API, additionally to the issue of symbolic/numeric decompositions being the same the batch size is not flexible therefore it's not possible to create the sparse solver without knowing the size of the batch beforehand.
Therefore I would call "reset" only if it's necessary to change the batch size (or the number of contexts).

maurimo: "reset is not necessary if the Solver is created with a valid linearization (if so reset is called automatically in the constructor, reset is available for tests and because of the limitations of Cuda api).
Keep in mind that every time reset is called the symbolic decomposition is recomputed. This is consequence of the limitations of the CuSolver-Refactor API, additionally to the issue of symbolic/numeric decompositions being the same the batch size is not flexible therefore it's not possible to create the sparse solver without knowing the size of the batch beforehand.
Therefore I would call "reset" only if it's necessary to change the batch size (or the number of contexts).

maurimo: looks good to me!

PR #94, opened 2022-02-25
Change RobotModel.forward_kinematics() interface

luisenp:
## Motivation and Context
As discussed in #84

### DIFF
--- a/theseus/embodied/kinematics/robot_model.py
+++ b/theseus/embodied/kinematics/robot_model.py
@@ -4,10 +4,14 @@
 # LICENSE file in the root directory of this source tree.
 
 import abc
+from typing import Dict, Union
 
 import torch
 
 from theseus.geometry import SE2, LieGroup, Point2, Vector
+
+RobotModelInput = Union[torch.Tensor, Vector]
+RobotModelOuput = Union[LieGroup, Dict[str, LieGroup]]
 
 
 class RobotModel(abc.ABC):
@@ -15,7 +19,7 @@ class RobotModel(abc.ABC):
         pass
 
     @abc.abstractmethod
-    def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
+    def forward_kinematics(self, robot_pose: RobotModelInput) -> RobotModelOuput:
         pass
 
     @abc.abstractmethod
@@ -27,12 +31,12 @@ class IdentityModel(RobotModel):
     def __init__(self):
         super().__init__()
 
-    def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
+    def forward_kinematics(self, robot_pose: RobotModelInput) -> RobotModelOuput:
         if isinstance(robot_pose, SE2):
-            return robot_pose.translation.data.view(-1, 2, 1)
+            return robot_pose.translation
         if isinstance(robot_pose, Point2) or isinstance(robot_pose, Vector):
             assert robot_pose.dof() == 2
-            return robot_pose.data.view(-1, 2, 1)
+            return robot_pose
         raise NotImplementedError(
             f"IdentityModel not implemented for pose with type {type(robot_pose)}."
         )
--- a/theseus/embodied/kinematics/robot_model.py
+++ b/theseus/embodied/kinematics/robot_model.py
@@ -11,7 +11,6 @@ import torch
 from theseus.geometry import LieGroup, Point2, Vector
 
 RobotModelInput = Union[torch.Tensor, Vector]
-RobotModelOuput = Union[LieGroup, Dict[str, LieGroup]]
 
 
 class RobotModel(abc.ABC):
@@ -19,7 +18,7 @@ class RobotModel(abc.ABC):
         pass
 
     @abc.abstractmethod
-    def forward_kinematics(self, robot_pose: RobotModelInput) -> RobotModelOuput:
+    def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]:
         pass
 
     @abc.abstractmethod
@@ -31,10 +30,10 @@ class IdentityModel(RobotModel):
     def __init__(self):
         super().__init__()
 
-    def forward_kinematics(self, robot_pose: RobotModelInput) -> RobotModelOuput:
+    def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]:
         if isinstance(robot_pose, Point2) or isinstance(robot_pose, Vector):
             assert robot_pose.dof() == 2
-            return robot_pose
+            return {"state": robot_pose}
         raise NotImplementedError(
             f"IdentityModel not implemented for pose with type {type(robot_pose)}."
         )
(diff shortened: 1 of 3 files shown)


mhmukadam (approved):

exhaustin:
I think the typing makes sense given our current context. 
If it's okay with you, I would like to gain more understanding by prototyping the inverse kinematics test first, then return and confirm the changes in this PR.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -4,18 +4,22 @@
...
 
 import abc
+from typing import Dict, Union
 
 import torch
 
 from theseus.geometry import SE2, LieGroup, Point2, Vector
 
+RobotModelInput = Union[torch.Tensor, Vector]
+RobotModelOuput = Union[LieGroup, Dict[str, LieGroup]]
```
exhaustin, line 14: I think `RobotModelInput` makes  sense as torch tensors and theseus vectors are effectively two representations of the same concept.
`RobotModelOutput` can be trouble some as the user can be unsure what to expect. I'm thinking maybe we can have two different FK implementations, a `forward_kinematics_to_link(RobotModelInput, str) -> LieGroup` that computes pose of a specified link, and a `forward_kinematics_all_links(RobotModelInput) -> Dict[str, LieGroup]` that computes poses of all links in the kinematics tree.
Do you think that makes sense? Would be great to have @mhmukadam 's input on this.

theseus/embodied/kinematics/robot_model.py
```diff
@@ -27,12 +31,12 @@ class IdentityModel(RobotModel):
     def __init__(self):
         super().__init__()
 
-    def forward_kinematics(self, robot_pose: LieGroup) -> torch.Tensor:
+    def forward_kinematics(self, robot_pose: RobotModelInput) -> RobotModelOuput:
         if isinstance(robot_pose, SE2):
-            return robot_pose.translation.data.view(-1, 2, 1)
+            return robot_pose.translation
```
exhaustin, line 36: Should we remove this condition since the input will not be SE2?

theseus/embodied/kinematics/robot_model.py
```diff
@@ -27,12 +31,12 @@ class IdentityModel(RobotModel):
...
         if isinstance(robot_pose, Point2) or isinstance(robot_pose, Vector):
             assert robot_pose.dof() == 2
-            return robot_pose.data.view(-1, 2, 1)
+            return robot_pose
```
exhaustin, line 39: The output should be a LieGroup, but robot_pose is a `Point2` or a `Vector` here?

exhaustin (commented):
Added some comments after going through the IK test.

theseus/embodied/kinematics/robot_model.py
mhmukadam: `Point2` is derived from `Vector` which is derived from `LieGroup` :)

theseus/embodied/kinematics/robot_model.py
mhmukadam: > as the user can be unsure what to expect

Good point! Although the long function names would be tedious to use too. Here are two options:
1. We just use `Union[LieGroup, Dict[str, LieGroup]]` and not worry about `RobotModelOuput `
2. We have a wrapper function `forward_kinematics()` that internally calls either `..._to_link()` or `..._all_links()`

exhaustin: I don't know what robot models outside of URDFs might be implemented in Theseus in the future, but I would expect most forward kinematics computations to be specified by an URDF file, thus having named links.
I'm actually thinking that if there isn't a second use case, maybe we shouldn't over-modularize things and maintain only one RobotModel class?

theseus/embodied/kinematics/robot_model.py
exhaustin: Ahh I wasn't aware of that. Thanks!

theseus/embodied/kinematics/robot_model.py
luisenp: @mhmukadam Any thoughts? I don't have a too strong opinion on this, but I'm hesitant about having a single `RobotModel` class that's hard wired to a specific format (URDF in this case). 

How about keeping the current class hierarchy but just have the output be a plain `Dict[str, LieGroup]` and revisit this later if needed?

mhmukadam: >  just have the output be a plain `Dict[str, LieGroup]` and revisit this later if needed?

This sounds good, and the hierarchy in Austin's PR: `KinematicsModel` -> `UrdfRobotModel`
I imagine human models, hand models, etc can later derive from `KinematicsModel`.

theseus/embodied/kinematics/robot_model.py
luisenp: Fixed.

mhmukadam: Actually a question for @exhaustin: how do you handle an input space in DRM for a mobile manipulator which would be base + arm (SE2 x Vector)?

luisenp: Regarding `Point2`, we currently allow `Vector` inputs to the robot model.

mhmukadam: Ah yes, but when orientation is involved we would need SE2. We don't need to have all the variants implemented but thinking about the API would be useful so we don't have to refactor later. Here are some possible types of robot bodies with their input space:
- Mobile base: SE2
- Arm(N): Vector(N) [we are currently focusing on this]
- Arm(N) + hand(H): Vector(N+H)
- Mobile base and arm: SE2 x Vector(N)
- Mobile base, vertical actuator,  arm: SE2 x Vector(N+1)
- Mobile base, vertical actuator,  left arm, right arm : SE2 x Vector(2N+1)
- Aerial robot: SE3
- Aerial robot + arm: SE3 x Vector(N)
- more combinations for legged robots or humanoids or robots with a camera that have a pan/tilt joint.

exhaustin: > Actually a question for @exhaustin: how do you handle an input space in DRM for a mobile manipulator which would be base + arm (SE2 x Vector)?

Simulators and kinematics libraries I've worked with seems to be designed for rotary joints (for example, the PyBullet API `setMotorControlArray` makes the assumption that all actuators are rotary motors). I don't have a clear idea of how we could generalize to structured joints that aren't just an array of joint angles, and I can't find any examples out there.

If I imagine how I would set up a problem on mobile manipulators, I would set up the forward kinematics separately as the transform of the mobile base + the transform of the manipulator computed using the urdf. I think this could exist in user code instead of having a single forward kinematics function that encapsulates all special cases you listed above.

mhmukadam: That makes sense, base transform + FK is the version I am familiar with as well. Let's proceed with the version where URDFRobotModel take a Vector as an input that represents the joint angles of a kinematic chain.

An aside question: does DRM current support branching chains for robot hand fingers? (I don't think this would affect the input since it would still be a Vector)

exhaustin: Yes we support kinematic trees as well as chains. The input is still a vector of actuated joints, in the order they are defined in the URDF file.

Issue #78, opened 2022-02-09
Move random LieGroup creation from tests to theseus/utils


luisenp:
We could move more generally useful stuff like create_random_se2 from tests to theseus/utils.

_Originally posted by @mhmukadam in https://github.com/facebookresearch/theseus/pull/75#r801535776_

fantaosha:
Closing this after merging #95.

PR #95, opened 2022-02-25
Add rand() to LieGroup

fantaosha:
## Motivation and Context

See #78.

## Types of changes

## Checklist


theseus/geometry/lie_group.py
```diff
@@ -44,6 +45,17 @@ def _init_data(*args: Any) -> torch.Tensor:
     def dof(self) -> int:
         pass
 
+    @staticmethod
+    @abc.abstractmethod
+    def rand(
+        *size: int,
```
mhmukadam, line 51: Is `size` meant for batch size? If so we should rename and put `1` as default.

theseus/geometry/tests/test_se2.py
```diff
@@ -47,31 +47,31 @@ def test_compose():
     rng = torch.Generator()
     rng.manual_seed(0)
     for batch_size in [1, 20, 100]:
-        se2_1 = create_random_se2(batch_size, rng)
-        se2_2 = create_random_se2(batch_size, rng)
+        se2_1 = th.SE2.rand(batch_size, generator=rng, dtype=torch.float64)
```
mhmukadam, line 50: Since these tests now depend on the `rand` wrapper function should we add a test for the `rand` function itself first?

theseus/geometry/lie_group.py
```diff
@@ -44,6 +45,17 @@ def _init_data(*args: Any) -> torch.Tensor:
...
+        generator: Optional[torch.Generator] = None,
```
mhmukadam, line 52: Can this be `Optional`, what's the behavior when it is left as `None`?

theseus/geometry/vector.py
```diff
@@ -42,6 +43,28 @@ def _init_data(dof: int) -> torch.Tensor:  # type: ignore
     def dof(self) -> int:
         return self.data.shape[1]
 
+    @staticmethod
+    def rand(
+        *size: int,
```
mhmukadam, line 48: In reference to the batch size comment above, it could get confusing here if we need both the batch and vector size.

mhmukadam (approved):
Looks good, a useful utility to absorb in the class!

theseus/geometry/lie_group.py
fantaosha: `size` should be a `tuple` rather than `int`. The `inputs` of our `rand()` functions are consistent with [`torch.rand()`](https://pytorch.org/docs/stable/generated/torch.rand.html), which has no default values for `size`.

theseus/geometry/tests/test_se2.py
fantaosha: Well, I can not think of a way to test these `rand()` functions.

theseus/geometry/vector.py
fantaosha: `*size` here is more than a `batch size`. For `theseus.Vector`, `*ize` could be both the `batch size` and `vector dimensions`, whereas for `theseus.SO3`, `size` only refers to the `batch size`.

theseus/geometry/lie_group.py
fantaosha: It will use the default generator for `torch.rand()`.

theseus/geometry/vector.py
mhmukadam: Okay thanks for clarifying. I think once we add documentation this stuff will be more obvious.

examples/se2_inverse.py
```diff
@@ -12,12 +12,11 @@
 from theseus import LieGroupTensor
 from theseus.geometry.lie_group import LieGroup
 
-from theseus.geometry.tests.test_se2 import create_random_se2
 
 rng = torch.Generator()
 rng.manual_seed(0)
-x1 = create_random_se2(1, rng)
-x2 = create_random_se2(1, rng)
+x1 = th.SE2.rand(1, generator=rng)
```
luisenp, line 18: Let's make aliases for all of these, as in `th.randse2()`, `th.randse3()`, `th.randpoint2()`, etc.

theseus/geometry/se2.py
```diff
@@ -36,6 +37,32 @@ def __init__(
             translation = Point2(data=x_y_theta[:, :2])
             self.update_from_rot_and_trans(rotation, translation)
 
+    @staticmethod
+    def rand(
+        *size: int,
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
```
luisenp, line 46: `requires_grad: bool = False`

theseus/geometry/se3.py
```diff
@@ -33,6 +34,34 @@ def __init__(
         if x_y_z_quaternion is not None:
             self.update_from_x_y_z_quaternion(x_y_z_quaternion=x_y_z_quaternion)
 
+    @staticmethod
+    def rand(
+        *size: int,
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
```
luisenp, line 43: `requires_grad: bool = False`

theseus/geometry/so2.py
```diff
@@ -35,6 +36,30 @@ def __init__(
                 )
             self.update_from_angle(theta)
 
+    @staticmethod
+    def rand(
+        *size: int,
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
```
luisenp, line 45: `requires_grad: bool = False`

theseus/geometry/so3.py
```diff
@@ -33,6 +34,30 @@ def __init__(
         if quaternion is not None:
             self.update_from_unit_quaternion(quaternion)
 
+    @staticmethod
+    def rand(
+        *size: int,
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
```
luisenp, line 43: `requires_grad: bool = False`

theseus/geometry/tests/test_se2.py
luisenp: Could we do something like checking that the set of `sample.log_map()`  is uniformly distributed in R^dof?

theseus/geometry/vector.py
```diff
@@ -42,6 +43,28 @@ def _init_data(dof: int) -> torch.Tensor:  # type: ignore
...
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
```
luisenp, line 52: `requires_grad: bool = False`

theseus/geometry/lie_group.py
```diff
@@ -44,6 +45,17 @@ def _init_data(*args: Any) -> torch.Tensor:
     def dof(self) -> int:
         pass
 
+    @staticmethod
+    @abc.abstractmethod
+    def rand(
```
luisenp, line 50: In `torch` and `numpy`, function `rand()` has the specific meaning of sampling in the range [0, 1]. Wondering if naming this function `rand()` might be confusing to users. Should we use a different name? Maybe `random()`, `sample()`, something else?

theseus/geometry/point_types.py
```diff
@@ -62,6 +85,29 @@ def __init__(
...
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
+    ) -> "Point3":
+        if len(size) != 1:
+            raise ValueError("The size should be 1D.")
+        return Point3(
+            data=2
+            * torch.rand(
```
luisenp, line 100: Why limit to [-1, 1]?

theseus/geometry/tests/test_se2.py
fantaosha: Seems a little difficult. Maybe we can create a issue for future work?

theseus/geometry/lie_group.py
fantaosha: I don't think this is a big issue. But I do consider using `LieGroup.rand()` and `LieGroup.randn()`  for uniform and for normal distributions, respectively, which is consistent with `torch.rand()` and `torch.randn()`. Any thoughts?

theseus/geometry/point_types.py
fantaosha: No reasons. I can change it to any range.

theseus/geometry/point_types.py
```diff
@@ -37,14 +37,37 @@ def __init__(
         dof, data = _prepare_dof_and_data(2, data)
         super().__init__(dof=dof, data=data, name=name, dtype=dtype)
 
+    @staticmethod
+    def rand(
+        *size: int,
+        generator: Optional[torch.Generator] = None,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+        requires_grad: Boolean = False,
```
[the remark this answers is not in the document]
fantaosha: Fixed.

examples/se2_inverse.py
fantaosha: Do you mean we add `theseus.rand_se2()` etc. in `theseus.geometry.utils`?

theseus/geometry/se3.py
fantaosha: Fixed.

theseus/geometry/so3.py
fantaosha: Fixed.

theseus/geometry/lie_group.py
luisenp: Yeah, I actually like that suggestion of having both versions. But in that case, let's make the point types sampling consistent with what the user would expect from the torch. For example, `Point2.rand()` returns points in [0, 1].

fantaosha: > Yeah, I actually like that suggestion of having both versions. But in that case, let's make the point types sampling consistent with what the user would expect from `torch`. For example, `Point2.rand()` returns points in [0, 1].

Shall we keep the range of `theta` as `[-pi, pi]`?

theseus/geometry/tests/test_se2.py
luisenp: Actually, we could just test that these functions are nothing but wrappers by mocking `torch.rand()` with a function that returns some pre-determined set of tensors, and then check that the samples's log_map are matching those tensors.

examples/se2_inverse.py
luisenp: You can define them each in the file as. For example., in `so3.py` you can define `randnso3 = SO3.radn` at the bottom, and make them accessible at `theseus` level in the `__init__.py` files.

theseus/geometry/lie_group.py
luisenp: Yes, I meant that only for subclasses of `Vector`.

theseus/geometry/tests/test_se2.py
fantaosha: That's difficult since for `|SO3.log_map()|` is no greater to `pi` and `SO3.log_map(SO3.exp(map))` may be different.

> Actually, we could just test that these functions are nothing but wrappers by mocking `torch.rand()` with a function that returns some pre-determined set of tensors, and then check that the samples's log_map are matching those tensors.

theseus/geometry/point_types.py
luisenp: See comment in the discussion of `rand` and `randn`.

fantaosha: Change the range of `rand()` for `Vector`, `Point2`, `Point3` and `SE2` to `[0, 1]`

examples/se2_inverse.py
fantaosha: Will do.

luisenp (changes requested):
LGTM. The only thing missing are the aliases (e.g., `th.randse2`, `th.randnse3`).

examples/se2_inverse.py
fantaosha: Added.

theseus/geometry/point_types.py
```diff
@@ -161,3 +161,9 @@ def exp_map(tangent_vector: torch.Tensor) -> "Point3":
     # added to avoid casting downstream
     def copy(self, new_name: Optional[str] = None) -> "Point3":
         return cast(Point3, super().copy(new_name=new_name))
+
+
+rand_point2 = Point2.rand
```
luisenp, line 166: We need to add these to `__init__.py` all the way to the top level. A good example to follow here are the group [operations](https://github.com/facebookresearch/theseus/blob/main/theseus/__init__.py#L25).

mhmukadam (approved):
Looks good to merge once you address Luis' open comment.

PR #96, opened 2022-03-03
Fixed bug in Variable.update() that was breaking torch graph...

luisenp:
…when batch_ignore_mask.any() == True.

## Motivation and Context
This was causing backward errors in @maurimo's notebook whenever there were batch elements that converged before others. Not sure why we hadn't seen this error before, since I think some of our test had elements converging at different rates (but not 100% sure). cc: @fantaosha who was also looking into this notebook today.

## How Has This Been Tested
Maurizio's notebook now runs w/o the torch errors in backward pass. 

## Types of changes

## Checklist

### DIFF
--- a/theseus/core/variable.py
+++ b/theseus/core/variable.py
@@ -56,8 +56,8 @@ class Variable:
                 f"{self.name} has dtype {self.dtype}."
             )
         if batch_ignore_mask is not None and batch_ignore_mask.any():
-            good_indices = ~batch_ignore_mask
-            self[good_indices] = data[good_indices]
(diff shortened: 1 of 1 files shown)


mhmukadam (approved):
Nice catch! Do any of the current unit tests need to be updated or should we add a new unit test that will check this?

luisenp:
@mhmukadam I just tried to write a small test that would isolate this error but I couldn't replicate it easily. It's possible that it only happens after some interaction with the Lie Group code, which also has other in place operations. It'll be easier to try again once all the Lie Group code is merged.

PR #98, opened 2022-03-03
Add projection for sparse Jacobian matrices

fantaosha:
## Motivation and Context
Update `LieGroup.project()` for sparse jacobian matrices.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #99, opened 2022-03-03
Add LieGroup Support for AutoDiffFunction

fantaosha:
## Types of changes

## Checklist


theseus/core/tests/test_cost_function.py
```diff
@@ -185,3 +185,64 @@ def error_fn(optim_vars, aux_vars):
         weighted_error = cost_function.weighted_error()
         direct_error_computation = cost_weight_value * torch.ones(batch_size, 1)
         assert torch.allclose(weighted_error, direct_error_computation)
+
+
+def test_autodiff_cost_function_error_and_jacobians_shape_on_SO3():
```
luisenp, line 190: Do we have to test the shape (taken care already by another test) or only that the numerical value of the jacobians are correct?

luisenp (changes requested):
This LGTM. I think the test can be simplified, but otherwise this looks good.

theseus/core/tests/test_cost_function.py
fantaosha: Numerical values are tested in previous PR #98.

luisenp: If we already have other tests in place to check that sparse gradient computation is correct, here we should mostly test that the full combination of `AutodiffCostFunction` + `project` is put together properly and gives the expected result in a few test cases.

fantaosha: Added value tests.

mhmukadam (approved):

luisenp (approved):

PR #105, opened 2022-03-08
error and errorsquaredNorm optional data

jeffin07:
## Motivation and Context

API improvement in Objective: error and errorSquaredNorm can optionally take in var_data which if passed would call update internally. Document the behavior that if var_data is passed this will update the objective. #4 

## How Has This Been Tested

Testes using unit test and a custom example

## Types of changes

## Checklist

### DIFF
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -337,7 +337,14 @@ class Objective:
     def size_aux_vars(self) -> int:
         return len(self.aux_vars)
 
-    def error(self) -> torch.Tensor:
+    def error(self, input_data: Optional[Dict[str, torch.Tensor]]=None, also_update: bool=False) -> torch.Tensor:
+        old_data = {}
+        if input_data is not None:
+            if also_update == False:
+                for var in self.optim_vars:
+                    old_data[var] = self.optim_vars[var].data
+                self.update(input_data=input_data)
+            self.update(input_data=input_data)
         error_vector = torch.zeros(self.batch_size, self.dim()).to(
             device=self.device, dtype=self.dtype
         )
@@ -347,10 +354,12 @@ class Objective:
                 :, pos : pos + cost_function.dim()
             ] = cost_function.weighted_error()
             pos += cost_function.dim()
+        if old_data:
+            self.update(old_data)
         return error_vector
 
-    def error_squared_norm(self) -> torch.Tensor:
-        return (self.error() ** 2).sum(dim=1)
+    def error_squared_norm(self, input_data: Optional[Dict[str, torch.Tensor]]=None, also_update: bool=False) -> torch.Tensor:
+        return (self.error(input_data=input_data, also_update = also_update) ** 2).sum(dim=1)
 
     def copy(self) -> "Objective":
         new_objective = Objective()
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -337,13 +337,16 @@ class Objective:
     def size_aux_vars(self) -> int:
         return len(self.aux_vars)
 
-    def error(self, input_data: Optional[Dict[str, torch.Tensor]]=None, also_update: bool=False) -> torch.Tensor:
+    def error(
+        self,
+        input_data: Optional[Dict[str, torch.Tensor]] = None,
+        also_update: bool = False,
+    ) -> torch.Tensor:
         old_data = {}
         if input_data is not None:
-            if also_update == False:
+            if not also_update:
                 for var in self.optim_vars:
                     old_data[var] = self.optim_vars[var].data
-                self.update(input_data=input_data)
             self.update(input_data=input_data)
         error_vector = torch.zeros(self.batch_size, self.dim()).to(
             device=self.device, dtype=self.dtype
@@ -354,12 +357,18 @@ class Objective:
                 :, pos : pos + cost_function.dim()
             ] = cost_function.weighted_error()
             pos += cost_function.dim()
-        if old_data:
+        if not also_update:
             self.update(old_data)
         return error_vector
 
-    def error_squared_norm(self, input_data: Optional[Dict[str, torch.Tensor]]=None, also_update: bool=False) -> torch.Tensor:
-        return (self.error(input_data=input_data, also_update = also_update) ** 2).sum(dim=1)
+    def error_squared_norm(
+        self,
+        input_data: Optional[Dict[str, torch.Tensor]] = None,
+        also_update: bool = False,
+    ) -> torch.Tensor:
+        return (self.error(input_data=input_data, also_update=also_update) ** 2).sum(
+            dim=1
+        )
 
     def copy(self) -> "Objective":
         new_objective = Objective()
--- a/theseus/core/tests/test_objective.py
+++ b/theseus/core/tests/test_objective.py
@@ -286,8 +286,10 @@ def test_objective_error():
         v2_data = torch.ones(batch_size, dof) * f2 * 0.1
 
         input_data = {"v1": v1_data, "v2": v2_data}
-        error = objective.error(input_data=input_data, also_update=True)
-        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+        error = objective.error(input_data=input_data, also_update=False)
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=False
+        )
 
         _check_error_for_data(v1_data, v2_data, error, error_norm_2)
 
@@ -296,7 +298,19 @@ def test_objective_error():
 
         input_data = {"v1": v1_data, "v2": v2_data}
         error = objective.error(input_data=input_data, also_update=True)
-        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=True
+        )
+
+        # To check if variables are updated
+        assert (
+            objective.optim_vars["v1"].data.numpy().all()
+            == input_data["v1"].numpy().all()
+        )
+        assert (
+            objective.optim_vars["v2"].data.numpy().all()
+            == input_data["v2"].numpy().all()
+        )
 
         _check_error_for_data(v1_data, v2_data, error, error_norm_2)
 
--- a/theseus/core/tests/test_objective.py
+++ b/theseus/core/tests/test_objective.py
@@ -292,8 +292,8 @@ def test_objective_error():
         )
 
         # To check if variables are not updated
-        assert not torch.equal(objective.optim_vars["v1"].data, input_data["v1"])
-        assert not torch.equal(objective.optim_vars["v2"].data, input_data["v2"])
+        assert objective.optim_vars["v1"].data is not input_data["v1"]
+        assert objective.optim_vars["v2"].data is not input_data["v2"]
 
         _check_error_for_data(v1_data, v2_data, error, error_norm_2)
 
@@ -307,8 +307,8 @@ def test_objective_error():
         )
 
         # To check if variables are updated
-        assert torch.equal(objective.optim_vars["v1"].data, input_data["v1"])
-        assert torch.equal(objective.optim_vars["v2"].data, input_data["v2"])
+        assert objective.optim_vars["v1"].data is input_data["v1"]
+        assert objective.optim_vars["v2"].data is input_data["v2"]
 
         _check_error_for_data(v1_data, v2_data, error, error_norm_2)
 
--- a/theseus/core/tests/test_objective.py
+++ b/theseus/core/tests/test_objective.py
@@ -280,6 +280,26 @@ def test_objective_error():
         assert error.shape == (batch_size, 2 * dof)
         _check_error_for_data(v1_data, v2_data, error, error_norm_2)
 
+        # For input_data and also_update
+
+        v1_data = torch.ones(batch_size, dof) * f1 * 0.1
+        v2_data = torch.ones(batch_size, dof) * f2 * 0.1
+
+        input_data = {"v1": v1_data, "v2": v2_data}
+        error = objective.error(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+
+        _check_error_for_data(v1_data, v2_data, error, error_norm_2)
+
+        v1_data = torch.ones(batch_size, dof) * f1 * 0.4
+        v2_data = torch.ones(batch_size, dof) * f2 * 0.4
+
+        input_data = {"v1": v1_data, "v2": v2_data}
+        error = objective.error(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+
+        _check_error_for_data(v1_data, v2_data, error, error_norm_2)
+
 
 def test_get_cost_functions_connected_to_vars():
     (
--- a/theseus/core/tests/test_objective.py
+++ b/theseus/core/tests/test_objective.py
@@ -256,11 +256,6 @@ def test_objective_error():
         else:
             assert error_.allclose(expected_error.norm(dim=1) ** 2)
 
-    # def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
-    #     expected_error = torch.cat([v1_data_, v2_data_], dim=1) * w
-    #     assert error_.allclose(expected_error)
-    #     assert error_norm_2_.allclose(expected_error.norm(dim=1) ** 2)
-
     def _check_variables(objective, input_data, v1_data, v2_data, also_update):
 
         if also_update:
@@ -273,7 +268,7 @@ def test_objective_error():
             assert objective.optim_vars["v1"].data is v1_data
             assert objective.optim_vars["v2"].data is v2_data
 
-    def _check_error(
+    def _check_error_and_variables(
         v1_data_, v2_data_, error_, error_type, objective, input_data, also_update
     ):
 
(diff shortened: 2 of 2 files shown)


luisenp:
Thanks for putting up the PR, @jeffin07. BTW, have you installed the pre-commit hooks? See instructions [here](https://github.com/facebookresearch/theseus/blob/main/CONTRIBUTING.md#code-style).

theseus/core/objective.py
```diff
@@ -337,7 +337,14 @@ def size_variables(self) -> int:
     def size_aux_vars(self) -> int:
         return len(self.aux_vars)
 
-    def error(self) -> torch.Tensor:
+    def error(self, input_data: Optional[Dict[str, torch.Tensor]]=None, also_update: bool=False) -> torch.Tensor:
```
mhmukadam, line 340: So `input_data` is of the form `Dict[str, torch.Tensor]]` when we have this in `TheseusLayer.forward` but here it might be useful for us to add the option to pass a combination of above and a list of `Variables` for convenience.

For instance, evaluating obstacle cost on a given SE2 robot pose. In the current form, you would need to separate the name and data out of the `Variable` (SE2) object first.

theseus/core/objective.py
```diff
@@ -337,7 +337,14 @@ def size_variables(self) -> int:
...
+        old_data = {}
+        if input_data is not None:
+            if also_update == False:
```
mhmukadam, line 343: -> `if not also_update`

theseus/core/tests/test_objective.py
```diff
@@ -280,6 +280,26 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+
+        _check_error_for_data(v1_data, v2_data, error, error_norm_2)
+
+        v1_data = torch.ones(batch_size, dof) * f1 * 0.4
+        v2_data = torch.ones(batch_size, dof) * f2 * 0.4
+
+        input_data = {"v1": v1_data, "v2": v2_data}
+        error = objective.error(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
```
mhmukadam, line 299: We also need a test which covers the case where some input_data is passed and also_update is False.

mhmukadam (commented):
Looking good, left some comments. Thanks for the contributions and welcome to the community!

Also @luisenp might be looking into getting the CI up for PRs coming from forks.

theseus/core/objective.py
jeffin07: @mhmukadam  Is this looks okay ?

`def error(self,input_data: Optional[Dict[str, torch.Tensor]] = None or Optional[List[th.Variable]], also_update: bool = False) > torch.Tensor`

I have a doubt regarding this. If this is the case then we need to check for the type of  `input_data` because we need to copy the older values right ? So how can we copy old values if it's a list ?

luisenp: For this, I think we shouldn't use a list but a dictionary, as in `Optional[Union[Dict[str, torch.Tensor], Dict[str, torch.Variable]]]. Then we do similar logic, except when the type is a variable we use/copy-from `input_data[key].data` instead of `input_data[key]`.

luisenp: Actually, thinking a bit more about this, I don't completely understand the use case you have in mind @mhmukadam. To pass a list of variables, you would already have gone through the effort of creating the necessary variables, giving them the same name as your previous variables, and populating them with the data that you want to evaluate your error on. This seems like more effort than just passing a map of name to tensors?

theseus/core/tests/test_objective.py
luisenp: Agreed.

theseus/core/tests/test_objective.py
```diff
@@ -280,6 +280,26 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+        # For input_data and also_update
+
+        v1_data = torch.ones(batch_size, dof) * f1 * 0.1
+        v2_data = torch.ones(batch_size, dof) * f2 * 0.1
+
+        input_data = {"v1": v1_data, "v2": v2_data}
+        error = objective.error(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+
+        _check_error_for_data(v1_data, v2_data, error, error_norm_2)
```
luisenp, line 292: Besides checking that the error has the correct value, we also need to check that when `also_update=True` the internal variables where actually updated.

theseus/core/objective.py
```diff
@@ -347,10 +354,12 @@ def error(self) -> torch.Tensor:
                 :, pos : pos + cost_function.dim()
             ] = cost_function.weighted_error()
             pos += cost_function.dim()
+        if old_data:
```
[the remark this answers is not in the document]
luisenp: Agreed.

luisenp (changes requested):
Thanks a lot for the contribution @jeffin07, almost there!

@mhmukadam not sure what the CI issue is, I've noticed tests not firing sometimes even for our own PRs.

theseus/core/objective.py
jeffin07: @luisenp Makes more sense. 
> For this, I think we shouldn't use a list but a dictionary, as in `Optional[Union[Dict[str, torch.Tensor], Dict[str, torch.Variable]]]`. Then we do similar logic, except when the type is a variable we use/copy-from `input_data[key].data` instead of `input_data[key]`.

I'll wait for @mhmukadam before i add a commit

luisenp: > Here's an example:
> 
> ```
> p1 = Point2(..., 'p1')
> collision = Collision2D(p1, ...)
> ...
> p2 = ... // some operations that give p2, p2.name could be None
> err = collision.error(p2, also_update=false)
> ```
> 
> Wondering if this is too tricky to implement vs how much we need this? Handling name mismatch would be too complicated but if we restrict to the same name then passing Variable vs torch.Tensor is not that different.

Oh, recall that the error we are computing in this PR is the global objective error. Which means that any input needs to map data to the full set of variables appropriately.

mhmukadam: That sounds good. We can keep this PR independent of that feature and proceed as is.

theseus/core/tests/test_objective.py
jeffin07: @luisenp Did you mean to add a condition when `also_update=True`  and `also_update=False`?

luisenp: No, here I mean that when we pass `also_update=True` the objective should change the values of variables under the hood, and this unit test should confirm that this in fact happened.

jeffin07: @luisenp can we do something like 
```
assert self.optim_vars["a1"] == input_data["a1"]

```

luisenp: Something like that, but for both "v1" and "v2".

jeffin07:
@luisenp Updated Changes after reivew

theseus/core/tests/test_objective.py
```diff
@@ -286,8 +286,10 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
         v2_data = torch.ones(batch_size, dof) * f2 * 0.1
 
         input_data = {"v1": v1_data, "v2": v2_data}
-        error = objective.error(input_data=input_data, also_update=True)
-        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+        error = objective.error(input_data=input_data, also_update=False)
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=False
+        )
```
mhmukadam, line 292: We might need a check after this that makes sure the older variables are still there and didn't get updated?

theseus/core/tests/test_objective.py
```diff
@@ -296,7 +298,19 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+
+        # To check if variables are updated
+        assert (
+            objective.optim_vars["v1"].data.numpy().all()
+            == input_data["v1"].numpy().all()
+        )
+        assert (
+            objective.optim_vars["v2"].data.numpy().all()
+            == input_data["v2"].numpy().all()
+        )
```
mhmukadam, line 313: Maybe use torch.allclose instead?

theseus/core/tests/test_objective.py
luisenp: Oh yes, good catch, let's add this.

theseus/core/tests/test_objective.py
luisenp: See my comment above.

luisenp (changes requested):
Only a few minor changes to test code.

theseus/core/tests/test_objective.py
luisenp: Oh yes, good catch, let's add this.

theseus/core/tests/test_objective.py
```diff
@@ -296,7 +298,19 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
         input_data = {"v1": v1_data, "v2": v2_data}
         error = objective.error(input_data=input_data, also_update=True)
-        error_norm_2 = objective.error_squared_norm(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=True
+        )
+
+        # To check if variables are updated
+        assert (
+            objective.optim_vars["v1"].data.numpy().all()
```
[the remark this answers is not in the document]
luisenp: In fact, we should check that `objective.optim_vars["v1"] is input_data["v1"]`.

theseus/core/tests/test_objective.py
mhmukadam: Ah yes, for the compute graph!

theseus/core/tests/test_objective.py
```diff
@@ -280,6 +280,38 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+        v2_data = torch.ones(batch_size, dof) * f2 * 0.1
+
+        input_data = {"v1": v1_data, "v2": v2_data}
+        error = objective.error(input_data=input_data, also_update=False)
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=False
+        )
+
+        # To check if variables are not updated
+        assert not torch.equal(objective.optim_vars["v1"].data, input_data["v1"])
```
mhmukadam, line 295: We want to verify if the objects are same or not for the compute graph, right @luisenp?
-> `objective.optim_vars["v1"] is not input_data["v1"]`

theseus/core/tests/test_objective.py
```diff
@@ -280,6 +280,38 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+        v2_data = torch.ones(batch_size, dof) * f2 * 0.4
+
+        input_data = {"v1": v1_data, "v2": v2_data}
+        error = objective.error(input_data=input_data, also_update=True)
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=True
+        )
+
+        # To check if variables are updated
+        assert torch.equal(objective.optim_vars["v1"].data, input_data["v1"])
```
mhmukadam, line 310: -> `objective.optim_vars["v1"] is input_data["v1"]`

theseus/core/tests/test_objective.py
mhmukadam: Also just realized these assert are happening after `objective.error_squared_norm` so technically we are not checking `objective.error` yet.

luisenp: I think the safest is to use new tensor variables in line 285-286, and then check here that `objective.optim_vars["v1"] is v1_data` and that `v1_data is not v1_data_new`.

theseus/core/tests/test_objective.py
```diff
@@ -292,8 +292,8 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
         )
 
         # To check if variables are not updated
-        assert not torch.equal(objective.optim_vars["v1"].data, input_data["v1"])
-        assert not torch.equal(objective.optim_vars["v2"].data, input_data["v2"])
+        assert objective.optim_vars["v1"].data is not input_data["v1"]
```
luisenp, line 295: In addition to this check, I still think it's worth it to also confirm that the previous tensor is still there. It's a bit conservative, but it might turn up to be useful in the future, and it's not too much extra effort.

theseus/core/tests/test_objective.py
mhmukadam: No worries.

mhmukadam:
Tried to convert to draft and back but that didn't trigger CI tests either.

theseus/core/tests/test_objective.py
jeffin07: @luisenp like to store the variables before update like `old_data = objective.optim_vars` and checking after `assert old_data is objective.optim_vars`

luisenp: Mostly why I outlined in [this](https://github.com/facebookresearch/theseus/pull/105#discussion_r823873662) comment.

jeffin07: @luisenp Updated the test

theseus/core/tests/test_objective.py
```diff
@@ -282,10 +282,10 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
 
-        v1_data = torch.ones(batch_size, dof) * f1 * 0.1
-        v2_data = torch.ones(batch_size, dof) * f2 * 0.1
+        v1_data_new = torch.ones(batch_size, dof) * f1 * 0.1
+        v2_data_new = torch.ones(batch_size, dof) * f2 * 0.1
 
-        input_data = {"v1": v1_data, "v2": v2_data}
+        input_data = {"v1": v1_data_new, "v2": v2_data_new}
         error = objective.error(input_data=input_data, also_update=False)
         error_norm_2 = objective.error_squared_norm(
```
mhmukadam, line 290: Only thing left is to do these test after both `error` and `error_norm_2`. Maybe you can use a common function since the code will mostly be the same for both.

jeffin07: @mhmukadam updated the unit test

mhmukadam: The common function is good. We also need something like this:
```
error = objective.error(...)
_check_variables(...)
...
error_norm_2 = objective.error_squared_norm(...)
_check_variables(...)
```
So we check on both `objective.error` and `objective.error_squared_norm`.

luisenp: Yeah, at this point is probably best to refactor the test code a bit so we run the same checks on `error` and `error_squared_norm` separately. You could write a function that does all checks, and which is parameterized by the function to test (i..e, `error` or `error_squared_norm`), and the expected output of the computation to check against, and then loop it for the two functions.

luisenp:
CI is failing due to some `mypy` error that they probably just added in a new version. I added a type: ignore to `main` branch, so hopefully it will be fixed in the next CI run.

jeffin07:
@luisenp Do i need to do anything ?

theseus/core/tests/test_objective.py
jeffin07: @luisenp  so you're saying about two function `_check_variables()` and `_check_error_for_data()` which will be called inside loop for `error` and `error_squared_norm` right ?

mhmukadam: You still need to do the checks after both `error` and `error_squared_norm` each. With another commit for this the CI issue should hopefully also go away.

luisenp: @jeffin07 I think you have the right idea. Basically just take the current overall test logic, and repeat it once for `error` and once for `error_squared_norm`.

theseus/core/tests/test_objective.py
```diff
@@ -275,10 +303,65 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+            v2_data_new,
+            error,
+            "error",
+            objective,
+            input_data,
+            also_update=False,
+        )
+
+        error_norm_2 = objective.error_squared_norm(
+            input_data=input_data, also_update=False
```
luisenp, line 329: Can we use a different value of `v1_data_new` and `v2_data_new` when passing to `error_squared_norm`, so it's not the same as the one that used for `error()`?

theseus/core/tests/test_objective.py
```diff
@@ -248,10 +248,38 @@ def _check_all_vars(v1_lis_, v2_lis_, v3_lis_, cw1_opt_lis_, a1_lis_, a2_lis_):
...
+            assert objective.optim_vars["v1"].data is input_data["v1"]
+            assert objective.optim_vars["v2"].data is input_data["v2"]
+        else:
+            assert objective.optim_vars["v1"].data is not input_data["v1"]
+            assert objective.optim_vars["v2"].data is not input_data["v2"]
+
+            assert objective.optim_vars["v1"].data is v1_data
+            assert objective.optim_vars["v2"].data is v2_data
+
+    def _check_error(
```
luisenp, line 276: Let's rename as `_check_error_and_variables`

theseus/core/tests/test_objective.py
```diff
@@ -275,10 +303,65 @@ def _check_error_for_data(v1_data_, v2_data_, error_, error_norm_2_):
...
+            v1_data_new,
+            v2_data_new,
+            error_norm_2,
+            "error_norm_2",
+            objective,
+            input_data,
+            also_update=False,
+        )
+
+        v1_data = torch.ones(batch_size, dof) * f1 * 0.4
```
luisenp, line 342: Shouldn't these be called `v1_data_new` and `v2_data_new`?

luisenp (changes requested):
This looks almost ready to me! I left a few minor comments that should be easy to fix. Thanks!

jeffin07:
@luisenp Updated the changes

luisenp (approved):
LGTM! If @mhmukadam has no additional comments, we can merge. Thanks so much for your contrubution @jeffin07!

PR #108, opened 2022-03-08
Refactored SE3.log_map_impl() to avoid in place operations

luisenp:
Fixes @exhaustin torch backward errors in [this](https://github.com/facebookresearch/theseus/blob/austinw.forward_kinematics/theseus/embodied/kinematics/tests/test_inverse_kinematics.py) script.

The script still has other errors, where the final system is not positive definite so it cannot be solved with CholeskyDense. But unclear if this is related to Lie groups or something else.

### DIFF
--- a/theseus/geometry/se3.py
+++ b/theseus/geometry/se3.py
@@ -16,6 +16,9 @@ from .so3 import SO3
 
 
 class SE3(LieGroup):
+    NEAR_PI_EPS = 1e-7
+    NEAR_ZERO_EPS = 5e-3
+
     def __init__(
         self,
         x_y_z_quaternion: Optional[torch.Tensor] = None,
@@ -174,7 +177,8 @@ class SE3(LieGroup):
             matrix[:, :3, :3].transpose(1, 2) + matrix[:, :3, :3]
         ).abs().max().item() > theseus.constants.EPS:
             raise ValueError(
-                "The 3x3 top-left corner of hat matrices of SE(3) can only be skew-symmetric."
+                "The 3x3 top-left corner of hat matrices of SE(3) can only "
+                "be skew-symmetric."
             )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


theseus/geometry/se3.py
```diff
@@ -250,67 +254,80 @@ def exp_map(tangent_vector: torch.Tensor) -> "SE3":
...
+        translation = a.view(-1, 1) * self[:, :, 3]
+        translation -= 0.5 * torch.cross(rotation, self[:, :, 3])
+        translation += b.view(-1, 1) * (
+            tangent_vector_ang
+            @ (tangent_vector_ang.transpose(1, 2) @ self[:, :, 3].view(-1, 3, 1))
         ).view(-1, 3)
+        return translation
 
-        return ret
+    def _log_map_impl(self) -> torch.Tensor:
```
mhmukadam, line 316: We should make sure whether we need similar fixes in other functions or classes (SE2/SO2/SO3).

mhmukadam (commented):
Nice work on debugging this!

theseus/geometry/se3.py
```diff
@@ -250,67 +254,80 @@ def exp_map(tangent_vector: torch.Tensor) -> "SE3":
...
-        sine_nz = torch.where(near_zero, non_zero, sine[not_near_pi])
+        # theta is !~ pi
         scale = torch.where(
-            near_zero, 1 + sine[not_near_pi] ** 2 / 6, theta[not_near_pi] / sine_nz
-        )
-        ret[not_near_pi, 3:] *= scale.view(-1, 1)
-        # theta is near pi
-        near_pi = ~not_near_pi
-        ddiag = torch.diagonal(self[near_pi], dim1=1, dim2=2)
+            not_near_pi.logical_and(near_zero), 1 + sine**2 / 6, theta / sine
```
fantaosha, line 270: It should be `theta /sine_nz` to avoid `NaN` errors for backward propagation.

fantaosha (commented):
The current implementation might have `NaN` errors for backward propagation when `zero` is the divisor.

theseus/geometry/se3.py
fantaosha: I checked `SE2` and `SO2` and they should be `OK`.

fantaosha:
I also fixed the bugs of `SE3.log_map` in #109, which should avoid `NaN` errors.

theseus/geometry/se3.py
```diff
@@ -250,67 +254,80 @@ def exp_map(tangent_vector: torch.Tensor) -> "SE3":
...
-        theta2_nz = torch.where(near_zero, non_zero, theta2)
-        a = torch.where(
-            near_zero, 1 - theta2 / 12, -sine_theta / two_cosine_minus_two_nz
-        )
+        a = torch.where(near_zero, 1 - theta2 / 12, -sine_theta / two_cosine_minus_two)
         b = torch.where(
             near_zero,
             1.0 / 12 + theta2 / 720,
-            (sine_theta + two_cosine_minus_two) / (theta2_nz * two_cosine_minus_two_nz),
+            (sine_theta + two_cosine_minus_two) / (theta2 * two_cosine_minus_two),
```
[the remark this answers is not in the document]
luisenp: Below is an example of what I mean. Wouldn't the same thing happen here?

image

Output is

```text
Right side has inf True
Does filtered tensor has inf? False
Grad norm tensor(0.8457)
```

theseus/geometry/se3.py
```diff
@@ -16,6 +16,9 @@
 
 
 class SE3(LieGroup):
+    NEAR_PI_EPS = 1e-7
+    NEAR_ZERO_EPS = 5e-3
```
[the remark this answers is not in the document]
luisenp: I did a quick search before making the change and noticed the same value was used everywhere, so I figured I would keep one value for now. The reason I moved it here is because I split `_log_map_impl()` into two functions that were using the same EPS variable. 

In any case, we can probably defer this until we do more testing to see what the right EPS is for each function, and refactor accordingly.

luisenp:
> @luisenp the divide by zero issues @fantaosha pointed out above should have failed unit tests?
> 
> > I also fixed the bugs of SE3.log_map in #109, which should avoid NaN errors.
> 
> @fantaosha, wait so if the fixes are in #109 then do you propose to merge this PR directly to main as is? Might have been cleaner to push your fixes to this PR.
> 
> We should also add unit tests that will fail with the previous version.

I think the potential `nan` errors @fantaosha pointed out would only happen in backward pass, if I understand him correctly. But I'm not sure this would actually happen, as I mentioned in a comment above. But maybe he saw them come up in practice?

Assuming we go along with the fix, I think it's fine to merge #108 and #109 in order, since we will merge both pretty quickly. But I agree that in the future it will be cleaner to submit fixes directly to the relevant PR.

luisenp:
Oh, after I looked at #109, it seems that it overrides this one, so we would only need to merge #109, is this correct? I also checked and looks like @fantaosha's changes also fix Austin's backward errors in the inverse kinematics script, so looks good to me. I'm OK closing with this PR and merging #109. 

We need to add unit tests for this, but as discussed offline, this can wait until @fantaosha finishes the derivatives for log_map, and exp_map; then it would be much easier to test.

fantaosha:
> @luisenp the divide by zero issues @fantaosha pointed out above should have failed unit tests?
> 
> > I also fixed the bugs of SE3.log_map in #109, which should avoid NaN errors.
> 
> @fantaosha, wait so if the fixes are in #109 then do you propose to merge this PR directly to main as is? Might have been cleaner to push your fixes to this PR.
> 
> We should also add unit tests that will fail with the previous version.

I will merge #109 to #108 first, and then merge to main.

PR #110, opened 2022-03-09
Add analytical jacobians for `LieGroup.exp_map`

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/theseus/geometry/point_types.py
+++ b/theseus/geometry/point_types.py
@@ -94,7 +94,11 @@ class Point2(Vector):
         if jacobians is not None:
             shape = tangent_vector.shape
             Point2._check_jacobians_list(jacobians)
-            jacobians.append(torch.eye(2).repeat(shape[0], 1, 1))
+            jacobians.append(
+                torch.eye(
+                    2, dtype=tangent_vector.dtype, device=tangent_vector.device
+                ).repeat(shape[0], 1, 1)
+            )
 
         return Point2(data=tangent_vector.clone())
 
@@ -174,7 +178,11 @@ class Point3(Vector):
         if jacobians is not None:
             shape = tangent_vector.shape
             Point2._check_jacobians_list(jacobians)
-            jacobians.append(torch.eye(3).repeat(shape[0], 1, 1))
+            jacobians.append(
+                torch.eye(
+                    3, dtype=tangent_vector.dtype, device=tangent_vector.device
+                ).repeat(shape[0], 1, 1)
+            )
 
         return Point3(data=tangent_vector.clone())
 
--- a/theseus/geometry/point_types.py
+++ b/theseus/geometry/point_types.py
@@ -3,7 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Optional, Tuple, cast
+from typing import List, Optional, Tuple, cast
 
 import torch
 
@@ -85,7 +85,17 @@ class Point2(Vector):
         return self[:, 1]
 
     @staticmethod
-    def exp_map(tangent_vector: torch.Tensor) -> "Point2":
+    def exp_map(
+        tangent_vector: torch.Tensor, jacobians: Optional[List[torch.Tensor]] = None
+    ) -> "Point2":
+        if tangent_vector.ndim != 2 or tangent_vector.shape[1] != 2:
+            raise ValueError("Tangent vectors of Point2 should be 2-D vectors.")
+
+        if jacobians is not None:
+            shape = tangent_vector.shape
+            Point2._check_jacobians_list(jacobians)
+            jacobians.append(torch.eye(2).repeat(shape[0], 1, 1))
+
         return Point2(data=tangent_vector.clone())
 
     # added to avoid casting downstream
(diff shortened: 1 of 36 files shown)


theseus/geometry/point_types.py
```diff
@@ -155,7 +169,21 @@ def z(self) -> torch.Tensor:
...
-    def exp_map(tangent_vector: torch.Tensor) -> "Point3":
+    def exp_map(
+        tangent_vector: torch.Tensor, jacobians: Optional[List[torch.Tensor]] = None
+    ) -> "Point3":
+        if tangent_vector.ndim != 2 or tangent_vector.shape[1] != 3:
+            raise ValueError("Tangent vectors of Point3 should be 3-D vectors.")
+
+        if jacobians is not None:
+            shape = tangent_vector.shape
+            Point2._check_jacobians_list(jacobians)
```
luisenp, line 180: Should it be `Point3` here? BTW, better to add a common method here for `Point2` and `Point3`, parameterized by the class/dof, since the logic is pretty much the same.

theseus/geometry/se3.py
```diff
@@ -232,35 +234,103 @@ def exp_map(tangent_vector: torch.Tensor) -> "SE3":
...
     def _log_map_impl(self) -> torch.Tensor:
         NEAR_PI_EPS = 1e-7
         NEAR_ZERO_EPS = 5e-3
 
-        ret = torch.zeros(self.shape[0], 6, dtype=self.dtype, device=self.device)
-
-        ret[:, 3] = 0.5 * (self[:, 2, 1] - self[:, 1, 2])
-        ret[:, 4] = 0.5 * (self[:, 0, 2] - self[:, 2, 0])
-        ret[:, 5] = 0.5 * (self[:, 1, 0] - self[:, 0, 1])
+        sine_axis = torch.zeros(self.shape[0], 3, dtype=self.dtype, device=self.device)
```
luisenp, line 328: This part was already merged, correct? Is this GitHub not syncing correctly?

theseus/geometry/vector.py
```diff
@@ -204,7 +204,21 @@ def _project_impl(
...
 
     @staticmethod
-    def exp_map(tangent_vector: torch.Tensor) -> "Vector":
+    def exp_map(
+        tangent_vector: torch.Tensor, jacobians: Optional[List[torch.Tensor]] = None
+    ) -> "Vector":
+        if tangent_vector.ndim != 2:
+            raise ValueError("The dimension of tangent vectors should be 2.")
+
+        if jacobians is not None:
```
luisenp, line 213: In fact, maybe we can just have `Point2` and `Point3` reuse this method, and maybe just pass a custom error message?

luisenp (changes requested):
LGTM! I left some smallish comments.

luisenp:
BTW, do you mind sneaking in a fix for this [typo](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/point_types.py#L24) here? (Provied --> Provided).

theseus/geometry/se3.py
fantaosha: Yes.

theseus/geometry/point_types.py
mhmukadam: Any way to just call the Vector.exp_map with the dof instead of a common Point2/Point3 function?

theseus/geometry/vector.py
```diff
@@ -204,7 +204,21 @@ def _project_impl(
         return euclidean_grad.clone()
 
     @staticmethod
-    def exp_map(tangent_vector: torch.Tensor) -> "Vector":
+    def exp_map(
```
mhmukadam, line 207: We should update tests for Vector and Point2/3 as well.

theseus/geometry/point_types.py
fantaosha: The reason why I use `Point3` is to pass the `precommit` checks. I forgot the exact issues if using `Vector`.

theseus/geometry/vector.py
fantaosha: Yes, I agree.

theseus/geometry/tests/common.py
```diff
@@ -219,3 +219,27 @@ def func(g):
...
+    batch_size = tangent_vector.shape[0]
+    dof = tangent_vector.shape[1]
+    aux_id = torch.arange(batch_size)
+
+    def exp_func(xi):
+        return Group.exp_map(xi).to_matrix()
+
+    actual = []
+    group = Group.exp_map(tangent_vector, jacobians=actual).to_matrix()
+    jac_raw = torch.autograd.functional.jacobian(exp_func, (tangent_vector.data))[
```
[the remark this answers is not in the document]
fantaosha: There should be no errors of `exp_map` as long as `torch.autograd.functional.jacobian` works. `torch.autograd.funcational.jacobian` just implements backward pass to calculate the jacobians.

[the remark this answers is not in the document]
luisenp: I know there should be no errors. I want to see if this test would have caught the **old** errors.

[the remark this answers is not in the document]
fantaosha: OK, I will check this.

theseus/geometry/point_types.py
fantaosha: Ah, I see. I will fix this. But there should be no differences since both of them call `LieGroup._check_jacobians_list`.

theseus/geometry/vector.py
mhmukadam: Just for sanity check and completeness :) and to make sure the compute graph didn't break anywhere which the torch.autograd would then reveal.

fantaosha: Tested.

mhmukadam (approved):
Looks good!

PR #112, opened 2022-03-09
First version of the Bundle Adjustment script

luisenp:
This is adapted from Maurizio's original notebook. For this PR, I'll mostly do clean up and add a few bug fixes I found. Note that this is not yet a full Bundle Adjustment implementation. After this PR is merged, a few TODOs are (not necessarily complete list):

- Add full bundle adjustment implementation (i.e., lots of cameras, lots of points, also optimize points locations).
- Replace `SO3` + `Point3` by `SE3` objects. 
- Add data loading. 
- Consider moving the `ReprojectionError` factor somewhere and adding unit tests.

### DIFF
--- a/examples/bundle_adjustment.py
+++ b/examples/bundle_adjustment.py
@@ -1,5 +1,4 @@
 import math
-from typing import cast
 
 import torch
 import theseus as th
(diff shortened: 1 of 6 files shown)


theseus/utils/examples/__init__.py
```diff
@@ -3,6 +3,11 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+from .bundle_adjustment import (
+    LocalizationDataset,
+    LocalizationSample,
+    ReprojectionError,
```
mhmukadam, line 9: `ReprojectionError` (-> `Reprojection`) could go to `embodied/camera` as this cost function would be more widely useful.

examples/bundle_adjustment.py
```diff
@@ -0,0 +1,92 @@
+import torch
```
mhmukadam, line 1: This and many other new files in this PR need the license header.

theseus/utils/examples/bundle_adjustment/util.py
```diff
@@ -0,0 +1,57 @@
...
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    ratio = (x + radius) / radius
+    sq = torch.sqrt(ratio)
+    val = (sq - 1) * radius
+    der = 0.5 / sq
+    return val, der
+
+
+# returns a uniformly random point of the 2-sphere
+def random_s2(dtype: torch.dtype = torch.float64) -> torch.Tensor:
```
mhmukadam, line 27: Do we need these or can we replace with Taosha's inside the class version?

theseus/utils/examples/bundle_adjustment/util.py
```diff
@@ -0,0 +1,57 @@
...
+def soft_loss_cauchy(
+    x: torch.Tensor, radius: torch.Tensor
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    ratio = (x + radius) / radius
+    val = torch.log(ratio) * radius
+    der = 1.0 / ratio
+    return val, der
+
+
+def soft_loss_huber_like(
```
mhmukadam, line 16: I suspect many of these will be useful for @ddetone in his homography example too so we could move them up to main utils?

theseus/utils/examples/bundle_adjustment/util.py
```diff
@@ -0,0 +1,57 @@
...
+        [
+            torch.sqrt(1 - u) * torch.sin(math.tau * v),
+            torch.sqrt(1 - u) * torch.cos(math.tau * v),
+            torch.sqrt(u) * torch.sin(math.tau * w),
+            torch.sqrt(u) * torch.cos(math.tau * w),
+        ]
+    ).to(dtype=dtype)
+
+
+def random_small_quaternion(
```
mhmukadam, line 47: This could go to the quaternion class we will add in #60.

theseus/utils/examples/bundle_adjustment/reprojection_error.py
```diff
@@ -0,0 +1,118 @@
...
+        self.world_point = world_point
+        self.image_feature_point = image_feature_point
+
+        self.register_optim_vars(["camera_rotation", "camera_translation"])
+        self.register_aux_vars(
+            ["loss_radius", "focal_length", "image_feature_point", "world_point"]
+        )
+
+    def error(self) -> torch.Tensor:
+        point__cam = (
```
mhmukadam, line 43: Is the double underscore intentional? (I am not familiar with this convention)

luisenp: I took this from Paloma's tactile pushing [code](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/motionmodel/quasi_static_pushing_planar.py#L62), where it is used to indicate the frame of reference for a variable. I don't know if it's standard or something she came up with, but I like it, seems concise and easy to parse.

theseus/utils/examples/bundle_adjustment/util.py
luisenp: @maurimo, can you check [this](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/so3.py#L46) and see if the sampling functions you added are complementary to those. If not, then it's better to replace with `th.rand_so3` or `th.randn_so3`.

PR #114, opened 2022-03-12
Add analytical derivatives for `LieGroup.log_map()`

fantaosha:
## Types of changes

## Checklist


fantaosha:
Now the computation of `SO3.log_map()` and `SE3.log_map()` has been improved around `pi` and all the analytical derivatives tests are passed.

theseus/geometry/point_types.py
```diff
@@ -91,7 +91,7 @@ def exp_map(
         if tangent_vector.ndim != 2 or tangent_vector.shape[1] != 2:
             raise ValueError("Tangent vectors of Point2 should be 2-D vectors.")
 
-        Vector._exp_map_jacobian(tangent_vector, jacobians)
+        Vector._exp_map_jacobian_impl(tangent_vector, jacobians)
```
mhmukadam, line 94: Should there be functions in Point2/3 for logmap calling `Vector._log_map_jacobian_impl`?

mhmukadam (approved):

theseus/geometry/point_types.py
fantaosha: Not necessary since `Point2.log_map()` and `Point3.log_map()` calls `Vector.log_map()`

mhmukadam: Ah right. For exp map you needed it explicitly here for the tangent vector size check.

fantaosha: Yes, we checked the sizes [here](https://github.com/facebookresearch/theseus/blob/e487c03917ad19747c2336bdd39f8dbe1d8b70d5/theseus/geometry/point_types.py#L91).

theseus/geometry/tests/common.py
```diff
@@ -259,3 +275,19 @@ def exp_func(xi):
     ]
 
     assert torch.allclose(actual[0], expected, atol=atol)
+
+
+def check_projection_for_vector_log_map(tangent_vector, Group, atol=1e-8):
```
luisenp, line 280: We could reuse `check_projection_for_log_map()` but pass an extra kwarg that specifies if we should call `group.project()` before or not. I think that's the only difference, right?

luisenp (approved):
LGMT! Left only a very minor comment.

theseus/geometry/tests/common.py
fantaosha: The `check_projection_for_map` is reused.

PR #117, opened 2022-03-14
A complete version of Bundle Adjusment 

luisenp:
## Motivation and Context
This merges #112 into @maurimo's follow up [branch](https://github.com/facebookresearch/theseus/tree/mau.bundle_adjustment). I discarded the `camera_localization.py` script because it seems to have errors (using an old `ReprojectionError` version).

### DIFF
--- a/theseus/utils/examples/bundle_adjustment/data.py
+++ b/theseus/utils/examples/bundle_adjustment/data.py
@@ -159,7 +159,7 @@ class BundleAdjustmentDataset:
         self.gt_points = gt_points
 
     @staticmethod
-    def load_bal_dataset(path):
+    def load_bal_dataset(path: str):
         observations = []
         cameras = []
         points = []
@@ -202,7 +202,12 @@ class BundleAdjustmentDataset:
         return cameras, points, observations
 
     @staticmethod
-    def save_bal_dataset(path, cameras, points, observations):
+    def save_bal_dataset(
+        path: str,
+        cameras: List[Camera],
+        points: List[th.Point3],
+        observations: List[Observation],
+    ):
         with open(path, "wt") as out:
             print(f"{len(cameras)} {len(points)} {len(observations)}", file=out)
             for obs in observations:
@@ -324,3 +329,12 @@ class BundleAdjustmentDataset:
         return BundleAdjustmentDataset(
             cameras, points, observations, gt_cameras, gt_points
         )
+
+
+def ba_histogram(
+    cameras: List[Camera],
+    points: List[th.Point3],
+    observations: List[Observation],
+):
+    ba = BundleAdjustmentDataset(cameras, points, observations)
+    ba.histogram()
--- a/theseus/utils/examples/bundle_adjustment/data.py
+++ b/theseus/utils/examples/bundle_adjustment/data.py
@@ -17,9 +17,9 @@ def add_noise_and_outliers(
 
     if linear_noise:
         feat_image_points = proj_points + noise_size * (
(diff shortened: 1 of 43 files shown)


theseus/utils/examples/bundle_adjustment/data.py
```diff
@@ -0,0 +1,340 @@
...
+
+import numpy as np
+import torch
+
+import theseus as th
+
+from .util import random_small_quaternion
+
+
+class Camera:
```
mhmukadam, line 11: Move Camera and ReprojectionError to embodied?

Maybe not urgent as we prioritize experiments but also not sure how much we should punt as 'cleanup' for later.

luisenp: My hesitation is that I would consider the standard for moving to `embodied` higher than keeping it here, and this would mean spending extra time writing a full test suite, possibly refactor the code, then iterating during code review, etc. I would prefer to defer until we have more progress on the next milestone.

theseus/utils/examples/bundle_adjustment/reprojection_error.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class ReprojectionError(th.CostFunction):
+    def __init__(
+        self,
+        camera_pose: th.SE3,
+        log_loss_radius: th.Vector,
+        world_point: th.Point3,
+        image_feature_point: th.Point2,
+        focal_length: th.Vector,
```
mhmukadam, line 17: nit: for ease in reading maybe order them based on hoe they are registered, optim first then aux?

theseus/utils/examples/bundle_adjustment/util.py
```diff
@@ -0,0 +1,63 @@
...
+    val = (sq - 1) * radius
+    der = 0.5 / sq
+    return val, der
+
+
+# ------------------------------------------------------------ #
+# ----------------------------- RNG -------------------------- #
+# ------------------------------------------------------------ #
+# returns a uniformly random point of the 2-sphere
+def random_s2(dtype: torch.dtype = torch.float64) -> torch.Tensor:
```
mhmukadam, line 33: Can we replace with the class version?

mhmukadam (commented):
Looks good. Are any other unit tests needed beyond testing reprojection error? Also probably incorporate relevant comments from #112 if not above already.

theseus/utils/examples/bundle_adjustment/util.py
luisenp: cc: @maurimo

mhmukadam (approved):
Looks good! Feel free to merge after resolving the two open comments.

theseus/utils/examples/bundle_adjustment/reprojection_error.py
luisenp: Done!

theseus/utils/examples/bundle_adjustment/util.py
maurimo: certainly - this can be cleaned altogether, it's dangling code and not even used in the bundle adjustment IIRC!

maurimo (approved):
Looks good to me, thanks for improving the code quality @luisenp !

theseus/utils/examples/bundle_adjustment/util.py
luisenp: It looks like these are still used. I don't quite understand the difference between this sampling and the one @fantaosha  implemented in `.rand()`, so in the interest of time I propose to leave this as is for the time being.

PR #121, opened 2022-03-16
Add ManifoldGaussian class for messages in belief propagation

joeaortiz:
## Motivation and Context

It would be useful to have an optional covariance / precision matrix as part of the Manifold class as Gaussian Belief Propagation involves sending Gaussian distributions over the Manifold variables. Currently I'm using a wrapper Gaussian class but it could be more widely useful to have covariance / precision matrix as an attribute of the Manifold class?

## Types of changes

## Checklist


theseus/geometry/manifold.py
```diff
@@ -130,3 +130,10 @@ def local(
 # Alias for Manifold.retract()
 def retract(variable: Manifold, delta: torch.Tensor) -> Manifold:
     return variable.retract(delta)
+
+
+class Gaussian:
```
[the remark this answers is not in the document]
joeaortiz: Yes I think that would be a good way to add this feature. 

Precision is preferable over covariance because in GBP the messages are Gaussians over the variables and so are also represented with the Manifold class. The messages can represent relative constraints (without a global anchor) and so can have infinite uncertainty along a direction, i.e. they have rank deficient covariance matrices so we need to use the precision matrix to represent the message. Adding a positive definite check could still make sense for representing the beliefs, but for the messages I would need go disable this check.

joeaortiz:
I assume that all Manifold instances passed in the mean Sequence have the same batch size, dtype and device.
I haven't included the get and set item functions as the class stores two internal properties: mean and precision.

theseus/optimizer/marginals.py
```diff
@@ -3,23 +3,98 @@
...
+        dof = 0
+        for v in mean:
+            dof += v.dof()
+        self._dof = dof
+
+        self.mean = mean
+        self.precision = (
+            torch.zeros(mean[0].shape[0], self.dof, self.dof)
+            .to(mean[0].dtype)
+            .to(mean[0].device)
```
mhmukadam, line 38: You can pass these in the same `to()` call.

theseus/optimizer/marginals.py
```diff
@@ -3,23 +3,98 @@
...
         )
 
+    @property
     def dof(self) -> int:
-        return self.tot_dof
+        return self._dof
+
+    @property
+    def device(self) -> torch.device:
+        return self.precision[0].device
```
mhmukadam: Should device, dtype, and to also operate over `mean` or just precision? cc @luisenp

luisenp: It seems that `to` operates over mean and precision, which seems OK to me. `device` and `dtype` returning precision's properties seems fine, but maybe we can add a consistency check between `mean` and `precision` in the constructor.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,98 @@
...
+        else:
+            self.name = f"{self.__class__.__name__}__{self._id}"
+
+        dof = 0
+        for v in mean:
+            dof += v.dof()
+        self._dof = dof
+
+        self.mean = mean
+        self.precision = torch.zeros(mean[0].shape[0], self.dof, self.dof).to(
```
mhmukadam, line 35: Since precision is an optional argument to init you want a 'if none' check here before assigning zeros, and if something is passed then you can add the check for dtype/device Luis suggested as well as check for the shape to be correct (and if not then raise value error with a message).

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,107 @@
...
+        if id(self) in memo:
+            return memo[id(self)]
+        the_copy = self.copy()
+        memo[id(self)] = the_copy
+        return the_copy
+
+    def update(
+        self,
+        mean: Optional[Sequence[Manifold]] = None,
+        precision: Optional[torch.Tensor] = None,
```
mhmukadam, line 77: Why is precision and mean are both optional. Would there be cases where you just update mean but not precision or vice-versa?

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,107 @@
...
+    ):
+        if mean is not None:
+            if len(mean) != len(self.mean):
+                raise ValueError(
+                    f"Tried to update mean with sequence of different"
+                    f"lenght to original mean sequence. Given {len(mean)}. "
```
mhmukadam, line 83: lenght -> length

mhmukadam (commented):
Looks good so far! On to unit tests. Any other features you are considering adding still?

theseus/optimizer/manifold_gaussian.py
joeaortiz: Probably there aren't any cases when you'll want to update one and not the other. I'll make both required arguments.

joeaortiz:
I don't think so. I was thinking of adding the local and retract functions but generalised to gaussians on manifold variables, but they would be better placed in the Manifold class.

joeaortiz:
> Looks good so far! On to unit tests. Any other features you are considering adding still?

I don't think so. I was thinking of adding the local and retract functions but generalised to gaussians on manifold variables, but they would be better placed in the Manifold class.

mhmukadam:
> local and retract functions but generalised to gaussians on manifold variables
Okay probably add it in this PR. Manifold already has a local and retract, how will this Gaussian version get incorporated?

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,103 @@
...
+        self,
+        mean: Sequence[Manifold],
+        precision: Optional[torch.Tensor] = None,
+        name: Optional[str] = None,
+    ):
+        self._id = next(ManifoldGaussian._ids)
+        if name:
+            self.name = name
+        else:
+            self.name = f"{self.__class__.__name__}__{self._id}"
```
mhmukadam, line 27: Nit: you can do the same thing here like precision below.
```
if name is None:
    name = autogenerated
self.name = name
```

theseus/optimizer/manifold_gaussian.py
```diff
@@ -21,10 +21,9 @@ def __init__(
         name: Optional[str] = None,
     ):
         self._id = next(ManifoldGaussian._ids)
-        if name:
-            self.name = name
-        else:
+        if name is None:
             self.name = f"{self.__class__.__name__}__{self._id}"
+        self.name = name
```
mhmukadam, line 26: This line will override with `None`. You need the line above to be `name = ` instead of `self.name = `.

joeaortiz:
Moved `local_gaussian` and `retract_gaussian` fns to `manifold_gaussian.py` for now to avoid circular imports. 

Still to do unit tests for `local_gaussian` and `retract_gaussian`.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -98,5 +98,42 @@ def update(
                 f"Tried to update using tensor on device: {precision.dtype} but precision "
                 f"is on device: {self.device}."
             )
+        if not torch.allclose(precision, precision.transpose(1, 2)):
```
mhmukadam, line 101: Good catch on adding this!

theseus/optimizer/tests/test_manifold_gaussian.py
```diff
@@ -0,0 +1,146 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import copy
+
+import numpy as np
+import pytest  # noqa: F401
+import torch
+
+import theseus as th
+from theseus.optimizer.manifold_gaussian import ManifoldGaussian
```
mhmukadam, line 13: You should add `ManifoldGaussian` to the appropriate inits to access from `th.` directly.

theseus/optimizer/tests/test_manifold_gaussian.py
```diff
@@ -0,0 +1,146 @@
...
+
+        t = ManifoldGaussian(mean, precision=precision, name=name)
+        all_ids.append(t._id)
+        if name is not None:
+            assert name == t.name
+        else:
+            assert t.name == f"ManifoldGaussian__{t._id}"
+        assert t.dof == dof
+        for i in range(n_vars):
+            assert t.mean[i] == mean[i]
```
mhmukadam, line 56: Any reason mean are being save with [i]?

theseus/optimizer/tests/test_manifold_gaussian.py
```diff
@@ -0,0 +1,146 @@
...
+        precision = torch.zeros(mean[0].shape[0], dof, dof)
+    else:
+        precision = None
+
+    return mean, precision
+
+
+def test_init():
+    all_ids = []
+    for i in range(100):
```
mhmukadam, line 39: Is probably sufficient for all of these loop to be range(10) so CI isn't slowed down.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -98,5 +98,42 @@ def update(
...
+    gaussian: ManifoldGaussian,
+    return_mean: bool = True,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    # mean vector in the tangent space at variable
+    mean_tp = variable.local(gaussian.mean[0])
+
+    jac: List[torch.Tensor] = []
+    variable.exp_map(mean_tp, jacobians=jac)
+    # precision matrix in the tangent space at variable
+    lam_tp = torch.bmm(torch.bmm(jac[0].transpose(-1, -2), gaussian.precision), jac[0])
```
mhmukadam, line 118: Would be good to cite something here for the math.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -98,5 +98,42 @@ def update(
...
         self.precision = precision
+
+
+def local_gaussian(
+    variable: LieGroup,
+    gaussian: ManifoldGaussian,
+    return_mean: bool = True,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    # mean vector in the tangent space at variable
+    mean_tp = variable.local(gaussian.mean[0])
```
mhmukadam, line 113: Does this assume the passed ManifoldGaussian is over just one Manifold object?

theseus/optimizer/manifold_gaussian.py
```diff
@@ -98,5 +98,42 @@ def update(
                 f"Tried to update using tensor on device: {precision.dtype} but precision "
                 f"is on device: {self.device}."
             )
+        if not torch.allclose(precision, precision.transpose(1, 2)):
+            raise ValueError("Tried to update precision with non-symmetric matrix.")
 
         self.precision = precision
+
+
+def local_gaussian(
```
mhmukadam, line 107: Should these be class methods as well? I am wondering if calling this just `local` is okay then?

mhmukadam (commented):
Looks good. Left some comments.

theseus/optimizer/manifold_gaussian.py
joeaortiz: Yes, it assumes the ManifoldGaussian is over just one Manifold object which must be the same class as the variable LieGroup object.

theseus/optimizer/manifold_gaussian.py
joeaortiz: The should really be class methods of the `LieGroup` / `Manifold` class as then they would follow the same format as the existing `local` and `retract` functions. I think it wouldn't make sense to make this a class method of `ManifoldGaussian`. 

I initially added `local_gaussian` and `retract_gaussian` as class methods of `Manifold` and `LieGroup` but removed them as it caused a circular import. Maybe it is fine to just leave them here for now?

theseus/optimizer/manifold_gaussian.py
joeaortiz: I've now added a check for this.

theseus/geometry/se3.py
```diff
@@ -285,14 +285,26 @@ def exp_map(
...
+            minus_one_by_twelve = torch.full(
+                near_zero.shape,
+                -1 / 12.0,
+                dtype=sine_by_theta.dtype,
+                device=sine_by_theta.device,
+            )
             d_one_minus_cosine_by_theta2 = torch.where(
                 near_zero,
-                -1 / 12.0,
+                minus_one_by_twelve,
```
mhmukadam, line 296: Can you add this commit as a separate PR to main so the fix is disentangled from this PR and other PRs can immediately rebase on (once merged to main you can rebase here as well).

theseus/optimizer/manifold_gaussian.py
```diff
@@ -109,6 +110,21 @@ def local_gaussian(
     gaussian: ManifoldGaussian,
     return_mean: bool = True,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
+    # assumes gaussian is over just one Manifold object
+    if len(gaussian.mean) != 1:
+        raise ValueError(
+            "ManifoldGaussian should be over just one Manifold object. "
```
mhmukadam, line 116: local on ManifoldGaussian should be over ...

theseus/optimizer/tests/test_manifold_gaussian.py
```diff
@@ -0,0 +1,196 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import copy
+
+import numpy as np
+import pytest  # noqa: F401
+import torch
+
+import theseus as th
+from theseus.optimizer.manifold_gaussian import local_gaussian, retract_gaussian
```
mhmukadam, line 13: We should also add these to the appropriate init files if we are not moving them into the ManifoldGaussian class. If we are then it would makes sense to just call then local and retract? Unfortunately we can't rename the external ones since the Manifold class already has a local/retract in the `th.` namespace.

theseus/optimizer/manifold_gaussian.py
mhmukadam: Makes sense. 
How about we move them to be class functions with names local and retract. If you then continue using the class functions it will be easier to figure out later if there is a way to combine with the Manifold local/retract without needing changes everywhere.

mhmukadam (commented):
Looks ready once we decide whether to move the local/retract inside the class.

theseus/optimizer/tests/test_manifold_gaussian.py
luisenp: Maybe in a later PR we could rename the old `local` and `retract` to something like `_local_lie_group`, and these ones to something like `_local_gaussian`. Then in Theseus' `__init__.py` we define `local` to route to the appropriate one based on the type.

theseus/optimizer/manifold_gaussian.py
luisenp: Oh, sorry, in my previous comment/question I thought that you had added `local_gaussian` as a class method of `ManifoldGaussian` (instead of `Manifold`).

theseus/optimizer/tests/test_manifold_gaussian.py
mhmukadam: Nice suggestion! Added issue #159.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -113,7 +113,7 @@ def local_gaussian(
     # assumes gaussian is over just one Manifold object
     if len(gaussian.mean) != 1:
         raise ValueError(
-            "ManifoldGaussian should be over just one Manifold object. "
+            "local on manifold should be over just one Manifold object. "
```
mhmukadam, line 116: local on ManifoldGaussian?

mhmukadam (approved):
LGTM! Is the plan to keep local/retract outside of the class?

theseus/optimizer/manifold_gaussian.py
joeaortiz: I don't think it makes sense to add these as class functions to the `ManifoldGaussian` class. It would be confusing as they would not be consistent with the existing local and retract functions in the `Manifold` class. 

e.g. `var1.local(var2)` gives the value of `var2` in the tangent plane at `var1` where both are `Manifold` instances.
`local_gaussian(var1, gauss1)` gives the mean and precision of `gauss1` in the tangent plane at `var1`, so it wouldn't make sense to make it a class function of `ManifoldGaussian`.

joeaortiz:
Should be ready to merge now.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,156 @@
...
+                f"Tried to update using tensor on device: {precision.dtype} but precision "
+                f"is on device: {self.device}."
+            )
+        if not torch.allclose(precision, precision.transpose(1, 2)):
+            raise ValueError("Tried to update precision with non-symmetric matrix.")
+
+        self.precision = precision
+
+
+def local_gaussian(
```
luisenp, line 108: Can you add a bit more detailed comment as to what operation this performs in the reference below? (e.g., point to Eq. number, page, or subsection). And also what the two return values represent.

luisenp (changes requested):
Looks good! Left some comments/questions. I didn't check the math closely.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,156 @@
...
+        self.name = name
+
+        dof = 0
+        for v in mean:
+            dof += v.dof()
+        self._dof = dof
+
+        self.mean = mean
+        if precision is None:
+            precision = torch.zeros(mean[0].shape[0], self.dof, self.dof).to(
```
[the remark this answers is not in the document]
mhmukadam: Good catch! Zero default would't be invertible and could cause problems.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,156 @@
...
+    def to(self, *args, **kwargs):
+        for var in self.mean:
+            var = var.to(*args, **kwargs)
+        self.precision = self.precision.to(*args, **kwargs)
+
+    def copy(self, new_name: Optional[str] = None) -> "ManifoldGaussian":
+        if not new_name:
+            new_name = f"{self.name}_copy"
+        mean_copy = [var.copy() for var in self.mean]
+        return ManifoldGaussian(mean_copy, name=new_name)
```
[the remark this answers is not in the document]
joeaortiz: Thanks for spotting this, now fixed.

theseus/optimizer/tests/test_manifold_gaussian.py
```diff
@@ -0,0 +1,195 @@
...
+    mean = []
+    dof = 0
+    for i in range(n_vars):
+        ix = np.random.randint(len(manif_types))
+        var = manif_types[ix].rand(batch_size)
+        mean.append(var)
+        dof += var.dof()
+
+    if np.random.random() < 0.5:
+        precision = torch.zeros(mean[0].shape[0], dof, dof)
```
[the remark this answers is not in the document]
joeaortiz: Thanks, now fixed this!

theseus/optimizer/manifold_gaussian.py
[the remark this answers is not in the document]
joeaortiz: I had initialised the precision with zeros because then I can directly use it for messages which are initialised with 0 precision (i.e. zero information). But I agree that for wider use it makes more sense to initialise with the identity.

theseus/optimizer/manifold_gaussian.py
```diff
@@ -141,6 +148,12 @@ def local_gaussian(
         return eta_tp, lam_tp
 
 
+# Computes the ManifoldGaussian that corresponds to the gaussian in the tangent plane
+# at variable, parameterised by hte mean (mean_tp) and precision (precision_tp).
```
mhmukadam, line 153: typo: hte -> the

theseus/optimizer/manifold_gaussian.py
```diff
@@ -0,0 +1,170 @@
...
+            dof += v.dof()
+        self._dof = dof
+
+        self.mean = mean
+        if precision is None:
+            precision = torch.eye(self.dof).to(
+                dtype=mean[0].dtype, device=mean[0].device
+            )
+            precision = precision[None, ...].repeat(mean[0].shape[0], 1, 1)
+        self.update(mean, precision)
```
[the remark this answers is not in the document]
joeaortiz: Within the `update` function there is a check that the matrix is symmetric. 
On positive definiteness, I need to allow non positive definite precision matrices as GBP messages can have rank deficient precisions. e.g. in bundle adjustment the reprojection factors send rank deficient messages.

luisenp (approved):
LGTM!

Issue #113, opened 2022-03-10
Override `Vector` operators in `Point2` and `Point3` with the correct return type


luisenp:
## 🚀 Feature

Something like 

```python
class Point2(Vector):
    ...
    
    def __add__(self, other: Vector) -> "Point2":
        return cast(Point2, super().__add__(other))
```

## Motivation

Eliminates unnecessary casting when using typing. 

## Alternatives

There might be some way for `mypy` to do the correct thing that wouldn't require overriding these methods.

jeffin07:
@luisenp I am intrested in this Can you please guide me ?

jeffin07:
@luisenp Is this issue assinged to anyone ? Or do you want me to pick another one ?

luisenp:
Hi @jeffin07, no, in fact it'd be great if you can take a look at this one! Sorry for the delay, I've been busy this week, so I forgot about this issue. 

We have a `Vector` class for our optimization variables, which in turn has two subclasses,`Point2` and `Point3`. `Vector` [overrides](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/vector.py#L94) a lot of the Python operators so that we can do things like `x + y`, `x * y`, and so on. The point subclasses inherit these operators, but when using `mypy` for typing, doing something like the code below will throw a typing error. 

```python
x = th.Point2(...)
y = th.Point2(...)
z = x + y
out = function_expects_point2(z)  # mypy errors here because z is th.Vector, not th.Point2
```

So, the goal of this issue is to override the operators in the subclasses, to avoid this typing issue. The obvious way would be to do something like what I described in the original comment above, inside classes [Point2](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/point_types.py#L29) and [Point3](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/point_types.py#L103).  

Let me know if that makes sense, and thanks!

jeffin07:
@luisenp So we need to override `add`,`sub`,`mul`,`truediv`,`matmul` functions right ?

luisenp:
Yes, and there are a few others as well, but we can start with those for now.

PR #124, opened 2022-03-18
Override Vector operators in Point2 and Point3

jeffin07:
## Motivation and Context

Overrides vector operations for `Point2`, `Point3` #113 

## Types of changes

## Checklist

### DIFF
--- a/theseus/geometry/tests/point_types_mypy_check.py
+++ b/theseus/geometry/tests/point_types_mypy_check.py
@@ -1,17 +1,24 @@
 import theseus as th
 
+x = th.Point2()
+y = th.Point2()
 
-def mypy_operations_check():
+z1: th.Point2 = x + y  # noqa: F841
+z2: th.Point2 = x - y  # noqa: F841
+z3: th.Point2 = x * y  # noqa: F841
+z4: th.Point2 = x / y  # noqa: F841
+z5: th.Point2 = -x  # noqa: F841
+z6: th.Point2 = x.cat(y)  # noqa: F841
+z7: th.Point2 = x.abs()  # noqa: F841
 
-    for point_cls in [th.Point2, th.Point3]:
 
-        x = point_cls()
-        y = point_cls()
+x1 = th.Point3()
+y1 = th.Point3()
 
-        z: point_cls = x + y  # noqa: F841
-        z: point_cls = x - y  # noqa: F841
-        z: point_cls = x * y  # noqa: F841
-        z: point_cls = x / y  # noqa: F841
-        z: point_cls = -x  # noqa: F841
-        z: point_cls = x.cat(y)  # noqa: F841
-        z: point_cls = x.abs()  # noqa: F841
+w1: th.Point3 = x1 + y1  # noqa: F841
+w2: th.Point3 = x1 - y1  # noqa: F841
+w3: th.Point3 = x1 * y1  # noqa: F841
+w4: th.Point3 = x1 / y1  # noqa: F841
+w5: th.Point3 = -x1  # noqa: F841
+w6: th.Point3 = x1.cat(y1)  # noqa: F841
+w7: th.Point3 = x1.abs()  # noqa: F841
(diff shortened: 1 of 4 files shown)


jeffin07:
@luisenp overrides `add`, `sub`, `mul`, `truediv` Please review.Will add test after finalizing

luisenp (changes requested):
This looks good me, just need to add unit tests :)

mhmukadam (commented):
Looks good. Luis mentioned unit tests already.

Other functions in Vector that return Vector and need overriding (@luisenp please confirm): neg, abs, cat, _retract_impl, _compose_mpl, _inverse_impl

luisenp:
> @luisenp we're already checking for the return type [here](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/tests/test_point_types.py#L32l)
> 
> Do we need any additional unit test for the return type ?

Oh, right, we have some test that check that the dynamic type is correct. However, this PR is mostly for mypy type checking, so that test won't capture this behavior. 

Here is one way to test this. You can add the following test:

```python
def test_operators_mypy_cast():
    import mypy.api.  # Put this import at the top of the file

    result = mypy.api.run(["theseus/geometry/tests/point_types_mypy_check.py"])
    assert result[2] == 0
```

And then in file `theseus/geometry/tests/point_types_mypy_check.py` do something like the following:

```python
import theseus as th

x = th.Point2()
y = th.Point2()
z: th.Point2 = x + y
``` 

This test fails for me and it shouldn't fail after the changes in this PR.

jeffin07:
@luisenp  added test please review

luisenp:
LGTM! Let's now add `cat`, `neg`, `abs`, and `exp_map`. The functions with `_impl_` in the name are not user facing, so we don't need to fix the casting for those.

jeffin07:
@luisenp `exp_map` is already returns correct type do we need to change it ?. Also can you please help me about how to test `cat` method

jeffin07:
@luisenp  I have added the changes. Please review

mhmukadam:
LGTM. The CI is failing because mypy isn't available, might need to update the CI config.

luisenp:
@jeffin07 Can you try adding `pip install -r requirements/dev.txt` below [this](https://github.com/facebookresearch/theseus/blob/main/.circleci/config.yml#L112) line?

jeffin07:
@luisenp Added the CI changes

mhmukadam (approved):

theseus/geometry/tests/point_types_mypy_check.py
```diff
@@ -0,0 +1,17 @@
+import theseus as th
+
+
+def mypy_operations_check():
```
luisenp, line 4: For some reason `mypy` doesn't catch any errors if this function definition is here. We don't need this, so let's remove it.

theseus/geometry/tests/point_types_mypy_check.py
```diff
@@ -0,0 +1,17 @@
...
+
+    for point_cls in [th.Point2, th.Point3]:
```
luisenp, line 6: I tested this locally, and `mypy` doesn't like the syntax `z: point_cls`. I get this error:

```bash
error: Variable "theseus.geometry.tests.point_types_mypy_check.point_cls" is not valid as a type  [valid-type]
theseus/geometry/tests/point_types_mypy_check.py:9: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

I suggest to remove the loop and replace all `: point_cls` for with `: th.Point2` (and again for `: th.Point3`). Also, `mypy` doesn't like repeating the variable `z`, so let's use `z1`, `z2`, `z3`, etc.

luisenp (changes requested):
Need to fix the unit tests, but otherwise almost done. Thanks!

jeffin07:
@luisenp  Added the changes but test for cuda fails

luisenp:
@jeffin07 There seems to be some issue affecting all of our PRs right now, I'm looking into it. Your latest changes look good!

luisenp (approved):

luisenp:
@jeffin07 This looks ready to me, thanks again for your contribution!

PR #132, opened 2022-03-24
Added batching support to tactile pushing example.

luisenp:
@psodhi I added batching support, if you could take a look to double check everything that would be great!

### DIFF
--- a/examples/tactile_pose_estimation.py
+++ b/examples/tactile_pose_estimation.py
@@ -85,6 +85,9 @@ def run_learning_loop(cfg):
         step_window_moving_frame=cfg.tactile_cost.step_win_mf,
         rectangle_shape=(cfg.shape.rect_len_x, cfg.shape.rect_len_y),
         device=device,
+        optimizer_cls=getattr(th, cfg.inner_optim.optimizer),
+        max_iterations=cfg.inner_optim.max_iters,
+        regularization_w=cfg.inner_optim.reg_w,
     )
     time_steps = pose_estimator.time_steps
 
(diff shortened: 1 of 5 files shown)


examples/tactile_pose_estimation.py
```diff
@@ -108,15 +116,20 @@ def run_learning_loop(cfg):
...
-    obj_poses_gt = dataset.obj_poses[0:time_steps, :].clone().requires_grad_(True)
-    eff_poses_gt = dataset.eff_poses[0:time_steps, :].clone().requires_grad_(True)
-    theseus_inputs = {}
-    for _ in range(cfg.train.num_epochs):
+    measurements = dataset.get_measurements(time_steps)
+    for epoch in range(cfg.train.num_epochs):
+        print(" ********************* EPOCH {epoch} *********************")
         losses = []
+        image_idx = 0
         for batch_idx, batch in enumerate(measurements):
```
mhmukadam, line 124: Why is this loop over batch (and later loss.backward inside this loop)? Inner loop would be called only once per outer loop i.e. epoch, no?

mhmukadam (approved):

examples/tactile_pose_estimation.py
luisenp: This is because we can have multiple batches (e.g., 2 batches each with 4 trajectories). If we set `cfg.train.batch_size = cfg.max_episodes`, then inner loop would only be called once. This is analogous to the standard `for batch in dataloader` when training neural nets in `torch`.

PR #136, opened 2022-03-29
Merge `Between` and `VariableDiff` with `RelativePoseError` and `PosePriorError`

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/examples/pose_graph.py
+++ b/examples/pose_graph.py
@@ -17,8 +17,8 @@ num_verts, verts, edges = theg.pose_graph.read_3D_g2o_file(file_path)
 objective = th.Objective(torch.float64)
 
 for edge in edges:
-    cost_function = theg.RelativePoseError(
-        verts[edge.i], verts[edge.j], edge.relative_pose, edge.weight
+    cost_function = th.eb.Between(
+        verts[edge.i], verts[edge.j], edge.weight, edge.relative_pose
     )
     objective.add(cost_function)
 
@@ -33,16 +33,17 @@ num_verts, verts, edges = theg.pose_graph.read_3D_g2o_file(file_path)
 objective = th.Objective(torch.float64)
 
 for edge in edges:
-    cost_func = theg.RelativePoseError(
-        verts[edge.i], verts[edge.j], edge.relative_pose, edge.weight
+    cost_func = th.eb.Between(
+        verts[edge.i], verts[edge.j], edge.weight, edge.relative_pose
     )
     objective.add(cost_func)
 
-objective.add(
-    theg.PosePriorError(
-        pose=verts[0], pose_prior=verts[0].copy(new_name=verts[0].name + "PRIOR")
-    )
+pose_prior = th.eb.VariableDifference(
+    var=verts[0],
+    cost_weight=th.ScaleCostWeight(torch.tensor(1e-6, dtype=torch.float64)),
+    target=verts[0].copy(new_name=verts[0].name + "PRIOR"),
 )
+objective.add(pose_prior)
 
 optimizer = th.LevenbergMarquardt(  # GaussNewton(
     objective,
--- a/examples/pose_graph.py
+++ b/examples/pose_graph.py
@@ -11,14 +11,13 @@ import theseus.utils.examples as theg
 torch.manual_seed(1)
 
 file_path = "datasets/tinyGrid3D.g2o"
+dtype = torch.float64
 
 th.SO3.SO3_EPS = 1e-6
 
-num_verts, verts, edges = theg.pose_graph.read_3D_g2o_file(
-    file_path, dtype=torch.float32
-)
+num_verts, verts, edges = theg.pose_graph.read_3D_g2o_file(file_path, dtype=dtype)
 
-objective = th.Objective(torch.float32)
+objective = th.Objective(dtype)
 
 for edge in edges:
     cost_func = th.eb.Between(
@@ -28,7 +27,7 @@ for edge in edges:
 
 pose_prior = th.eb.VariableDifference(
     var=verts[0],
-    cost_weight=th.ScaleCostWeight(torch.tensor(1e-6, dtype=torch.float32)),
+    cost_weight=th.ScaleCostWeight(torch.tensor(1e-6, dtype=dtype)),
     target=verts[0].copy(new_name=verts[0].name + "PRIOR"),
 )
 objective.add(pose_prior)
--- a/examples/pose_graph.py
+++ b/examples/pose_graph.py
@@ -12,25 +12,13 @@ torch.manual_seed(1)
 
 file_path = "datasets/tinyGrid3D.g2o"
 
-num_verts, verts, edges = theg.pose_graph.read_3D_g2o_file(file_path)
-
-objective = th.Objective(torch.float64)
-
-for edge in edges:
-    cost_function = th.eb.Between(
-        verts[edge.i], verts[edge.j], edge.weight, edge.relative_pose
-    )
-    objective.add(cost_function)
-
-torch.manual_seed(1)
-
-file_path = "datasets/tinyGrid3D.g2o"
-
 th.SO3.SO3_EPS = 1e-6
 
-num_verts, verts, edges = theg.pose_graph.read_3D_g2o_file(file_path)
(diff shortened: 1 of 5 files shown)


mhmukadam (commented):
Looks good, only one comment.

theseus/geometry/se3.py
```diff
@@ -22,12 +22,13 @@ def __init__(
...
         name: Optional[str] = None,
         dtype: Optional[torch.dtype] = None,
+        requires_check: bool = True,
     ):
         if x_y_z_quaternion is not None and data is not None:
             raise ValueError("Please provide only one of x_y_z_quaternion or data.")
         if x_y_z_quaternion is not None:
             dtype = x_y_z_quaternion.dtype
-        if data is not None:
+        if data is not None and requires_check:
```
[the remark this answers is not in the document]
fantaosha: Yes, if `data` is used for initialization, the default `dtype` will be `data.dtype`.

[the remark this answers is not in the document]
mhmukadam: Ah yes super goes all the way back to `Variable` where it is assigned. Sounds good.

examples/pose_graph.py
```diff
@@ -12,37 +12,26 @@
...
+    cost_func = th.eb.Between(
+        verts[edge.i], verts[edge.j], edge.weight, edge.relative_pose
     )
     objective.add(cost_func)
 
-objective.add(
-    theg.PosePriorError(
-        pose=verts[0], pose_prior=verts[0].copy(new_name=verts[0].name + "PRIOR")
-    )
+pose_prior = th.eb.VariableDifference(
```
luisenp, line 29: I'm confused about this cost function. Is this a constraint to make the initial pose exactly some value? Why is the weight so small?

theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -47,7 +49,7 @@ def read_3D_g2o_file(path: str) -> Tuple[int, List[th.SE3], List[PoseGraphEdge]]
 
                 x_y_z_quat = torch.from_numpy(
                     np.array([tokens[3:10]], dtype=np.float64)
-                )
+                ).to(dtype)
```
luisenp, line 52: Why not change the line above to do `, dtype=dtype)`?

theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -56,6 +58,7 @@ def read_3D_g2o_file(path: str) -> Tuple[int, List[th.SE3], List[PoseGraphEdge]]
                 sel = [0, 6, 11, 15, 18, 20]
                 weight = th.Variable(
                     torch.from_numpy(np.array(tokens[10:], dtype=np.float64)[sel])
+                    .to(dtype)
```
luisenp, line 61: Same comment about `, dtype=dtype)`

theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -74,7 +77,9 @@ def read_3D_g2o_file(path: str) -> Tuple[int, List[th.SE3], List[PoseGraphEdge]]
             elif tokens[0] == "VERTEX_SE3:QUAT":
                 i = int(tokens[1])
 
-                x_y_z_quat = torch.from_numpy(np.array([tokens[2:]], dtype=np.float64))
+                x_y_z_quat = torch.from_numpy(
+                    np.array([tokens[2:]], dtype=np.float64)
+                ).to(dtype)
```
luisenp, line 82: Same comment about `, dtype=dtype)`

theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -113,14 +120,17 @@ def read_2D_g2o_file(path: str) -> Tuple[int, List[th.SE2], List[PoseGraphEdge]]
 
                 n = len(edges)
 
-                x_y_theta = torch.from_numpy(np.array([tokens[3:6]], dtype=np.float64))
+                x_y_theta = torch.from_numpy(
+                    np.array([tokens[3:6]], dtype=np.float64)
```
luisenp, line 124: Same comment about `, dtype=dtype)`

theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -113,14 +120,17 @@ def read_2D_g2o_file(path: str) -> Tuple[int, List[th.SE2], List[PoseGraphEdge]]
...
+                ).to(dtype)
                 relative_pose = th.SE2(
                     x_y_theta=x_y_theta, name="EDGE_SE2__{}".format(n)
                 )
 
                 sel = [0, 3, 5]
                 weight = th.Variable(
                     torch.from_numpy(np.array(tokens[6:], dtype=np.float64)[sel])
+                    .to(dtype)
```
luisenp, line 133: Same comment about `, dtype=dtype)`

theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -139,7 +149,9 @@ def read_2D_g2o_file(path: str) -> Tuple[int, List[th.SE2], List[PoseGraphEdge]]
             elif tokens[0] == "VERTEX_SE2:QUAT":
                 i = int(tokens[1])
 
-                x_y_theta = torch.from_numpy(np.array([tokens[2:]], dtype=np.float64))
+                x_y_theta = torch.from_numpy(
```
luisenp, line 152: Same comment about `, dtype=dtype)`

luisenp (commented):
Looks good! Just left a few comments and a question.

theseus/utils/examples/pose_graph/dataset.py
fantaosha: It looks like [`torch.from_numpy`](https://pytorch.org/docs/stable/generated/torch.from_numpy.html) has no `dtype` options.

examples/pose_graph.py
fantaosha: Correct. The only reason we introduce `pose_prior` is to make the linear system not ill-conditioned. A too large weight will restrict the search space and yield slow convergence.

fantaosha: `GTSAM` also has similar factors [[link](https://github.com/borglab/gtsam/blob/0850e89b0e6148d16f00b7741d83b6325ada977d/examples/Pose2SLAMExample.cpp#L73)].

luisenp: If I understand that code, they are anchoring that variable to a zero/origin value, which is not the case here? Here the target is set to whatever value `verts[0]` has at this point.

theseus/utils/examples/pose_graph/dataset.py
luisenp: But there is a `dtype` inside the `np.array` which we can use. I think torch will respect the type of the underlying numpy array.

fantaosha: I know. But the `dtype` in `np.array()` does not accept types like `torch.float32`.

examples/pose_graph.py
fantaosha: I don't think the `target` value matters. In fact, the `target` value can be anything. In PGO problems, we mainly consider the relative pose errors, which, however, has no constraints on the absolute pose. Thus, we introduce pose prior add absolute pose constraint.

The reason why I use `verts[0]` is that it is closed to the initial guess.

theseus/utils/examples/pose_graph/dataset.py
luisenp: Why can't you do something like this

image

examples/pose_graph.py
luisenp: I see. Thanks for the explanation!

theseus/utils/examples/pose_graph/dataset.py
fantaosha: Yes, I can do this. But `dtype` is one the input of [`read_3D_g2o_file`](https://github.com/facebookresearch/theseus/blob/bed363f5c1643ffa146c57d0ddde80dd5e28fc5b/theseus/utils/examples/pose_graph/dataset.py#L32). Is it OK to mix the use of `torch.dtype` and `np.dtype`?

mhmukadam (approved):

PR #138, opened 2022-04-01
Add robust cost function

fantaosha:
## Types of changes

## Checklist


theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,56 @@
...
+        return self.cost_function.error()
+
+    def jacobians(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        return self.cost_function.jacobians()
+
+    def weighted_error(self) -> torch.Tensor:
+        return self.cost_function.weighted_error()
+
+    def weighted_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        return self.weighted_jacobians_error()
```
mhmukadam, line 39: Is this returning itself? Could be a bug.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,56 @@
...
+
+from .loss import Loss
+
+
+class RobustCostFunction(th.CostFunction):
+    def __init__(
+        self,
+        loss_function: Loss,
+        log_loss_radius: th.Vector,
+        cost_function: th.CostFunction,
```
mhmukadam, line 20: Is the idea of this implementation that any cost function can be turned into its robust version?

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import List, Tuple
+
+import torch
+
+import theseus as th
+
+from .loss import Loss
+
+
+class RobustCostFunction(th.CostFunction):
```
mhmukadam, line 15: Add unit testing for this.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,56 @@
...
+    def weighted_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        return self.weighted_jacobians_error()
+
+    def value(self) -> torch.Tensor:
+        weighted_error = self.weighted_error()
+        squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
+        loss_radius = torch.exp(self.log_loss_radius.data)
+        return self.loss_function.evaluate(squared_norm, loss_radius)
+
+    def rescaled_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
```
mhmukadam, line 47: I am wondering if we can simply the API here and avoid introducing this new function.

If `weighted_jacobians_error()` is overridden for `RobustCostFunction` to do what `rescaled_jacobians_error()` is doing then you don't need a separate new function for this.

theseus/core/robust_cost_function.py
fantaosha: Yes.

theseus/core/robust_cost_function.py
fantaosha: Will do. I'm not considering to merge this branch to main. This is only for profiling and benchmarking.

theseus/core/robust_cost_function.py
fantaosha: Since this feature is not ready to merge to main and has not been fully tested, I just implemented another function for simplification.

theseus/core/objective.py
```diff
@@ -348,15 +348,13 @@ def error(
                 for var in self.optim_vars:
                     old_data[var] = self.optim_vars[var].data
             self.update(input_data=input_data)
-        error_vector = torch.zeros(self.batch_size, self.dim()).to(
+        error_vector = torch.zeros(self.batch_size, len(self.cost_functions)).to(
```
mhmukadam, line 351: Does `len(self.cost_functions)` give the number (n) of cost functions? I think we want \sum_i{cost_i.dim}, where i=1,..,n.

theseus/core/loss.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import abc
+
+import torch
+
+
+class Loss:
```
mhmukadam, line 11: Loss might be a confusing name here given the outer-loop training and learning related terminology.

Also if this is only relevant to `RobustCostFunction` we can probably move all this to that file.

mhmukadam (commented):
Left some comments and discussion points.

theseus/core/loss.py
fantaosha: How about `LossFunction`?

theseus/core/cost_function.py
```diff
@@ -57,6 +57,13 @@ def weighted_jacobians_error(
         jacobian, err = self.jacobians()
         return self.weight.weight_jacobians_and_error(jacobian, err)
 
+    def value(self) -> torch.Tensor:
```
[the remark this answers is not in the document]
fantaosha: It calculates the subobjective value of `CostFunction`.

theseus/core/loss.py
mhmukadam: `RobustCost`?
That way `RobustCost` + `CostFunction` -> `RobustCostFunction` would make sense too.

fantaosha:
@luisenp @mhmukadam I completely rewrote robust cost functions. I merged `RobustCostFunction` to `CostFunction`. Maybe we need a discussion about this.

fantaosha:
@luisenp @mhmukadam I completely rewrote robust cost functions. I merged `RobustCostFunction` to `CostFunction`. Maybe we need a discussion about this.

theseus/core/loss_function.py
```diff
@@ -0,0 +1,158 @@
...
+        pass
+
+    @abc.abstractmethod
+    def rescale(
+        self,
+        jacobians: List[torch.Tensor],
+        error: torch.Tensor,
+    ) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        pass
+
```
mhmukadam, line 31: Add a `_copy_impl` abstract method here like [this](https://github.com/facebookresearch/theseus/blob/main/theseus/core/cost_function.py#L62).

theseus/core/loss_function.py
```diff
@@ -0,0 +1,158 @@
...
+    def copy(
+        self, new_name: Optional[str] = None, keep_variable_names: bool = False
+    ) -> "TrivialLoss":
+        return cast(
+            TrivialLoss,
+            super().copy(new_name=new_name, keep_variable_names=keep_variable_names),
+        )
+
+
+class RobustLoss(LossFunction):
```
mhmukadam, line 68: Since you have abstract methods in this class
`class RobustLoss(LossFunction, abc.ABC)`

theseus/embodied/collision/collision.py
```diff
@@ -23,11 +23,12 @@ def __init__(
         sdf_data: Variable,
         sdf_cell_size: Variable,
         cost_eps: Variable,
+        loss_function: Optional[LossFunction] = None,
```
mhmukadam, line 26: What's the behavior if we don't add this here. The super call would not have anything so the base class would assign trivial loss?

mhmukadam (commented):
Looks good. We should add units tests before we merge to main.

theseus/core/loss_function.py
fantaosha: Will do.

theseus/embodied/collision/collision.py
fantaosha: That's correct. We can skip `loss_function`, and then `collision` uses `TrivialLoss`.

fantaosha:
> Looks good. We should add units tests before we merge to main.

@luisenp @mhmukadam Do you have any ideas about what unit tests we should have?

mhmukadam:
> @luisenp @mhmukadam Do you have any ideas about what unit tests we should have?

You could add mock tests similar to ones for cost function, and also augment the current tests for objective, etc to account for the changes (checking add/erase, etc). Maybe Luis also has some suggestions.

fantaosha:
Merged to `experimental`.

PR #148, opened 2022-04-08
Implement Robust Cost Function

luisenp:
This replaces #138. It reuses some of the code in that PR (and all of the math implemented there) but reorganizes the classes differently. I went back to the version that had `RobustCostFunction` as a separate class and `Loss` as a basic functional form.


luisenp:
This should work after #205 is merged.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,102 @@
+import abc
```
mhmukadam, line 1: Missing license header

theseus/__init__.py
```diff
@@ -7,11 +7,15 @@
     CostFunction,
     CostWeight,
     DiagonalCostWeight,
-    AutoDiffCostFunction,
     Objective,
     ScaleCostWeight,
     Variable,
     Vectorize,
+    Loss,
```
mhmukadam, line 14: Have we settled on Loss or are there any better alternatives? Loss good lead to some confusion of loss in the ML context.

theseus/core/loss.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import abc
+
+import torch
+
+
+class Loss:
```
mhmukadam, line 11: We should add some unit tests for loss and robust loss.

mhmukadam: This is an abstract class. Need to do `class Loss(abc.ABC):`

theseus/core/loss.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @staticmethod
+    @abc.abstractmethod
+    def linearize(x: torch.Tensor, raidus: torch.Tensor) -> torch.Tensor:
+        pass
+
+
+class WelschLoss(Loss):
+    @staticmethod
+    def evaluate(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
```
mhmukadam, line 25: Are there restrictions on the sizes of x and/or radius? Should we explicitly check for batch? If these are common checks then we can move it to the base class.

theseus/core/loss.py
```diff
@@ -0,0 +1,42 @@
...
+        return radius - radius * torch.exp(-x / radius)
```
mhmukadam, line 26: Not sure if this is needed but do we raise a warning if radius is zero/close to zero. Incase you are learning it, there is a chance this fails silently by sending nan/infs.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,102 @@
...
+# (see Theory section at http://ceres-solver.org/nnls_modeling.html#theory and
+# references therein); here we use alpha=0. The implementation of this part is done via
+# `RobustCostFunction.weighted_jacobians_error()`
+#
+# Let e := `robust_cost_fn.cost_function.error()`. Currently, the convention is that:
+#     -`robust_cost_fn.error()` returns e.
+#     -`robust_cost_fn.weighted_error()` returns a vectorized version of loss(||e||^2).
+#
+# Also, `robust_cost_fn.jacobians()` is not implemented.
+class RobustCostFunction(CostFunction, abc.ABC):
```
mhmukadam, line 28: I don't think this is an abstract class so we can remove `abc.ABC`.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,102 @@
...
+        # ||weighted_error||^2 = error_loss
+        # By doing this we avoid having to change the objective's error computation
+        # specifically for robust cost functions. The issue for this type of cost
+        # function is that the theory requires us to maintain scaled errors/jacobians
+        # of dim = robust_fn.cost_function.dim() to do the linearization properly,
+        # but the actual error has dim = 1, being the result of loss(||error||^2).
+        #
+        # Other options explored so far involve adding new methods to CostFunction,
+        # and/or changing Objective/Optimizer class. I'd prefer to avoid changing
+        # core class for experimental code, as long as it is possible [lep].
```
mhmukadam, line 72: Are these comments still needed (no longer experimental code)? Maybe needs some revision at least.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,102 @@
...
+        return (
+            torch.ones_like(weighted_error) * (error_loss / self.dim() + 1e-20).sqrt()
+        )
+
+    def jacobians(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        raise NotImplementedError
```
mhmukadam, line 78: Is the unweighted Jacobian not need anywhere? Maybe add a comment.

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,102 @@
...
+        # specifically for robust cost functions. The issue for this type of cost
+        # function is that the theory requires us to maintain scaled errors/jacobians
+        # of dim = robust_fn.cost_function.dim() to do the linearization properly,
+        # but the actual error has dim = 1, being the result of loss(||error||^2).
+        #
+        # Other options explored so far involve adding new methods to CostFunction,
+        # and/or changing Objective/Optimizer class. I'd prefer to avoid changing
+        # core class for experimental code, as long as it is possible [lep].
+        return (
+            torch.ones_like(weighted_error) * (error_loss / self.dim() + 1e-20).sqrt()
```
mhmukadam, line 74: Hardcoded number, change to some EPS?

theseus/core/robust_cost_function.py
```diff
@@ -0,0 +1,102 @@
...
+        for attr in cost_function._aux_vars_attr_names:
+            setattr(self, attr, getattr(cost_function, attr))
+            self.register_aux_var(attr)
+
+        self.log_loss_radius = log_loss_radius
+        self.register_aux_var("log_loss_radius")
+        self.loss = loss_cls()
+
+    def error(self) -> torch.Tensor:
+        return self.cost_function.error()
```
mhmukadam, line 54: Should this return the robustified error (just not weighted)?

PR #150, opened 2022-04-10
Allow constant inputs to cost functions to be passed as floats

### LINKED ISSUE #38: Allow constant inputs to cost functions to be passed as floats
...and converted to `Variable` objects internally. 

This includes things like `dt` in GP cost function, and `cost_eps` or `cell_size` in `Collision2D`.

jeffin07:
## Motivation and Context
To close #38 

## Types of changes

## Checklist


jeffin07:
@luisenp @mhmukadam I have added some  changes to `collisions.py` and  `double_integrator.py`. Tests not included. I will add test once the feature is finalized.Also i do see lot of `variables` in other cost functions,Could you please tell me which one are elligible for this change ?

theseus/embodied/collision/collision.py
```diff
@@ -32,7 +32,12 @@ def __init__(
         self.sdf_origin = sdf_origin
         self.sdf_data = sdf_data
         self.sdf_cell_size = sdf_cell_size
-        self.cost_eps = cost_eps
+        if not isinstance(cost_eps, Variable):
+            if not isinstance(cost_eps, torch.Tensor):
+                cost_eps = torch.tensor(cost_eps)
```
mhmukadam, line 37: Likely need to add batch size here (and in rest of cost functions) same as pose.

theseus/embodied/motionmodel/double_integrator.py
```diff
@@ -28,6 +28,12 @@ def __init__(
             raise ValueError(
                 "All variables for a DoubleIntegrator must have the same dimension."
             )
+        if not isinstance(dt, Variable):
+            if not isinstance(dt, torch.Tensor):
+                dt = torch.tensor(dt)
+            self.dt = Variable(dt)
+        else:
+            self.dt = dt
```
luisenp, line 36: After the `ndim` check below, we can add `dt.data = self.dt.data.view(-1, 1)`, to make sure we have a batch dimension as well, as discussed above.

theseus/embodied/motionmodel/double_integrator.py
```diff
@@ -96,11 +101,13 @@ class GPCostWeight(CostWeight):
...
         self,
         Qc_inv: Union[Variable, torch.Tensor],
-        dt: Union[Variable, torch.Tensor],
+        dt: Union[float, Variable, torch.Tensor],
         name: Optional[str] = None,
     ):
         super().__init__(name=name)
         if not isinstance(dt, Variable):
+            if not isinstance(dt, torch.Tensor):
+                dt = torch.tensor(dt)
```
luisenp, line 110: I have added a fix for this class in #157, make sure you merge those changes as well.

theseus/embodied/motionmodel/double_integrator.py
```diff
@@ -180,6 +187,12 @@ def __init__(
                 "GPMotionModel only accepts cost weights of type GPCostWeight. "
                 "For other weight types, consider using DoubleIntegrator instead."
             )
+        if not isinstance(dt, Variable):
+            if not isinstance(dt, torch.Tensor):
+                dt = torch.tensor(dt)
+            self.dt = Variable(dt)
+        else:
+            self.dt = dt
         super().__init__(pose1, vel1, pose2, vel2, dt, cost_weight, name=name)
```
luisenp, line 196: Similar comment about maintaining a batch shape here.

luisenp (changes requested):
This is looking good! Some tweaks needed but mostly easy fixes and unit tests. The only other variable I've seen so far where this applies is `eff_radius` in `EffectorObjectContactPlanar`. Poses and sdf variables (origin, cell size, data) are multidimensional, so no need to change those.

theseus/embodied/collision/collision.py
mhmukadam: Thanks for clarifying Luis. @jeffin07, the batch size here you can get from the passed pose object.
Nice catch on the bug as well, Luis.

jeffin07:
@luisenp Added the changes. Please review.Also do you have any suggestions for unit test ?

theseus/embodied/collision/collision.py
luisenp: It's probably better to use `(-1, 1)` instead of getting the batch size from the pose object, because `dt` is an element that we typically set a batch size of 1 for in our examples, regardless of the pose batch.

luisenp (changes requested):
LGTM! For unit tests, just constructing all of these classes with the different types of inputs, and checking that the corresponding member is a variable of the expected type and shape should be sufficient. Thanks!!

theseus/embodied/collision/tests/test_eff_obj_contact.py
```diff
@@ -149,3 +150,37 @@ def test_eff_obj_interesect_errors():
...
+        origin = th.Variable(torch.randn(batch_size, 2).double())
+        sdf_data = th.Variable(torch.randn(batch_size, 10, 10).double())
+        cell_size = th.Variable(torch.rand(batch_size, 1).double())
+        eff_radius = th.Variable(torch.rand(batch_size, 1).double())
+        cost_weight = th.ScaleCostWeight(1.0)
+        cost_function = th.eb.EffectorObjectContactPlanar(
+            obj, eff, cost_weight, origin, sdf_data, cell_size, eff_radius
+        )
+
+        assert isinstance(cost_function.eff_radius, Variable)
```
luisenp, line 170: Here we can check that `cost_function.eff_radius is eff_radius`.

theseus/embodied/collision/tests/test_eff_obj_contact.py
```diff
@@ -149,3 +150,37 @@ def test_eff_obj_interesect_errors():
...
+
+        eff_radius_t = torch.rand(batch_size, 1).double()
+
+        cost_function = th.eb.EffectorObjectContactPlanar(
+            obj, eff, cost_weight, origin, sdf_data, cell_size, eff_radius_t
+        )
+
+        assert isinstance(cost_function.eff_radius, Variable)
+
```
luisenp, line 179: Here we can add another check that `cost_function.eff_radius.data is eff_radius_t`.

theseus/embodied/collision/tests/test_eff_obj_contact.py
```diff
@@ -149,3 +150,37 @@ def test_eff_obj_interesect_errors():
...
+        eff_radius_f = 1.0
+
+        cost_function = th.eb.EffectorObjectContactPlanar(
+            obj, eff, cost_weight, origin, sdf_data, cell_size, eff_radius_f
+        )
+
+        assert isinstance(cost_function.eff_radius, Variable)
```
luisenp, line 186: Here we can first set up some random value for `eff_radius_f` and then add an additional check that `np.isclose(cost_function.eff_radius.data.item(), eff_radius_f)`.

theseus/embodied/motionmodel/tests/test_double_integrator.py
```diff
@@ -60,6 +61,30 @@ def test_gp_motion_model_cost_weight_copy():
     )
 
 
+def test_gp_motion_model_variable_type():
```
luisenp, line 64: Add similar modifications to the test here, as mentioned above.

theseus/embodied/motionmodel/tests/test_double_integrator.py
```diff
@@ -60,6 +61,30 @@ def test_gp_motion_model_cost_weight_copy():
...
+            assert isinstance(cost_weight.Qc_inv, Variable)
+            assert isinstance(cost_weight.dt, Variable)
+
+            q_inv_v = Variable(q_inv)
+            dt_v = Variable(dt)
+            cost_weight = th.eb.GPCostWeight(q_inv_v, dt_v)
+            assert isinstance(cost_weight.Qc_inv, Variable)
+            assert isinstance(cost_weight.dt, Variable)
+
+            q_inv_v = Variable(q_inv)
```
luisenp, line 81: We don't need the second check for `q_inv_v`.

luisenp (changes requested):
Looks good, almost there! Once the code is finished, I'll also run some tests on my side before confirming that we can merge. Thanks!

theseus/embodied/collision/tests/test_eff_obj_contact.py
jeffin07: @luisenp 
we're already giving a random value to `eff_radius_f=1.0`
So  is this enough ` assert np.isclose(cost_function.eff_radius.data.item(), eff_radius_f)`

theseus/embodied/collision/tests/test_eff_obj_contact.py
jeffin07: @luisenp While `assert cost_function.eff_radius.data is eff_radius_t` gives error as one is `tensor` and other is `th.Variable` .Is it better to use  `assert torch.allclose(cost_function.eff_radius.data,eff_radius_t)` ?

theseus/embodied/collision/tests/test_eff_obj_contact.py
luisenp: By random I mean sampling a new value every iteration rather than hardcoding to 1.0.

jeffin07:
@luisenp Added changes please review

jeffin07:
@luisenp @mhmukadam I'm unable to run test after fetching latest chainges from `upstream/main`. When running `pytest theseus/` gives the following output. Any idea what I've done wrong ?

```
ImportError while importing test module '/theseus/theseus/optimizer/tests/test_sparse_linearization.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
theseus/optimizer/tests/test_sparse_linearization.py:9: in <module>
    import theseus as th
theseus/__init__.py:71: in <module>
    import theseus.embodied as eb
theseus/embodied/__init__.py:6: in <module>
    from .collision import Collision2D, EffectorObjectContactPlanar, SignedDistanceField2D
theseus/embodied/collision/__init__.py:6: in <module>
    from .collision import Collision2D
theseus/embodied/collision/collision.py:11: in <module>
    from theseus.embodied.kinematics import IdentityModel, KinematicsModel
theseus/embodied/kinematics/__init__.py:6: in <module>
    from .kinematics_model import IdentityModel, KinematicsModel, UrdfRobotModel
theseus/embodied/kinematics/kinematics_model.py:9: in <module>
    import differentiable_robot_model as drm
E   ModuleNotFoundError: No module named 'differentiable_robot_model'





```

This keeps repating for every file

theseus/embodied/collision/tests/test_eff_obj_contact.py
jeffin07: @luisenp when i try to do `assert cost_function.eff_radius.data is eff_radius_t` i get this error which expained above 

```
>           assert cost_function.eff_radius.data is eff_radius_t
E           assert tensor([[0.4160]], dtype=torch.float64) is tensor([[0.4160]], dtype=torch.float64)
E            +  where tensor([[0.4160]], dtype=torch.float64) = Variable(data=tensor([[0.4160]], dtype=torch.float64), name=Variable__680).data
E            +    where Variable(data=tensor([[0.4160]], dtype=torch.float64), name=Variable__680) = <theseus.embodied.collision.eff_obj_contact.EffectorObjectContactPlanar object at 0x7f813704edd0>.eff_radius

theseus/embodied/collision/tests/test_eff_obj_contact.py:180: AssertionError


```

luisenp: Ah, right, it's because of the `view(-1, 1)` we added. In this case, let's stick to compare that they are close in value as you had initially.

jeffin07: @luisenp Then can you take a look at PR. I've already committed the changes

theseus/embodied/collision/tests/test_eff_obj_contact.py
luisenp: You can delete the line above that has the `isinstance` check now.

theseus/embodied/motionmodel/tests/test_double_integrator.py
```diff
@@ -71,17 +71,19 @@ def test_gp_motion_model_variable_type():
...
             assert isinstance(cost_weight.dt, Variable)
+            assert cost_weight.Qc_inv is q_inv_v
+            assert cost_weight.dt is dt_v
 
             q_inv_v = Variable(q_inv)
-            dt_f = 1.0
+            dt_f = torch.rand(1)
             cost_weight = th.eb.GPCostWeight(q_inv_v, dt_f)
-            assert isinstance(cost_weight.Qc_inv, Variable)
             assert isinstance(cost_weight.dt, Variable)
```
luisenp, line 87: Sorry, didn't think of this before, but we should also check that `cost_weight.dt` has a batch dimension. And similar for `eff_radius_t` in the other cost function. We are also missing the check `np.isclose(cost_function.eff_radius.data.item(), eff_radius_f)` (and similar for `dt`).

jeffin07:
@luisenp Updated unit tests

theseus/embodied/motionmodel/tests/test_double_integrator.py
luisenp: @jeffin07 Missing the `np.allclose()` check for `eff_radius_f` and also checking that the batch dimension is present.

jeffin07:
@luisenp Updated the changes.Please review

luisenp (approved):
Sorry for the delay, got busy with other PRs and forgot this was waiting for review. LGTM!

luisenp:
Thanks a lot for another great contribution, @jeffin07 :)

theseus/embodied/collision/tests/test_eff_obj_contact.py
```diff
@@ -8,6 +8,7 @@
 import torch
 
 import theseus as th
+from theseus.core import Variable
```
mhmukadam, line 11: Don't need this since it is already available as `th.Variable`

mhmukadam (approved):
Looks good! Thanks.

theseus/embodied/collision/tests/test_eff_obj_contact.py
jeffin07: @mhmukadam Do i need to change this ?

jeffin07:
@luisenp @mhmukadam Updated the changes. Please review

PR #155, opened 2022-04-12
Made AutoDiffCostFunction._tmp_optim_vars copies of original

luisenp:
This preserves optimization variables types and can solve errors like [this](https://github.com/facebookresearch/theseus/pull/84#discussion_r847801204). 

Besides running unit tests, I also ran the `backward_modes.py` script and the first tutorial. Seems to be working w/o issues after this change. 

cc @exhaustin

### DIFF
--- a/theseus/core/cost_function.py
+++ b/theseus/core/cost_function.py
@@ -135,7 +135,7 @@ class AutoDiffCostFunction(CostFunction):
 
         # The following are auxiliary Variable objects to hold tensor data
         # during jacobian computation without modifying the original Variable objects
-        self._tmp_optim_vars = tuple(Variable(data=v.data) for v in optim_vars)
+        self._tmp_optim_vars = tuple(v.copy() for v in optim_vars)
 
     def _compute_error(
         self,


exhaustin:
I don't have much context regarding the effect of this change, but thanks for fixing!

mhmukadam (approved):

theseus/core/cost_function.py
```diff
@@ -135,7 +135,7 @@ def _register_vars_in_list(var_list_, is_optim=False):
 
         # The following are auxiliary Variable objects to hold tensor data
         # during jacobian computation without modifying the original Variable objects
-        self._tmp_optim_vars = tuple(Variable(data=v.data) for v in optim_vars)
+        self._tmp_optim_vars = tuple(v.copy() for v in optim_vars)
```
mhmukadam, line 138: This makes sense. I dug again into `Variable.copy` to find that `data.clone()` is called. Want to make sure that the intent here is to have `_tmp_optim_vars` connected to the compute graph.

More generally, is this okay or should we actually be doing `data.detach().clone()`?

luisenp: More generally, I don't think we should be doing `detach().clone()` in our copy methods, but maybe we should add a flag that controls for this.

mhmukadam: Maybe we can provide a separate detach() function? If it makes sense we can flag an issue for this for now.

When you call Objective.clone() to get a new objective, is it still connected to the previous compute graph?

mhmukadam: Based on offline conversation we might revisit the detach idea later since it could be tricky to setup, what should be just cloned vs also detached when calling everything recursively from Objective.

luisenp: I guess it depends on where you call it. If you do something like the code below, I'm almost sure that the resulting graph would be connected.

```python
objective = create_objective()
optimizer = SomeOptimizer(objective)
layer = TheseusLayer(objective)
input_dict = compute_some_initial_vars()
sol_dict = layer.forward(input_dict)
new_objective = objective.copy()
new_optimizer = SomeOptimizer(new_objective)
layer2 = TheseusLayer(new_optimizer)
sol = layer2.forward()  # no input here, keep whatever variables you cloned
loss =  do_some_stuff(sol)
loss.backward()
```

PR #161, opened 2022-04-13
DLM gradients

rtqichen:
## Motivation and Context

Implements support for the "direct loss minimization" gradient computation for theseus.

## How Has This Been Tested

Tested in `theseus/optimizer/nonlinear/tests/test_backwards.py`, and ran the following examples:
- `examples/backward_mode.py` 
- `examples/tactile_pose_estimation.py`

## Types of changes

## Checklist

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -307,7 +308,7 @@ def _optimize_impl(
                 f"Error: {info.last_err.mean().item()}"
             )
 
-        if backward_mode == BackwardMode.FULL:
+        if backward_mode in [BackwardMode.FULL, BackwardMode.DLM]:
```
mhmukadam, line 311: Should DLM be handled separately where you have `with torch.no_grad()`?

examples/backward_modes.py
```diff
@@ -128,6 +128,22 @@ def quad_error_fn(optim_vars, aux_vars):
 print(da_dx.numpy())
 
 
+# We can also compute the direct loss minimization gradient.
+updated_inputs, info = theseus_optim.forward(
```
mhmukadam, line 132: Example looks great! Probably also add to unit tests like [here](https://github.com/facebookresearch/theseus/blob/8741e372d07252e1a14ab0fdc7afebcf23f0557e/theseus/optimizer/nonlinear/tests/test_backwards.py#L65).

theseus/theseus_layer.py
```diff
@@ -5,11 +5,17 @@
 
 from typing import Any, Dict, Optional, Tuple
 
+from functools import partial
+
 import torch
 import torch.nn as nn
+from torch.autograd.function import once_differentiable
 
+from theseus import Variable, GaussNewton
```
mhmukadam, line 14: Our convention is to use relative imports in lib files and theseus global imports in tests and examples.
`from theseus.core import Variable`
`from theseus.optimizer.nonlinear import GaussNewton`

theseus/theseus_layer.py
```diff
@@ -32,15 +38,19 @@ def forward(
...
         optimizer_kwargs = optimizer_kwargs or {}
-        info = self.optimizer.optimize(**optimizer_kwargs)
-        values = dict(
-            [
-                (var_name, var.data)
-                for var_name, var in self.objective.optim_vars.items()
-            ]
-        )
+        backward_mode = optimizer_kwargs.get("backward_mode", None)
+        DLM_epsilon = optimizer_kwargs.get("DLM_epsilon", 1e-2)
```
mhmukadam, line 43: nit: dlm_epsilon (lower case with underscores for any object or class name)

theseus/theseus_layer.py
```diff
@@ -93,3 +103,88 @@ def device(self) -> torch.device:
...
+
+
+def _forward(objective, optimizer, optimizer_kwargs, input_data):
+    objective.update(input_data)
+    info = optimizer.optimize(**optimizer_kwargs)
+    vars = [var.data for var in objective.optim_vars.values()]
+    return vars, info
+
+
+class TheseusLayerDLMForward(torch.autograd.Function):
```
mhmukadam, line 115: Overall looks good. Once the functionality is settled and unit tested we should think about compartmentalizing this maybe as separate files/modules outside of TheseusLayer.

mhmukadam: @bamos, @luisenp, Joe will likely add more things in this vein for implicit differentiation for belief propagation - GBP (which will branch off from NonlinearOptimizer). So we should think a more general API here if possible.

theseus/theseus_layer.py
```diff
@@ -93,3 +103,88 @@ def device(self) -> torch.device:
...
+        # Can we put all of this into a single cost function?
+        for i, (name, var) in enumerate(bwd_objective.optim_vars.items()):
+            grad_var = Variable(grad_outputs[i], name=name + "_grad")
+            bwd_objective.add(AutoDiffCostFunction(
+                [var],
+                partial(_dlm_perturbation, epsilon=epsilon),
+                1,
+                aux_vars=[grad_var],
+                name="DLM_perturbation_" + name,
+            ))
```
mhmukadam, line 168: Like discussed offline, we should be able to build this once and have it saved.

theseus/theseus_layer.py
```diff
@@ -93,3 +103,88 @@ def device(self) -> torch.device:
...
+        objective = ctx.objective
+        epsilon = ctx.epsilon
+
+        # Update the optim vars to their solutions.
+        input_data = ctx.input_data
+        values = dict(zip(objective.optim_vars.keys(), optim_vars))
+        input_data.update(values)
+
+        # Construct backward objective.
+        bwd_objective = objective.copy()
```
mhmukadam, line 157: Like discussed offline, we should be able to build bwd_objective once first time DLM is triggered and then have it saved for future calls.

theseus/theseus_layer.py
```diff
@@ -93,3 +103,88 @@ def device(self) -> torch.device:
...
+
+        # Can we put all of this into a single cost function?
```
mhmukadam, line 159: Yeah you should be able to vectorize this and make one cost function. Or you could leave it this way and the automatic cost function grouping (to be implemented) will take care of vectorization.

mhmukadam (commented):
Look good so far. After adding discussed things and adding unit tests, you can try with the [tactile example](https://github.com/facebookresearch/theseus/blob/main/examples/tactile_pose_estimation.py) which we are using for the backward experiments.

mhmukadam:
Also CI fails linting. Make sure you are [using git precommit hooks](https://github.com/facebookresearch/theseus/blob/main/CONTRIBUTING.md#code-style).

bamos (approved):

theseus/theseus_layer.py
```diff
@@ -93,3 +103,88 @@ def device(self) -> torch.device:
...
+
+
+class TheseusLayerDLMForward(torch.autograd.Function):
+    """
+    Functionally the same as the forward method in a TheseusLayer
+    but computes the direct loss minimization in the backward pass.
+    """
+
+    @staticmethod
+    def forward(ctx, objective, optimizer, optimizer_kwargs, input_data, epsilon, *params):
```
luisenp, line 122: nit: Should we rename `params` to something more explicit? Maybe `differentiable_params`, `differentiable_aux_vars`, something else?

theseus/theseus_layer.py
```diff
@@ -93,3 +103,88 @@ def device(self) -> torch.device:
...
+        optim_vars, info = _forward(objective, optimizer, optimizer_kwargs, input_data)
```
luisenp, line 123: nit: someone used to Theseus terminology might see `optim_vars` and think these are `th.Variable` objects. How about `optim_tensors`? Similar comment for usage in `backward`.

luisenp (approved):
This is great! Left some comments/questions.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
rtqichen: The forward and backward functions inside TheseusLayerDLMForward (where this code will be called) automatically have a torch.no_grad(), so there's no need to explicitly change the behavior here.

examples/backward_modes.py
rtqichen: Will do.

theseus/theseus_layer.py
rtqichen: Good idea.

theseus/theseus_layer.py
```diff
@@ -32,15 +38,19 @@ def forward(
...
-            [
-                (var_name, var.data)
-                for var_name, var in self.objective.optim_vars.items()
-            ]
-        )
+        backward_mode = optimizer_kwargs.get("backward_mode", None)
+        DLM_epsilon = optimizer_kwargs.get("DLM_epsilon", 1e-2)
+        if backward_mode == BackwardMode.DLM:
+            # TODO: instantiate self.bwd_objective here.
+            names = set(self.objective.aux_vars.keys()).intersection(input_data.keys())
```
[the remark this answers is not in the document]
rtqichen: Good idea.. added.

theseus/optimizer/nonlinear/tests/test_backwards.py
```diff
@@ -119,3 +119,19 @@ def fit_x(data_x_np):
...
+            "verbose": False,
+            "backward_mode": th.BackwardMode.DLM,
+            "dlm_epsilon": 0.001,
+        },
+    )
+    da_dx_truncated = torch.autograd.grad(
+        updated_inputs["a"], data_x, retain_graph=True
+    )[0].squeeze()
+    print(da_dx_numeric)
+    print(da_dx_truncated)
```
mhmukadam, line 136: Can remove these print statements.

theseus/theseus_layer.py
rtqichen: Using autograd.Function also tends to introduce a bit of overhead and might not be very amenable to the theseus' object-oriented api. For one, it doesn't handle containers for tensors (e.g. dicts) very well.

theseus/theseus_layer.py
rtqichen: I think making this a single cost function just involves concatenating all of the error functions. But yeah, if there's going to an automatic grouping feature, might as well defer to that.

theseus/theseus_layer.py
```diff
@@ -93,3 +118,154 @@ def device(self) -> torch.device:
     @property
     def dtype(self) -> torch.dtype:
         return self.objective.dtype
+
+
+def _forward(objective, optimizer, optimizer_kwargs, input_data):
```
mhmukadam, line 123: Any reason this function should be outside of the `TheseusLayer` class? We could move it inside and rename to `_forward_impl()` to follow what we do in many other places.

theseus/theseus_layer.py
mhmukadam: Any major limitation of this right now, @rtqichen, that we should address? If not we can flag an issue and come back to it after.

rtqichen: None that I'm aware of feature-wise, but due to limitations of autograd.Function not being able to work with dicts, there's currently quite a bit of overhead from the use of python for-loops for converting between different containers for tensors (dicts, lists, subsets of lists). I can come back to this to see if I can speed up / remove some of this overhead later.

theseus/theseus_layer.py
```diff
@@ -93,3 +118,154 @@ def device(self) -> torch.device:
...
+                var.shape[1],
+                aux_vars=[grad_var, epsilon_var],
+                name="dlm_perturbation_" + name,
+            )
+        )
+
+    bwd_optimizer = GaussNewton(
+        bwd_objective,
+        max_iterations=1,
+        step_size=1.0,
```
mhmukadam, line 269: Would we ever need to expose these to be tunable?

rtqichen: Probably not. Would be nice but also don’t want to overwhelm the user with too many hyperparameters. And I think in most cases a single step is sufficient to give a good gradient direction. One thing that could be useful is if there’s ever a custom
Optimizer class for certain types of problems, then it might make sense to use that same optimizer for the backward objective instead of GaussNewton.

theseus/theseus_layer.py
```diff
@@ -93,3 +118,154 @@ def device(self) -> torch.device:
...
+            .to(grad_outputs[0])
+            .reshape(1, 1)
+        }
+        for i, name in enumerate(bwd_objective.optim_vars.keys()):
+            grad_data[name + TheseusLayerDLMForward._grad_suffix] = grad_outputs[i]
+        bwd_data.update(grad_data)
+
+        # Solve backward objective.
+        bwd_objective.update(bwd_data)
+        bwd_optimizer.optimize()
```
mhmukadam, line 215: Is there a chance here that the objective accidentally changes and the optimizer get's called on a stale version. Maybe using `bwd_optimizer.objective` is safer?

mhmukadam (approved):
Looks great! Thanks @rtqichen for adding DLM.

theseus/theseus_layer.py
mhmukadam: Sounds good. Ah yes, custom optimizer is a good thing to keep in mind for later.

theseus/theseus_layer.py
```diff
@@ -93,3 +118,154 @@ def device(self) -> torch.device:
...
+
+
+def _instantiate_dlm_bwd_objective(objective):
+    bwd_objective = objective.copy()
+    epsilon_var = Variable(torch.ones(1, 1), name=TheseusLayerDLMForward._dlm_epsilon)
+    for name, var in bwd_objective.optim_vars.items():
+        grad_var = Variable(
+            torch.zeros_like(var.data), name=name + TheseusLayerDLMForward._grad_suffix
+        )
+        bwd_objective.add(
```
mhmukadam, line 256: Also flagging that we can vectorize this or use Taosha's cost function grouping current in 'experimental'.

mhmukadam (approved):

PR #172, opened 2022-04-21
Add varying batch size support for LUCudaSparseSolver

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/theseus/optimizer/nonlinear/nonlinear_optimizer.py
+++ b/theseus/optimizer/nonlinear/nonlinear_optimizer.py
@@ -17,6 +17,7 @@ import theseus.constants
 from theseus.core import Objective, Variable
 from theseus.optimizer import Linearization, Optimizer, OptimizerInfo
 from theseus.optimizer.linear import LinearSolver
+from theseus.optimizer.linear.lu_cuda_sparse_solver import LUCudaSparseSolver
 
 
 @dataclass
@@ -300,6 +301,12 @@ class NonlinearOptimizer(Optimizer, abc.ABC):
     ) -> OptimizerInfo:
         with torch.no_grad():
             info = self._init_info(track_best_solution, track_err_history, verbose)
+
+        if (
+            type(self.linear_solver) is LUCudaSparseSolver
+            and self.linear_solver.batch_size != self.objective.batch_size
+        ):
+            self.linear_solver.reset(self.objective.batch_size)
 
         if verbose:
             print(
--- a/theseus/optimizer/nonlinear/nonlinear_optimizer.py
+++ b/theseus/optimizer/nonlinear/nonlinear_optimizer.py
@@ -16,8 +16,7 @@ import torch
 import theseus.constants
 from theseus.core import Objective, Variable
 from theseus.optimizer import Linearization, Optimizer, OptimizerInfo
-from theseus.optimizer.linear import LinearSolver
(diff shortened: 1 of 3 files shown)


theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -301,6 +301,12 @@ def _optimize_impl(
         with torch.no_grad():
             info = self._init_info(track_best_solution, track_err_history, verbose)
 
+        if (
```
mhmukadam, line 304: Is the idea to allow changing the batch size after the linear system was first setup? This seems like a nice fix!

I remember @maurimo had mentioned that the LUCudaSparse implementation currently needed to know a fixed batch size at setup time. But if you use the largest size you think you need then you can use any size smaller than that, for example see the [unit test here](https://github.com/facebookresearch/theseus/blob/eb84ce66ec0e550021ddbd0cb21bbd98ca4b09ba/theseus/optimizer/linear/tests/test_lu_cuda_sparse_solver.py#L93).

mhmukadam (approved):

mhmukadam:
Looks good to me. Should we merge this to main if this is a general fix and then merge into experimental?

fantaosha:
> Looks good to me. Should we merge this to main if this is a general fix and then merge into experimental?

This is a quick fix. Maybe we had better not merge it main at the moment.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
maurimo: This is certainly feasible, but is it a good idea? The reset is an expensive operation, because it's redoing the symbolic decomposition (actually, many symbolic decompositions, as many as contexts we have available - we need several contexts to allow unroll, this is a limitation of Cublas we cannot go around), so I would not make it happen automatically. It's an operation to be done explicity if the user really wants it, not because we encountered an irregular batch. In the previous discussions, the consensus was that probably the user prefers discarding a sporadic irregular batch instead of triggering the recomputation of all symbolic decompositions, then again when we go back to the ordinary batch size.

I think the way around this could be a function of the solver informing the user if the batch size is mandatory (eg `is_batch_size_fixed()`), so the user knows that the irregular batches are better discarded, or if not reset has to be called again. Is there any case where this would not be the best solution?

fantaosha: I agree this is not a good idea. In fact, this is a quick fix to the experimental branch. @maurimo Can we use smaller batch size after declaring a large one?

luisenp: I vaguely recalled some discussion about this point, but could not remember what the conclusion was, so I couldn't advise @fantaosha on this. Thanks for clarifying!

I think it's fine to merge this to `experimental`, but definitely not to `main`. We need to think a bit what would be a user friendly way to flag this. One option could be to add a flag somewhere (unsure what's the best place, TheseusLayer maybe) that specifies if dynamic batch size is allowed, and make it default to `False`. Then we throw an error whenever this happens, and mention that this option can be turned on, possibly at a huge cost in running time. 

PS: We should have a better registry of previous discussions. As the library gets bigger it's becoming harder to keep track of all resolutions :)

luisenp (approved):
Approving for merging to current experimental.

PR #177, opened 2022-04-25
Updated SDF object in collision cost functions whenever an aux var is updated

luisenp:
… updated.

## Motivation and Context

Calling `Objective.copy()` when there are `Collision2D` or `EffectorObjectContactPlanar` cost functions resulted in incorrect behavior, because the `sdf_data` variable container created was not being linked to the underlying `PlanarSDF` objects.  

## How Has This Been Tested

I ran the code in #161 and seems to be working (where this bug was first detected).

### DIFF
--- a/theseus/embodied/collision/collision.py
+++ b/theseus/embodied/collision/collision.py
@@ -71,3 +71,8 @@ class Collision2D(CostFunction):
 
     def dim(self) -> int:
         return 1
+
+    # This is needed so that the SDF container also updates with the new aux var
+    def set_aux_var_at(self, index: int, variable: Variable):
+        super().set_aux_var_at(index, variable)
+        self.sdf.update_data(self.sdf_origin, self.sdf_data, self.sdf_cell_size)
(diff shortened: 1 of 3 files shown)


theseus/embodied/collision/collision.py
```diff
@@ -71,3 +71,8 @@ def _copy_impl(self, new_name: Optional[str] = None) -> "Collision2D":
 
     def dim(self) -> int:
         return 1
+
+    # This is needed so that the SDF container also updates with the new aux var
+    def set_aux_var_at(self, index: int, variable: Variable):
```
mhmukadam, line 76: This fix is okay for now, but maybe worth adding an issue to think about a cleaner version later since it might be a tricky thing for a user to remember to implement something like this for their own custom cost. When inheriting from CostFunction, this is not an obvious thing to do.

mhmukadam (approved):

mhmukadam:
Cuda CI is failing, but maybe for some other reason?

theseus/embodied/collision/collision.py
luisenp: I agree, but not sure what the best thing would be to do. Having the user register this kind of "special" objects that also require access to variable containers seems a bit awkward as well?

mhmukadam: Yeah okay, I'll flag an issue. We can think about this more.

PR #182, opened 2022-04-26
Merge DLM to Experimental

luisenp:


theseus/core/objective.py
```diff
@@ -285,7 +285,7 @@ def _add_batched_cost_function(
         else:
             if batch_name not in self.batched_cost_functions:
                 cost_function_batch = cost_function.copy(
-                    batch_name + "__cost_function_batch"
+                    batch_name + "__cost_function_batch", keep_variable_names=True
```
fantaosha, line 288: Is there a reason to keep variable names? Just curious, it looks like there should be no problems with this.

theseus/theseus_layer.py
```diff
@@ -94,3 +120,157 @@ def device(self) -> torch.device:
     @property
     def dtype(self) -> torch.dtype:
         return self.objective.dtype
+
+
+def _forward(objective, optimizer, optimizer_kwargs, input_data):
+    objective.update(input_data)
```
fantaosha, line 126: I'm not sure if we should use `Objective.update()` or `Objective.setup()`. In `experimental`, `Objective.update()` assumes no changes of the batch size.

theseus/theseus_layer.py
```diff
@@ -94,3 +120,157 @@ def device(self) -> torch.device:
...
+            AutoDiffCostFunction(
+                [var],
+                _dlm_perturbation,
+                var.shape[1],
+                aux_vars=[grad_var, epsilon_var],
+                name="dlm_perturbation_" + name,
+                optim_var_names=["orig_var"],
+                aux_var_names=["grad_var", "epsilon_var"],
+            ),
+            use_batches=True,
```
fantaosha, line 268: If you want to use `automatic batching`, it is possible to provide a `batch name` for batching cost functions. FYI, `use_batches` is equivalent to assuming a batch only has one cost functions.

fantaosha (commented):
I don't understand `DLM` quite well. But the code overall looks good to me.

theseus/core/objective.py
luisenp: I was running into errors when populating the batched cost function because `AutoDiffCostFunction` would create attributes named like `"obj_pose_0"`, but the copy would have attribute names like `"obj_pose_0_copy"`.

theseus/theseus_layer.py
luisenp: I still need to run the full sweep, but it seems that this is working fine.

PR #187, opened 2022-05-06
Changes for grouping ablations.

luisenp:


theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -297,6 +302,7 @@ def _optimize_impl(
         track_err_history: bool = False,
         verbose: bool = False,
         backward_mode: BackwardMode = BackwardMode.FULL,
+        grouped_retract: bool = True,
```
fantaosha, line 305: Maybe `batched_retract` since our terminology is `batched_variable` and `batched_cost_function`? Or should we rename `batched_xxx` back to `grouped_xxx`?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -391,12 +400,22 @@ def retract_and_update_variables(
...
 
+        if not grouped_retract:
+            for var in self.linear_solver.linearization.ordering:
+                new_var = var.retract(delta[:, var_idx : var_idx + var.dof()])
+                if force_update:
+                    var.update(new_var.data)
+                else:
+                    var.update(new_var.data, batch_ignore_mask=converged_indices)
+                var_idx += var.dof()
+            return
```
fantaosha, line 416: I don't think it's safe to `return`. Since after update variables, we also need to update `batched_cost_functions` https://github.com/facebookresearch/theseus/blob/c1906a9aaca42e1d0d3d22fdd55337fd66fb3bb8/theseus/optimizer/nonlinear/nonlinear_optimizer.py#L452.

fantaosha (commented):
Left some comments.

fantaosha (approved):

PR #199, opened 2022-05-31
Added aliases for Difference and Between

luisenp:
We can now import `th.Difference`, `th.Local` as aliases for `th.eb.VariableDifference`, and also `th.Between` for `th.eb.Between`.


theseus/__init__.py
```diff
@@ -76,3 +76,8 @@
 from .theseus_layer import TheseusLayer
 
 import theseus.embodied as eb
+
+# Aliases for some standard cost functions
+Difference = eb.VariableDifference
+Local = eb.VariableDifference
+Between = eb.Between
```
mhmukadam, line 83: I am wondering if we should get rid of `eb` altogether? Just import everything from embodied into `th`.

Also should we change `VariableDifference` to `ManifoldDifference`?

mhmukadam (approved):

theseus/__init__.py
luisenp: Not sure about getting rid of `eb`, I kinda like having the separation for things like `th.eb.Collision2D`, `KinematicsModel`, and so on. 

For `VariableDifference`, maybe just `Local` and also keep the `Difference` alias?

luisenp: I meant the other way around. `VariableDifference` -> `Local` (since this is basically the manifold difference, AFAIU), and alias as `Difference` as well.

luisenp: Added.

PR #202, opened 2022-06-03
Added a check in TheseusFunction that enforces copy() also copies the variables

luisenp:
We are using copy to create new variables in the cost function vectorization code, and this would help avoid some unwanted bugs if an user creates a new cost function that re-uses some of the old variables rather than copying.

### DIFF
--- a/theseus/core/tests/test_theseus_function.py
+++ b/theseus/core/tests/test_theseus_function.py
@@ -6,15 +6,15 @@ import numpy as np
 import pytest  # noqa: F401
 import torch
 
-import theseus.core
+import theseus as th
 
 from .common import MockCostFunction, MockCostWeight
 
 
 def test_theseus_function_init():
     all_ids = []
-    variables = [theseus.core.Variable(torch.ones(1, 1), name="var_1")]
-    aux_vars = [theseus.core.Variable(torch.ones(1, 1), name="aux_1")]
+    variables = [th.Variable(torch.ones(1, 1), name="var_1")]
+    aux_vars = [th.Variable(torch.ones(1, 1), name="aux_1")]
     for i in range(100):
         cost_weight = MockCostWeight(torch.ones(1, 1), name=f"cost_weight_{i}")
         if np.random.random() < 0.5:
@@ -31,8 +31,8 @@ def test_theseus_function_init():
 
 
 def test_no_copy_vars_check():
-    variables = [theseus.core.Variable(torch.ones(1, 1), name="var_1")]
-    aux_vars = [theseus.core.Variable(torch.ones(1, 1), name="aux_1")]
+    variables = [th.Variable(torch.ones(1, 1), name="var_1")]
+    aux_vars = [th.Variable(torch.ones(1, 1), name="aux_1")]
     cost_weight = MockCostWeight(torch.ones(1, 1), name="cost_weight")
     cost_function = MockCostFunction(
         variables, aux_vars, cost_weight, no_copy_vars=True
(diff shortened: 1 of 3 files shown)


theseus/core/tests/test_theseus_function.py
```diff
@@ -29,3 +28,14 @@ def test_theseus_function_init():
             assert name == cost_function.name
 
     assert len(set(all_ids)) == len(all_ids)
+
+
+def test_no_copy_vars_check():
+    variables = [theseus.core.Variable(torch.ones(1, 1), name="var_1")]
```
mhmukadam, line 34: This might be a remnant of before, but we should switch to using `th.` like all other tests based on our convention.

mhmukadam, line 34: This might be a remnant of before, but we should switch to using `th.` like all other tests based on our convention.

theseus/core/theseus_function.py
```diff
@@ -80,6 +88,13 @@ def copy(
                 new_var.name = old_var.name
             for old_aux, new_aux in zip(self.aux_vars, new_fn.aux_vars):
                 new_aux.name = old_aux.name
+
+        if self._has_duplicate_vars(new_fn):
+            raise RuntimeError(
```
mhmukadam, line 93: Is the desired behavior to prevent reuse altogether or should we raise a warning instead and provide some other default behavior.

Could the error message be a bit cryptic, if it get's triggered from some leaf cost function?

mhmukadam, line 93: Is the desired behavior to prevent reuse altogether or should we raise a warning instead and provide some other default behavior.

Could the error message be a bit cryptic, if it get's triggered from some leaf cost function?

theseus/core/tests/test_theseus_function.py
```diff
@@ -29,3 +28,14 @@ def test_theseus_function_init():
...
+    aux_vars = [theseus.core.Variable(torch.ones(1, 1), name="aux_1")]
+    cost_weight = MockCostWeight(torch.ones(1, 1), name="cost_weight")
+    cost_function = MockCostFunction(
+        variables, aux_vars, cost_weight, no_copy_vars=True
+    )
+    with pytest.raises(RuntimeError):
+        cost_function.copy()
```
mhmukadam, line 41: If this is the incorrect way to use it which triggers the error, what would be the correct intended way to use it?

mhmukadam, line 41: If this is the incorrect way to use it which triggers the error, what would be the correct intended way to use it?

theseus/core/theseus_function.py
luisenp: I think this should be a full error. We were already implicit asking for this contract to be guaranteed by all subclasses (see [this](https://github.com/facebookresearch/theseus/blob/main/theseus/core/theseus_function.py#L67) comment), but we never enforced it in the code. 

I can make the message a bit more helpful, by saying which class should be fixed.

theseus/core/tests/test_theseus_function.py
luisenp: It's not the usage that would be wrong, it's the way the subclass was implemented that would be wrong. Subclasses must guarantee that variables are also copied, so that the new cost function has new variable objects as well.

theseus/core/theseus_function.py
mhmukadam: Ah okay. Yes the update message sounds good.

luisenp: Added.

theseus/core/tests/test_theseus_function.py
luisenp: Fixed.

theseus/core/theseus_function.py
mhmukadam: Nice reads much clearer!

mhmukadam (approved):

PR #205, opened 2022-06-06
Vectorization refactor

luisenp:
Refactored the vectorization logic. I did some quick tests on motion planning and tactile pose estimation, and seems to be working. 

Some TODOs left:

- [x] Add unit tests for:
  - [x] Shared aux. var computation
  - [x] That schemas are computed correctly
  - [x] Validate vectorized costs values
- [ ] Add vectorized retraction after computing descent direction (separate PR, and we can move `retract_and_update_variables` in that one as well)
- [ ] More thorough testing in the example applications. 
- [x] Add robust costs (separate PR, #146).


theseus/core/vectorizer.py
```diff
@@ -0,0 +1,291 @@
...
+        return f"{obj.__module__}.{obj.__class__.__name__}"
+
+    def _varinfo(var) -> str:
+        return f"{_fullname(var)}{tuple(var.shape[1:])}"
+
+    return (
+        (_fullname(cost_function),)
+        + tuple(_varinfo(v) for v in cost_function.optim_vars)
+        + tuple(_varinfo(v) for v in cost_function.aux_vars)
+        + (_fullname(cost_function.weight),)
```
fantaosha, line 24: Thought no strong preferences, I think it is better to keep it possible to vectorize cost functions of different types of weights. In most cases, `CostFunction.errors()` and `CostFunction.jacobians()` are the most time-consuming.

theseus/core/vectorizer.py
```diff
@@ -0,0 +1,291 @@
...
+        return self.cost_fn.dim()
+
+    def _copy_impl(self, new_name: Optional[str] = None) -> "_CostFunctionWrapper":
+        return _CostFunctionWrapper(self.cost_fn, name=new_name)
+
+    def __str__(self) -> str:
+        return f"Wrapper({str(_get_cost_function_schema(self.cost_fn))})"
+
+    def __repr__(self) -> str:
+        return str(self)
```
fantaosha, line 72: Is it guaranteed that `self._cached_error=None` and `self._cached_jacobians=None` whenever either `optim_vars` or `aux_vars` are updated?

fantaosha (commented):
Overall LGTM. The only thing I'm concerned with is to make sure `_cached_errs` and `cached_jacobians` in `CostFunctionWrapper` are cleared if the vars in corresponding cost functions are updated.

theseus/core/vectorizer.py
luisenp: I agree this is desirable to have, for the reason you mention. That being said, it seems OK to me to open an issue for this and revisit later, once we have more feedback from users, because of two reasons:

1. separating into error/jacobian and weight makes the code more complicated, 
2. more important, it seems to me that it will be a rare use case to have a bunch of cost functions of the same type but with different cost weight types, so we might be over engineering for an atypical use case.

theseus/core/vectorizer.py
luisenp: That's a great observation. In fact, this is actually only guaranteed if `Objective.update()` is used, but not if variables are updated via their stand alone `update` methods. This is why [this](https://github.com/facebookresearch/theseus/pull/205/files#diff-73a97268e17aeaa7f1e018f6f77e917666909b415d6d9567bea88977e53a576bR397) line is needed after the inner loop var update. 

Maybe the safest way to prevent unwanted behavior is (by default) enable the vectorized caches only when optimization is called via `TheseusLayer`?

fantaosha: I agree. As long as users have to no access to `CostFunctionWrapper` and developers are aware of this, it should be safe.

theseus/theseus_layer.py
```diff
@@ -9,22 +9,20 @@
...
 from theseus.optimizer.linear import LinearSolver
 from theseus.optimizer.nonlinear import BackwardMode, GaussNewton
 
 
 class TheseusLayer(nn.Module):
-    def __init__(
-        self,
-        optimizer: Optimizer,
-    ):
+    def __init__(self, optimizer: Optimizer, vectorize_cost_fns: bool = True):
```
mhmukadam, line 19: When we vectorize other things like cost weights or retraction (anything else?) would the interface need a separate bool for all of them or should we have a single `vectorize` for everything?

theseus/core/vectorizer.py
```diff
@@ -0,0 +1,301 @@
+from collections import defaultdict
```
mhmukadam, line 1: Need to add license block for all new .py files.

theseus/theseus_layer.py
```diff
@@ -9,22 +9,20 @@
...
         super().__init__()
         self.objective = optimizer.objective
         self.optimizer = optimizer
         self._objectives_version = optimizer.objective.current_version
+        if vectorize_cost_fns:
+            Vectorize(self.objective)
```
mhmukadam, line 25: Does the vectorized objective object not need to be saved?

mhmukadam: I see it's directly editing the underlying objective.

theseus/optimizer/linearization.py
```diff
@@ -51,6 +51,7 @@ def linearize(self):
             raise RuntimeError(
                 "Attempted to linearize an objective with an incomplete variable order."
             )
+        self.objective.update_vectorization()
```
mhmukadam, line 54: Is the overhead for doing this roughly the same as what we had in the experimental branch?

theseus/optimizer/sparse_linearization.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         cost_function_row_block_starts = []  # where data start for this row block
         cost_function_stride = []  # total jacobian cols
 
-        for _, cost_function in enumerate(self.objective):
+        for _, cost_function in enumerate(self.objective._vectorized_iter()):
```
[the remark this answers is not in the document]
mhmukadam: I am wondering if it is clearer to have something like objective.get_iterator(), which returns the regular one or the vectorized one based on whether the objective is vectorize or not.

theseus/theseus_layer.py
luisenp: I prefer the single `vectorize` for everything, seems cleaner. Any thoughts?

theseus/optimizer/sparse_linearization.py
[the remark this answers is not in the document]
luisenp: What you are describing is the current behavior, but I like your name better. But do note that I deliberately named this as private, since I think I'm not sure it's safe for the user to have access to this iterator. If they update variables and try to loop over vectorized cost functions, they will get stale data. This is only safe to use inside linearization classes.

theseus/optimizer/linearization.py
luisenp: I expect the overhead to be similar, but we need to test it to be sure. Not sure if @fantaosha did some additional optimizations that I forgot to include. 

This does work with the linear optimizers.

theseus/core/tests/test_objective.py
```diff
@@ -151,24 +151,21 @@ def test_add_and_erase_step_by_step():
...
-            with pytest.warns(RuntimeWarning):
-                # a warning should emit the first time cw1/cw2 are added
-                objective.add(cost_function)
-        else:
-            objective.add(cost_function)
+        objective.add(cost_function)
 
-    for name in ["var1", "ignored_optim_var"]:
-        assert name in objective.cost_weight_optim_vars
+    # for name in ["var1", "ignored_optim_var"]:
```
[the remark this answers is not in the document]
luisenp: I left these here in case someone (very) strongly opposed removed optim_vars from cost weights, and I had to roll this back. I can remove now.

theseus/core/cost_function.py
```diff
@@ -17,6 +17,19 @@
 from .variable import Variable
 
 
+def _register_vars_in_list(
+    cost_fn: TheseusFunction, var_list_: Iterable[Variable], is_optim: bool = False
```
mhmukadam, line 21: What's our convention for private objects: trailing underscore or leading underscore? We seem to have a mix at the moment so maybe we can just pick one (say leading?).

luisenp: Convention for private is leading underscore. I only use trailing when I have nested function (which this used to be before this PR), to avoid silent scoping errors. I missed the underscore on this, I'll fix. 

If there are other places where I've used trailing to indicate private, that's not intended, and we should fix.

theseus/core/cost_function.py
```diff
@@ -17,6 +17,19 @@
 from .variable import Variable
 
 
+def _register_vars_in_list(
```
mhmukadam, line 20: Trying to understand the underlying reason for the change of this from a class function to outside the class.

luisenp: We also use it in the `Vectorize` class. But now that I think about it we can probably make it a private function of base `TheseusFunction` class.

theseus/core/objective.py
```diff
@@ -471,9 +476,21 @@ def _get_batch_size(batch_sizes: Sequence[int]) -> int:
...
+
     # iterates over cost functions
     def __iter__(self):
-        return iter([f for f in self.cost_functions.values()])
+        return iter([cf for cf in self.cost_functions.values()])
+
+    def _vectorized_iter(self):
+        if self._cost_functions_iterable is None:
+            return iter([cf for cf in self.cost_functions.values()])
+        return iter([cf for cf in self._cost_functions_iterable])
```
mhmukadam, line 493: Could iter and _vectorized_iter be combined? (related to another comment)

theseus/core/objective.py
```diff
@@ -471,9 +476,21 @@ def _get_batch_size(batch_sizes: Sequence[int]) -> int:
         batch_sizes.extend([v.data.shape[0] for v in self.aux_vars.values()])
         self._batch_size = _get_batch_size(batch_sizes)
 
+    def update_vectorization(self):
+        if self._vectorization_run is None:
+            return
+        if self._batch_size is None:
+            self.update()
+        self._vectorization_run()
+
```
mhmukadam, line 485: Nit to add explicit `return None`
```
if self._vectorization_run is not None:
        if self._batch_size is None:
            self.update()
        self._vectorization_run()
return None
```

luisenp: Your code looks cleaner, but I'm a bit confused by the explicit `return None`. Why do we need that line?

theseus/core/vectorizer.py
mhmukadam: Makes sense. Let's open an issue for this to tackle later.

theseus/core/vectorizer.py
```diff
@@ -0,0 +1,301 @@
...
+    def __str__(self) -> str:
+        return f"Wrapper({str(_get_cost_function_schema(self.cost_fn))})"
+
+    def __repr__(self) -> str:
+        return str(self)
+
+
+# This class replaces the Objective's iterator for one that takes advantage of
+# cost function vectorization
+# TODO:
```
mhmukadam, line 77: Are theses still TODOs? We should add issues if we will address them in future PRs since it is easy to loose track of todos in code comments.

theseus/core/vectorizer.py
```diff
@@ -0,0 +1,301 @@
...
+
+        # Dict[_CostFunctionSchema, List[str]]
+        self._var_names = self._get_var_names()
+
+        # `vectorize()` will compute an error vector for each schema, then populate
+        # the wrappers with their appropriate weighted error slice.
+        # Replacing `obj._cost_functions_iterable` allows to recover these when
+        # iterating the Objective.
+        objective._cost_functions_iterable = self._cost_fn_wrappers
+        objective._vectorization_run = self._vectorize
```
mhmukadam, line 119: self._vectorize doesn't yet seem to be defined in the init function or anywhere else in this file.

theseus/core/vectorizer.py
```diff
@@ -0,0 +1,301 @@
...
+                    return f"{Vectorize._SHARED_TOKEN}{next(iter(name_set_))}"
+                return f"var_{var_idx}"
+
+            info[schema] = [
+                _get_name(i, s) for i, s in enumerate(var_names_per_position)
+            ]
+        return info
+
+    @staticmethod
+    def _get_all_vars(cf) -> List[Variable]:
```
mhmukadam, line 160: `cf `-> `cf: CostFunction`

theseus/core/tests/test_vectorizer.py
```diff
@@ -0,0 +1,189 @@
...
+                assert adummy in aux_vars
+                assert _check_attr(cf, v2) and _check_attr(cf, odummy)
+                assert w_err.allclose((v2.data - t1.data) * w2)
+                saw_cf2 = True
+            else:
+                assert False
+        assert saw_cf1 and saw_cf2
+
+
+def test_correct_schemas_and_shared_vars():
```
mhmukadam, line 69: The other two tests are checking with batch size. Do we want to have this one with some batch size as well?

mhmukadam (commented):
Nice work on the refactor @luisenp and of course nice work on the original features @fantaosha! Reviewed everything so far and left comments.

theseus/core/vectorizer.py
mhmukadam: > enable the vectorized caches only when optimization is called via `TheseusLayer`
Is this already incorporated?

theseus/theseus_layer.py
mhmukadam: Sounds fine to me, @fantaosha?

theseus/tests/test_theseus_layer.py
```diff
@@ -131,7 +131,7 @@ def error_fn(optim_vars, aux_vars):
         max_iterations=max_iterations,
     )
     assert isinstance(optimizer.linear_solver, linear_solver_cls)
-    theseus_layer = th.TheseusLayer(optimizer)
+    theseus_layer = th.TheseusLayer(optimizer, vectorize_cost_fns=False)
```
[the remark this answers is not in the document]
mhmukadam: Ah okay. In that case maybe we can just leave the default vectorize True for all unit tests unless there is a special case where we shouldn't.

theseus/optimizer/sparse_linearization.py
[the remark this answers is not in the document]
mhmukadam: That makes sense. Let's go with the more general name in that case for the private iterator.

I also realized there may be cases where the user still wants to access the default iterator through __iter__ even if they have vectorized, so we should still have a separate function like above.

theseus/optimizer/linearization.py
mhmukadam: Sounds good. We'll know more when we rerun profiling.

Any easy way to make it more general so it works at the optimizer level? Besides linear optimizers we'll likely continue to add other alternatives like GBP that Joe is working on right now.

theseus/core/cost_function.py
mhmukadam: Interesting scenario with the nested function. Okay good to know. We'll add in the documentation for conventions. In the final sweep we can find if there are any other trailing underscores.

theseus/theseus_layer.py
fantaosha: The vectorization of the retraction map should be the default choice and always `True`. I'm not sure if the vectorization of weights and robust losses would raise any problems. In particular, many cost functions might share weights and robust losses, and this might be a problem for `AutoDiffCostFunction` using `functorch`.

theseus/core/cost_function.py
mhmukadam: Another option for such commonly used things might be to put them in utils.

theseus/theseus_layer.py
fantaosha: In addition, I didn't see a performance gain for the vectorization of weights and robust losses in the experimental branch.

theseus/core/objective.py
mhmukadam: The explicit return None seems like a [good practice thing](https://realpython.com/python-return-statement/#using-the-python-return-statement-best-practices) and somewhat a convention. I can't remeber if there are other place in our code where we have functions not explicitly returning anything (aside from the mock ones that use pass).

theseus/core/tests/test_vectorizer.py
luisenp: This only test functionality that happens at construction, so it doesn't depend on the batch size.

theseus/core/cost_function.py
luisenp: I replaced by a public `TheseusFunction.register_vars()`.

theseus/core/vectorizer.py
luisenp: Created #208

theseus/core/vectorizer.py
luisenp: This is already incorporated via the private iterator. That's the only way to use the vectorization results.

theseus/theseus_layer.py
luisenp: Renamed as `vectorize: bool = True`.

theseus/core/vectorizer.py
luisenp: Added.

theseus/optimizer/linearization.py
luisenp: Our current requirements is to call `Objective.update_vectorization()` before any work that wants to use the vectorized cost functions. We may have a better idea if there is some better place than linearization to automatically do these for other optimization classes when we have one or more implemented.

theseus/core/cost_function.py
```diff
@@ -128,8 +115,8 @@ def __init__(
             raise ValueError(
                 "AutodiffCostFunction must receive at least one optimization variable."
             )
-        _register_vars_in_list(self, optim_vars, is_optim=True)
-        _register_vars_in_list(self, aux_vars, is_optim=False)
+        self.register_vars(optim_vars, optim_vars=True)
```
mhmukadam, line 118: `is_optim_vars`? The current name could lead to some error/confusion.

rtqichen (approved):

theseus/optimizer/linearization.py
luisenp: We also need to do it after every iteration in the internal loop, so that alone wouldn't be sufficient.

theseus/optimizer/sparse_linearization.py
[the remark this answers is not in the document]
luisenp: Renamed to `_get_iterator()`

theseus/tests/test_theseus_layer.py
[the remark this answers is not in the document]
luisenp: It's set to `True` now.

theseus/core/cost_function.py
luisenp: Changed.

theseus/optimizer/linearization.py
mhmukadam: Ah right, okay let's figure it out in future PRs.

theseus/core/theseus_function.py
```diff
@@ -64,6 +64,16 @@ def register_aux_vars(self, aux_var_names: Sequence[str]):
         for name in aux_var_names:
             self.register_aux_var(name)
 
+    def register_vars(self, vars: Iterable[Variable], is_optim_vars: bool = False):
+        for var_ in vars:
```
mhmukadam, line 68: Nit: trailing underscore.

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -80,7 +80,7 @@ def ee_pose_err_fn(optim_vars, aux_vars):
         max_iterations=15,
         step_size=0.5,
     )
-    theseus_optim = th.TheseusLayer(optimizer)
+    theseus_optim = th.TheseusLayer(optimizer, vectorize=False)
```
mhmukadam, line 83: True?

theseus/optimizer/nonlinear/tests/test_backwards.py
```diff
@@ -59,7 +59,7 @@ def quad_error_fn(optim_vars, aux_vars):
     "x": data_x,
     "y": data_y,
 }
-theseus_optim = th.TheseusLayer(optimizer)
+theseus_optim = th.TheseusLayer(optimizer, vectorize=False)
```
mhmukadam, line 62: same, True?

mhmukadam (approved):

bamos (approved):

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
luisenp: Oh, forgot about these two. Added.

theseus/core/theseus_function.py
luisenp: Good catch. Changed.

PR #212, opened 2022-06-10
Added option for auto resetting LUCudaSparseSolver if the batch size needs to change

luisenp:


theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -91,6 +95,9 @@ def solve(
         damping_eps: float = 1e-8,
         **kwargs,
     ) -> torch.Tensor:
+        if self._auto_reset:
+            if self._solver_contexts[0].batch_size != self._objective.batch_size:
```
mhmukadam, line 99: IIRC smaller batch size is okay so we could save some reset triggers by having this be a `<` check instead of `!=`?

mhmukadam (approved):

theseus/optimizer/linear/lu_cuda_sparse_solver.py
fantaosha: No, `<` yields an error and `LUCuda` strictly requires `=`.

fantaosha (commented):
LGTM.

PR #213, opened 2022-06-10
Address comments on RobustCostFunction implementation

luisenp:
This addresses some comments made in #148 after merging.


theseus/core/loss.py
```diff
@@ -7,8 +7,10 @@
 
 import torch
 
+_LOSS_EPS = 1e-20
```
luisenp, line 10: @mhmukadam Some comments:

- I'm OK with changing the term `Loss` for something else, but not sure what's a good alternative. Maybe `RobustLoss`? Is there another standard term for these?
- What would be good unit tests for these components? 
- I actually think that `error` and `jacobians` (unweighted) are operations that don't really make sense for this class, and we should just throw errors. Maybe we should just think of what a clear error message for these would be.  At least this is true in the way we had implemented these, which is weight first, robustify later.

theseus/core/loss.py
```diff
@@ -23,20 +25,20 @@ def linearize(x: torch.Tensor, raidus: torch.Tensor) -> torch.Tensor:
 class WelschLoss(Loss):
     @staticmethod
     def evaluate(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
-        return radius - radius * torch.exp(-x / radius)
+        return radius - radius * torch.exp(-x / (radius + _LOSS_EPS))
```
luisenp, line 28: I added some EPS for potentially problematic operations, @fantaosha can double check if these make sense. We should later consolidate these as part of #70.

theseus/core/loss.py
```diff
@@ -23,20 +25,20 @@ def linearize(x: torch.Tensor, raidus: torch.Tensor) -> torch.Tensor:
...
-        return torch.exp(-x / radius)
+        return torch.exp(-x / (radius + _LOSS_EPS))
 
 
 class HuberLoss(Loss):
     @staticmethod
     def evaluate(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
         return torch.where(
-            x > radius, 2 * torch.sqrt(radius * x.max(radius)) - radius, x
+            x > radius, 2 * torch.sqrt(radius * x.max(radius) + _LOSS_EPS) - radius, x
```
luisenp, line 39: I think it's fine to just rely on torch for checking if the shapes are good or not. There is no requirement about dimensions, since the arguments are just torch tensors.

luisenp (commented):
@mhmukadam Added responses to the original comments on #148 for quick reference.

theseus/core/loss.py
mhmukadam: - Yes, `RobustLoss` sounds good.
- We could test the robust loss and robust cost function on some some rand values over say a th.Difference (potentially using a Lie group) and verify the Jacobians against finite difference.
- Hmm, in that case we can raise warning and say robust cost requires weighting first and return the weighted and robustified version.

theseus/core/loss.py
fantaosha: I would like to replace `radius` with `log_radius` since there are no guarantees that `radius` is nonnegative. Instead, `radius = torch.exp(log_radius)` always yields a positive one.

theseus/core/robust_cost_function.py
```diff
@@ -84,7 +87,9 @@ def weighted_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
         ) = self.cost_function.weighted_jacobians_error()
         squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
         loss_radius = torch.exp(self.log_loss_radius.data)
```
fantaosha, line 89: I would like to move this line `log_radius` into `loss.py`.

fantaosha (commented):
LGTM except for some minor comments about where to put `log_radius` and `raidus`.

theseus/core/tests/test_robust_cost.py
```diff
@@ -1,7 +1,6 @@
 import torch
```
mhmukadam, line 1: Missing license header.

theseus/core/robust_cost_function.py
```diff
@@ -77,7 +95,11 @@ def weighted_error(self) -> torch.Tensor:
         )
 
     def jacobians(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
-        raise NotImplementedError
+        warnings.warn(
+            "Computing the robust cost error requires weighting first, so "
+            "jacobians() is equivalent to weighted_jacobians()."
```
mhmukadam, line 100: weighted_jacobians_error()

mhmukadam (approved):
Looks great!

theseus/core/loss.py
```diff
@@ -7,36 +7,46 @@
 
 import torch
 
+_LOSS_EPS = 1e-20
+
+
+class RobustLoss(abc.ABC):
```
mhmukadam, line 13: Oh we should update the file name from loss to robust_loss

PR #214, opened 2022-06-10
Moved the method that retracts all variables with a given delta to Objective

luisenp:
We have talked for quite a while about where to place this method. Proposing the solution in this PR. Planning to use this as the starting point for the vectorized retraction maps.

### DIFF
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -504,7 +504,7 @@ class Objective:
         if self._vectorization_to is not None:
             self._vectorization_to(*args, **kwargs)
 
-    def step_optim_vars(
+    def retract_optim_vars(
         self,
         delta: torch.Tensor,
         ordering: Iterable[Manifold],
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -503,3 +503,20 @@ class Objective:
         self.dtype = dtype or self.dtype
         if self._vectorization_to is not None:
             self._vectorization_to(*args, **kwargs)
+
+    def step_optim_vars(
+        self,
+        delta: torch.Tensor,
+        ordering: Iterable[Manifold],
+        ignore_mask: Optional[torch.Tensor] = None,
+        force_update: bool = False,
+    ):
+        var_idx = 0
+        for var in ordering:
+            new_var = var.retract(delta[:, var_idx : var_idx + var.dof()])
+            if ignore_mask is None or force_update:
+                var.update(new_var.data)
+            else:
+                var.update(new_var.data, batch_ignore_mask=ignore_mask)
+            var_idx += var.dof()
+        self.update_vectorization()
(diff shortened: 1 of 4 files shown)


theseus/core/objective.py
```diff
@@ -503,3 +503,20 @@ def to(self, *args, **kwargs):
         self.dtype = dtype or self.dtype
         if self._vectorization_to is not None:
             self._vectorization_to(*args, **kwargs)
+
+    def step_optim_vars(
```
mhmukadam, line 507: Nice, Objective class actually seems like a good place for this function.

`step_...` makes it sound like optimizer related, how about `update_...`? Maybe `update_optim_vars_with_delta` if we want to be really descriptive.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -262,8 +262,11 @@ def _optimize_loop(
                 step_size = self.params.step_size
                 force_update = False
 
-            self.retract_and_update_variables(
-                delta, converged_indices, step_size, force_update=force_update
+            self.objective.step_optim_vars(
+                delta * step_size,
```
mhmukadam, line 266: Maybe worth adding an explicit line that does `delta *= step_size` before this function call?

theseus/theseus_layer.py
```diff
@@ -218,8 +218,8 @@ def backward(ctx, *grad_outputs):
         with torch.no_grad():
             bwd_optimizer.linear_solver.linearization.linearize()
             delta = bwd_optimizer.linear_solver.solve()
-            bwd_optimizer.retract_and_update_variables(
-                delta, None, 1.0, force_update=True
+            bwd_optimizer.objective.step_optim_vars(
+                delta, bwd_optimizer.linear_solver.linearization.ordering
```
mhmukadam, line 222: The previous version was doing `force_update=True` here?

mhmukadam (approved):

theseus/core/objective.py
luisenp: Maybe `retract_optim_vars`?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Is this necessary? I don't see delta being used anywhere else after this function call.

theseus/core/objective.py
mhmukadam: Yeah `retract_optim_vars` is good.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
mhmukadam: Probably okay, just wondering if there is a chance of introducing bugs in the future by having a variable change in a function call.

rtqichen (approved):

theseus/core/objective.py
luisenp: Changed.

mhmukadam (approved):

PR #215, opened 2022-06-13
Vectorize optimization variables retraction step 

luisenp:
Still need to add unit tests.

### DIFF
--- a/theseus/core/tests/test_vectorizer.py
+++ b/theseus/core/tests/test_vectorizer.py
@@ -204,7 +204,9 @@ def test_vectorized_retract():
         batch_size = rng.choice(range(1, 11))
         n_vars = rng.choice([1, 10, 100])
         for _ in range(n_vars):
-            var_type: th.LieGroup = rng.choice([th.Vector, th.SE2, th.SE3])
+            var_type: th.LieGroup = rng.choice(
+                [th.Vector, th.SE2, th.SE3, th.SO2, th.SO3]
+            )
             if var_type == th.Vector:
                 dof = rng.integers(1, 10)
                 var = th.Vector.rand(batch_size, dof, generator=generator)


mhmukadam (commented):
Nice, this looks good to me. After unit tests and merging #214 this can be merged.

theseus/core/vectorizer.py
```diff
@@ -306,6 +309,66 @@ def _vectorize(self):
...
+            # Retract the vectorized data, then redistribute to the variables
+            vectorized_var.update(vectorized_data)
+            new_var = vectorized_var.retract(vectorized_delta)
+            start_idx = 0
+            for var in var_list:
+                retracted_data_slice = new_var[start_idx : start_idx + batch_size, :]
+                if ignore_mask is None or force_update:
+                    var.update(retracted_data_slice)
+                else:
+                    var.update(retracted_data_slice, batch_ignore_mask=ignore_mask)
```
fantaosha, line 369: How this update affects the vectorized cost function?

fantaosha (commented):
LGTM. Only concern with how the update affects the vectorized cost function.

theseus/core/vectorizer.py
luisenp: Good question. I think this should work, since `ignore_mask` applies to all variables, and it indexes over the batch. So we can first compute everything, then apply the (same) mask for each variable. But I'll try to add a unit tests for this as well, just in case.

fantaosha (approved):

theseus/core/vectorizer.py
fantaosha: Once the unit test is passed. I think we can merge this.

mhmukadam (approved):
Looks great!

PR #218, opened 2022-06-15
Moved Vectorize(objective) to the Optimizer class.

luisenp:
Note that the old stand-alone optimizer tests have `vectorize=False` because the copy method of the CostFunction used is not implemented. It's kind of a pain to implement it for not a lot of added value, since the end-to-end and backward tests do use vectorization.

### DIFF
--- a/theseus/optimizer/nonlinear/tests/common.py
+++ b/theseus/optimizer/nonlinear/tests/common.py
@@ -19,23 +19,47 @@ class ResidualCostFunction(th.CostFunction):
         point=None,
         multivar=False,
         noise_mag=0,
+        target=None,
     ):
         super().__init__(cost_weight, name=name)
         len_vars = len(optim_vars) if multivar else optim_vars[0].dof()
-        assert true_coeffs.ndim == 1 and true_coeffs.numel() == len_vars
-        if point.ndim == 1:
-            point = point.unsqueeze(0)
-        assert point.ndim == 2 and point.shape[1] == len_vars - 1
+        self._optim_vars = optim_vars
+
+        if isinstance(true_coeffs, torch.Tensor):
+            assert true_coeffs.ndim == 1 and true_coeffs.numel() == len_vars
+            self.true_coeffs = th.Variable(data=true_coeffs.unsqueeze(0))
+        else:
+            self.true_coeffs = true_coeffs
+
         batch_size = point.shape[0]
+        if isinstance(point, torch.Tensor):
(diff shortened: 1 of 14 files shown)


theseus/theseus_layer.py
```diff
@@ -264,6 +261,7 @@ def _instantiate_dlm_bwd_objective(objective):
 
     bwd_optimizer = GaussNewton(
         bwd_objective,
+        vectorize=False,
```
mhmukadam, line 264: We should have vectorize on for DLM as well?

theseus/optimizer/nonlinear/tests/common.py
```diff
@@ -110,7 +110,7 @@ def _check_nonlinear_least_squares_fit(
     objective.update(values)
     initial_error = objective.error_squared_norm() / 2
     max_iterations = 20
-    optimizer = nonlinear_optim_cls(objective)
+    optimizer = nonlinear_optim_cls(objective, vectorize=False)
```
mhmukadam, line 113: > old stand-alone optimizer tests have `vectorize=False` because the copy method of the CostFunction used is not implemented

This sounds fine. Maybe leave a comment in the code for this.

mhmukadam (approved):

theseus/optimizer/optimizer.py
```diff
@@ -22,8 +22,10 @@ class OptimizerInfo:
 
 
 class Optimizer(abc.ABC):
-    def __init__(self, objective: Objective, *args, **kwargs):
+    def __init__(self, objective: Objective, *args, vectorize: bool = False, **kwargs):
```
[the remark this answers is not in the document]
luisenp: Oh sorry, this was not intentional.

theseus/theseus_layer.py
luisenp: I changed it to `True` and it passes the backward tests, I'll be more comfortable with this part after we re-run the backward experiments, but I'll leave as is for now.

theseus/optimizer/optimizer.py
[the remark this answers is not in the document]
luisenp: Fixed.

luisenp:
Squeezing in a quick fix for a flaky unit tests in RobustCostFunction.

fantaosha (approved):
LGTM overall.

fantaosha (approved):
LGTM.

PR #221, opened 2022-06-17
Added on-demand vectorization and also vectorized Objective.error().

luisenp:


theseus/core/variable.py
```diff
@@ -60,6 +61,7 @@ def update(
             self.data = torch.where(batch_ignore_mask.view(mask_shape), self.data, data)
         else:
             self.data = data
+        self._num_updates += 1
```
mhmukadam, line 64: Any way this manual counter could break? In case you call copy() or to().

mhmukadam (approved):

theseus/core/variable.py
luisenp: One failure could be users trying to do `variable.data = some_tensor`, which would change the data but leave this unchanged, but this is already heavily discouraged anyway. 

Also note that the actual value of this counter is not too important, all that matters is to detect that something has changed with respect to version used in the last vectorization.

mhmukadam: Sounds good. Maybe for later: I am wondering if it is possible to disable `variable.data = some_tensor` and print an error saying use update() instead.

PR #222, opened 2022-06-20
Destroy cusolver context in CusolverLUSolver destructor

luisenp:


theseus/extlib/cusolver_lu_solver.cpp
```diff
@@ -26,6 +26,8 @@ enum Ordering {
 
 struct CusolverLUSolver {
 
+	~CusolverLUSolver();
```
mhmukadam, line 29: Maybe add a small unit test if necessary?

mhmukadam (approved):

theseus/extlib/cusolver_lu_solver.cpp
maurimo: Sorry for the late answer... yeah I don't know of an easy way to test this. I would make sure the code does the right thing (destroying all contexts that might have been created) and then rely of RAII logic of C++

Issue #61, opened 2022-01-31
Add float32 tests


fantaosha:
For SO3 class.

luisenp:
Closed with #227

PR #224, opened 2022-06-21
Added close-formed jacobian for DLM perturbation cost.

luisenp:
Tested with 64-bit precision. 32-bit precision fails. We should address with #61.

### DIFF
--- a/theseus/tests/test_dlm_perturbation.py
+++ b/theseus/tests/test_dlm_perturbation.py
@@ -45,3 +45,41 @@ def test_dlm_perturbation_jacobian():
             error = cf.error()
             assert torch.allclose(error_jac, error)
             assert torch.allclose(jacobians[0], expected_jacs[0], atol=1e-5)
+
+
+def test_backward_pass_se3_runs():
+    generator = torch.Generator()
+    generator.manual_seed(0)
+    dtype = torch.float64
+    batch_size = 10
+    var = th.rand_se3(batch_size, generator=generator)
+    var.name = "v1"
+    target = th.rand_se3(batch_size, generator=generator)
+    target.name = "target"
+
+    objective = th.Objective()
+    objective.add(th.Difference(var, th.ScaleCostWeight(1.0), target))
+    objective.to(dtype=dtype)
+    optimizer = th.GaussNewton(objective)
+    layer = th.TheseusLayer(optimizer)
+
+    target_data = torch.nn.Parameter(th.rand_se3(batch_size, dtype=dtype).data)
+    adam = torch.optim.Adam([target_data], lr=0.01)
+    loss0 = None
+    for _ in range(5):
+        adam.zero_grad()
+        with th.enable_lie_tangent():
+            out, _ = layer.forward(
+                {"target": target_data},
+                optimizer_kwargs={
+                    "backward_mode": th.BackwardMode.DLM,
+                    "verbose": False,
+                },
+            )
+
+            loss = out["v1"].norm()
+            if loss0 is None:
+                loss0 = loss.item()
+            loss.backward()
+            adam.step()
+    assert loss.item() < loss0
--- a/theseus/tests/test_dlm_perturbation.py
+++ b/theseus/tests/test_dlm_perturbation.py
@@ -1,0 +1,47 @@
+import numpy as np
+import torch
+
+import theseus as th
+from theseus.theseus_layer import _DLMPerturbation
+from theseus.utils import numeric_jacobian
+
+
+def test_dlm_perturbation_jacobian():
+    generator = torch.Generator()
+    generator.manual_seed(0)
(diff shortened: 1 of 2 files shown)


theseus/theseus_layer.py
```diff
@@ -237,27 +239,65 @@ def backward(ctx, *grad_outputs):
...
 
 def _instantiate_dlm_bwd_objective(objective):
     bwd_objective = objective.copy()
-    epsilon_var = Variable(torch.ones(1, 1), name=TheseusLayerDLMForward._dlm_epsilon)
+    epsilon_var = Variable(
+        torch.ones(1, 1, dtype=bwd_objective.dtype),
+        name=TheseusLayerDLMForward._dlm_epsilon,
+    )
+    unit_weight = ScaleCostWeight(1.0)
+    unit_weight.to(dtype=objective.dtype)
```
mhmukadam, line 293: Do we need to set device for epsilon and unit weight?

theseus/tests/test_dlm_perturbation.py
```diff
@@ -0,0 +1,85 @@
...
+                grad = group_cls.randn(batch_size, dtype=dtype, generator=generator)
+
+            w = th.ScaleCostWeight(1.0).to(dtype=dtype)
+            cf = _DLMPerturbation(var, epsilon, grad, w)
+
+            def new_error_fn(vars):
+                new_cost_function = _DLMPerturbation(vars[0], epsilon, grad, w)
+                return th.Vector(data=new_cost_function.error())
+
+            expected_jacs = numeric_jacobian(
```
mhmukadam, line 38: This is great! If we want to make sure results are same we could either check against autodiffcost version or just rerun the backward experiments to verify.

theseus/theseus_layer.py
luisenp: Yes, good catch!

theseus/theseus_layer.py
```diff
@@ -242,12 +249,17 @@ def backward(ctx, *grad_outputs):
...
     def __init__(
         self,
-        var: LieGroup,
+        var: Manifold,
         epsilon: Variable,
         grad: Variable,
         cost_weight: CostWeight,
         name: Optional[str] = None,
     ):
+        if not isinstance(var, LieGroup):
```
mhmukadam, line 258: LieGroup -> Manifold? following the function signature.

mhmukadam (approved):

theseus/theseus_layer.py
luisenp: No, we need this to be a `LieGroup` for the projection operator to be defined. Manifold is more general than needed, but avoids some annoying mypy errors when calling this function a level above.

theseus/tests/test_dlm_perturbation.py
```diff
@@ -0,0 +1,104 @@
...
+def test_dlm_perturbation_jacobian():
+    generator = torch.Generator()
+    generator.manual_seed(0)
+    rng = np.random.default_rng(0)
+    dtype = torch.float64
+    for _ in range(100):
+        group_cls = rng.choice([th.Vector, th.SE3, th.SE2, th.SO2, th.SO3])
+        for batch_size in [1, 10, 100]:
+            epsilon = th.Variable(
+                data=torch.randn(batch_size, 1, dtype=dtype, generator=generator)
```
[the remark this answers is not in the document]
luisenp: How about adding a check after [this](https://github.com/facebookresearch/theseus/blob/main/theseus/theseus_layer.py#L41) line to enforce that `dlm_epsilon` is a float?

theseus/theseus_layer.py
```diff
@@ -237,27 +246,70 @@ def backward(ctx, *grad_outputs):
...
+        cost_weight: CostWeight,
+        name: Optional[str] = None,
+    ):
+        if not isinstance(var, LieGroup):
+            raise ValueError(
+                f"DLM requires LieGroup-type variables, but "
+                f"{var.name} has type {var.__class__.__name__}"
+            )
+        super().__init__(cost_weight, name=name)
+        assert epsilon.ndim == 2 and epsilon.shape[1] == 1
```
rtqichen, line 264: Should probably also change this assert.

theseus/tests/test_dlm_perturbation.py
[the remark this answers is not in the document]
rtqichen: Yeah a type check there makes sense to me.

theseus/theseus_layer.py
luisenp: Do you mean because of the float requirement? I'm assuming the cost function can be more general than needed, unless you are picking up `cost_function.epsilon` for some computation later.

rtqichen: Ah I see what you mean.

PR #226, opened 2022-06-21
Vectorization handles singleton costs

luisenp:


fantaosha (commented):
LGTM. I wonder if it is possible to only vectorize cost functions with multiple instances, and keep single cost function as it is.

theseus/core/vectorizer.py
```diff
@@ -101,6 +102,16 @@ def __init__(self, objective: Objective):
             schema = _get_cost_function_schema(cost_fn)
             self._schema_dict[schema].append(wrapper)
 
+        # If all cost function groups have only one cost, there is no point in
+        # doing vectorization
+        if max([len(cfs) for cfs in self._schema_dict.values()]) <= 1:
```
mhmukadam, line 107: This is a good feature to have!

What's the behavior we expect for retraction in such a case?

I agree with Taosha, maybe in a future PR we can group only costs that repeat and leave others as is. Of course this will make the vectorization code even more complicated with maybe only a small advantage that the user doesn't need to do vectorize=False when setting up the optimizer.

theseus/tests/test_theseus_layer.py
```diff
@@ -139,6 +139,7 @@ def error_fn(optim_vars, aux_vars):
     assert not objective.vectorized
     theseus_layer = th.TheseusLayer(optimizer, vectorize=True)
     assert objective.vectorized
+    objective.disable_vectorization()
```
mhmukadam, line 142: So we are turning vectorization off for the end-to-end unit tests?

mhmukadam (approved):

theseus/tests/test_theseus_layer.py
luisenp: Yeah, they were too slow because of the extra overhead, since this tests only has one cost function. We could add an extra one where vectorization is turned on.

theseus/core/vectorizer.py
luisenp: Good point about retraction, this would disable it unnecessarily, but I think this can be easily fixed. 

I actually think that Taosha's suggestion would not be too hard to implement. I'll give it a try for a couple of hours. If it gets too complicated we can add an issue for this.

luisenp: Changed, turned out to be even easier than I thought!

mhmukadam (approved):

fantaosha (approved):
LGTM!

PR #234, opened 2022-07-12
Added state history tracking

### LINKED ISSUE #194: Support reading the state history after optimization
## 🚀 Feature
To help debug and visualize the inner loop optimization, it would be nice to have access to the optimization state after the optimization has been run.

## Motivation

In the homography example, this allows one to "playback" the optimization process to better understanding how the optimization worked. For example, it enables one to compute a visualization like this:

animation (5)

## Pitch

This is how it was implemented in the homography example:
image

But, this implementation doesn't seem to capture the state history correctly for the "implicit" mode backward pass.

luisenp:
#Closes #194

### DIFF
--- a/theseus/optimizer/nonlinear/nonlinear_optimizer.py
+++ b/theseus/optimizer/nonlinear/nonlinear_optimizer.py
@@ -49,6 +49,7 @@ class NonlinearOptimizerInfo(OptimizerInfo):
     err_history: Optional[torch.Tensor]
     last_err: torch.Tensor
     best_err: torch.Tensor
+    state_history: Optional[Dict[str, torch.Tensor]]  # variable name to state history
 
 
 class BackwardMode(Enum):
@@ -111,7 +112,10 @@ class NonlinearOptimizer(Optimizer, abc.ABC):
         return solution_dict
 
     def _init_info(
-        self, track_best_solution: bool, track_err_history: bool, verbose: bool
+        self,
+        track_best_solution: bool,
+        track_err_history: bool,
+        track_state_history: bool,
     ) -> NonlinearOptimizerInfo:
         with torch.no_grad():
             last_err = self.objective.error_squared_norm() / 2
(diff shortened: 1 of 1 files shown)


luisenp:
Note that this is different to the one implemented by @ddetone and @joeaortiz. The new version maps from variable name to a single tensor of shape (B, D, T + 1), where B is batch size, D is `var.dof()` and T is the max number of iterations. In the case of early convergence, unused entries have a value of `math.inf`.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -111,7 +112,10 @@ def _maybe_init_best_solution(
         return solution_dict
 
     def _init_info(
-        self, track_best_solution: bool, track_err_history: bool, verbose: bool
+        self,
+        track_best_solution: bool,
+        track_err_history: bool,
+        track_state_history: bool,
```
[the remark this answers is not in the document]
luisenp: `verbose` was not used in this method, not sure why `flake8` didn't pick up on this before.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -125,6 +129,22 @@ def _init_info(
...
+        if track_state_history:
+            state_history = {}
+            for var in self.objective.optim_vars.values():
+                state_history[var.name] = (
+                    torch.ones(
+                        self.objective.batch_size,
+                        var.dof(),
+                        self.params.max_iterations + 1,
+                    )
+                    * math.inf
```
mhmukadam, line 142: Is torch.empty better than inf?

mhmukadam (approved):

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: `torch.empty` will most likely initialize the tensor with pretty much random values, so it'll be very hard to figure out automatically what the EOS character is. I agree that `inf` is probably not great, but I'm not sure if this is much of an issue.

mhmukadam: Okay in that case we can just clarify in documentation since any non inf values will also be ambiguous.

PR #235, opened 2022-07-12
Unified all cost functions so that cost weight is the last non-default argument

luisenp:


mhmukadam (approved):
Looks great, thanks for finishing another painful PR!

PR #237, opened 2022-07-13
Add sphinx and readthedocs configuration

luisenp:


.readthedocs.yaml
```diff
@@ -0,0 +1,26 @@
+# .readthedocs.yaml
```
mhmukadam, line 1: Any chance this file can move inside docs/ to reduce clutter in the root dir?

mhmukadam (approved):

.readthedocs.yaml
luisenp: Looks like not 
image

PR #238, opened 2022-07-13
Ensure that CHOLMOD python interface casts to 64-bit precision

luisenp:


theseus/optimizer/autograd/tests/test_sparse_backward.py
```diff
@@ -59,3 +59,18 @@ def test_sparse_backward_step():
     )
 
     assert gradcheck(CholmodSolveFunction.apply, inputs, eps=3e-4, atol=1e-3)
+
+
+def test_float64_used():
```
fantaosha, line 64: Maybe add tests for `LUCuda` as well?

theseus/optimizer/autograd/tests/test_sparse_backward.py
```diff
@@ -59,3 +59,18 @@ def test_sparse_backward_step():
...
+    delta = CholmodSolveFunction.apply(
+        data["a"],
+        data["b"],
+        data["struct"],
+        decomp,
+        1e-06,
+    )
+
+    # With this matrix, if CHOLMOD is not casted to 64-bits, this value is > 100.0
+    assert delta.abs().max() < 1.0
```
fantaosha, line 76: I'm afraid this error depends on the specific linear solvers used. But seems not better solution exists other than solving a large-scale optimization problem.

fantaosha (approved):
LGTM. Only one minor comment

PR #240, opened 2022-07-13
Fixed bug in state history for matrix data tensors.

luisenp:


fantaosha (approved):
LGTM. Thanks for the quick fix!

PR #241, opened 2022-07-13
Added evaluation directory

luisenp:


examples/configs/tactile_pose_estimation.yaml
```diff
@@ -44,7 +51,8 @@ tactile_cost:
 
 options:
   vis_traj: True
+  vis_train: true
 
 hydra:
   run: 
-    dir: examples/outputs
\ No newline at end of file
+    dir: examples/outputs/tactile
```
mhmukadam, line 58: missing new line

experiments/evaluation/README.md
```diff
@@ -0,0 +1,18 @@
+This folder contains scripts for evaluating `theseus`'s components.
```
mhmukadam, line 1: Nice! Maybe also add a line about what each .sh script does and if there are any params involved in running the scripts.

mhmukadam: Oh I see each script has starting comment with this info. Might still be useful to have in the readme though.

mhmukadam (approved):

experiments/evaluation/README.md
luisenp: Added. @fantaosha can you confirm the descriptions for PGO look OK?

fantaosha (approved):
Removed `pose_graph_g2o`.

PR #242, opened 2022-07-13
Homography example

fantaosha:
## Types of changes

## Checklist


examples/homography_estimation.py
```diff
@@ -0,0 +1,463 @@
+import kornia
```
luisenp: Need to add copyright notice here.

examples/homography_estimation.py
```diff
@@ -0,0 +1,463 @@
...
+            img2 = torch.clamp(img2, 0.0, 1.0)
+            img1 = self.rpa.forward(img1)
+            img2 = self.rpa.forward(img2)
+
+        data = {"img1": img1[0], "img2": img2[0], "H_1_2": H_1_2[0]}
+
+        return data
+
+
+def grid_sample(image, optical):
```
luisenp, line 125: If this is code we got from somewhere else, we should add a link to the source.  And probably move it to a third party folder, since we shouldn't be claiming copyright on this. @ddetone @joeaortiz 

Does this also apply to any of the other code below?

examples/homography_estimation.py
```diff
@@ -0,0 +1,463 @@
...
+
+
+def warp_perspective_norm(H, img):
+    height, width = img.shape[-2:]
+    grid = kornia.utils.create_meshgrid(
+        height, width, normalized_coordinates=True, device=H.device
+    )
+    Hinv = torch.inverse(H)
+    warped_grid = kornia.geometry.transform.homography_warper.warp_grid(grid, Hinv)
+    # grid_sample = torch.nn.functional.grid_sample
```
luisenp, line 187: Remove commented out code.

examples/homography_estimation.py
```diff
@@ -0,0 +1,463 @@
...
+            os.system(cmd)
+            os.makedirs(dataset_path)
+            cmd = "tar -xf %s/%s -C %s" % (dataset_root, tar, dataset_path)
+            print("Running command: ", cmd)
+            os.system(cmd)
+
+    return dataset_paths
+
+
+class HomographyDataset(Dataset):
```
luisenp, line 57: Can we add some brief comments to all functions describing what they do? (both in/out of this class). Nothing too elaborate, just a couple of sentences. @ddetone @joeaortiz

examples/homography_estimation.py
```diff
@@ -0,0 +1,463 @@
...
+    viz_every = 10
+    disp_every = 10
+    save_every = 100
+    num_epochs = 999
+    outer_lr = 1e-4
+    use_cnn = True
+
+    log_dir = os.path.join(os.getcwd(), "viz")
+    os.makedirs(log_dir, exist_ok=True)
+    print("Writing to tensorboard at", os.getcwd())
```
luisenp, line 335: I don't think we need the tensorboard dependency, we can just log the loss to a csv file or something similar.

examples/homography_estimation.py
```diff
@@ -0,0 +1,463 @@
...
+
+        for _, data in enumerate(dataloader):
+
+            outer_optim.zero_grad()
+
+            img1 = data["img1"].to(device)
+            img2 = data["img2"].to(device)
+            Hgt_1_2 = data["H_1_2"].to(device)
+
+            objective = th.Objective()
```
luisenp, line 365: Is there a reason why we create the objective and theseus layer inside the loop? Couldn't we create the TheseusLayer outside, around the part where we create the cnn_model, and just populate the data using the forward method?

luisenp (changes requested):
Thanks everyone for the hard work on this!! Left some comments, mostly organization stuff.

examples/homography_estimation.py
mhmukadam: Remember to also appropriately update the examples/readme file.

examples/homography_estimation.py
fantaosha: Added.

examples/homography_estimation.py
fantaosha: Removed.

examples/homography_estimation.py
fantaosha: @ddetone @joeaortiz

examples/homography_estimation.py
fantaosha: @ddetone @joeaortiz

examples/homography_estimation.py
joeaortiz: Sounds good to me, replacing tensorboard with writing to a csv

examples/homography_estimation.py
joeaortiz: Yes, I agree with Luise, we should create the Theseus layer outside of the loop. I think it was put there before because the loop was over different problem formulations but should definitely be moved outside now.

examples/homography_estimation.py
joeaortiz: I've added some comments now describing the functions.

examples/homography_estimation.py
joeaortiz: I removed the logging of the loss altogether as it is being printed anyway.

mhmukadam (approved):
Looks great! Thank you @joeaortiz, @ddetone, @fantaosha and @luisenp for working on wrapping up this example.

mhmukadam (approved):

PR #249, opened 2022-07-18
Added option to clear cuda cache when vectorization cache is cleared.

luisenp:


examples/configs/pose_graph/pose_graph_synthetic.yaml
```diff
@@ -23,6 +23,7 @@ inner_optim:
   ratio_known_poses: 0.1
   reg_w: 1e-3
   vectorize: true
+  empty_cuda_cache: false
```
fantaosha, line 26: I would like to add `empty_cuda_cache=cfg.solver_device=='cuda' and cfg.inner_optim.solver=='dense'` to `pose_graph_synthetic.py`

examples/pose_graph/pose_graph_synthetic.py
```diff
@@ -197,10 +197,11 @@ def run(
         linearization_cls=LINEARIZATION_MODE[cast(str, cfg.inner_optim.solver)],
         linear_solver_cls=LINEAR_SOLVER_MODE[cast(str, cfg.inner_optim.solver)],
         vectorize=cfg.inner_optim.vectorize,
+        empty_cuda_cache=cfg.inner_optim.empty_cuda_cache,
     )
 
     # Set up Theseus layer
-    theseus_optim = th.TheseusLayer(optimizer, vectorize=cfg.inner_optim.vectorize)
+    theseus_optim = th.TheseusLayer(optimizer)
```
fantaosha, line 204: Add the following two lines:

`empty_cuda_cache=cfg.solver_device=='cuda' and cfg.inner_optim.solver=='dense'`
`theseus_optim = th.TheseusLayer(optimizer, empty_cuda_cache=empty_cuda_cache)`

theseus/theseus_layer.py
```diff
@@ -25,11 +25,16 @@
 
 
 class TheseusLayer(nn.Module):
-    def __init__(self, optimizer: Optimizer, vectorize: bool = True):
+    def __init__(
+        self,
+        optimizer: Optimizer,
+        vectorize: bool = True,
+        empty_cuda_cache: bool = False,
```
fantaosha, line 32: Is it possible to ignore `empty_cuda_cache` for sparse or CPU solvers?

fantaosha (commented):
LGTM. Some minor comments.

theseus/theseus_layer.py
luisenp: I don't think this is necessary, since the default is `False`. Presumably this should only be turned on if the user thinks it will be useful.

examples/pose_graph/pose_graph_synthetic.py
luisenp: This can be handled in the script via the hydra command line. We don't need to add any extra logic.

theseus/core/vectorizer.py
```diff
@@ -288,6 +289,8 @@ def _clear_wrapper_caches(self):
             for cf in cost_fn_wrappers:
                 cf._cached_error = None
                 cf._cached_jacobians = None
+        if self._empty_cuda_cache:
+            torch.cuda.empty_cache()
```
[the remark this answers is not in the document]
luisenp: Tested `torch.cuda.empty_cache()` on my Mac which doesn't have cuda and nothing happens, so I think this is safe even without that check.

PR #253, opened 2022-07-19
Adding simple example

mhmukadam:


luisenp (approved):
LGTM. Thanks for adding this.

Gralerfics:
Excuse me. Is there something wrong with the generator?
`
def y_model(x, c):
    return torch.exp(c * x)
`
modified for
`
def y_model(x, c):
    return c * torch.exp(x)
`

Gralerfics:
> Oh, good catch, thanks for pointing that out! Are you interested in submitting a PR with this fix?

Thank you very much for replying. I have submitted.

PR #257, opened 2022-07-20
Added a script for building wheels on a new docker image.

luisenp:


build_scripts/build_wheel.sh
```diff
@@ -0,0 +1,55 @@
...
+        # # Copy the wheel to host
+        CP_STR="cp"$(echo ${PYTHON_VERSION} | sed 's/[.]//g')
+        DOCKER_WHL="theseus/dist/theseus_ai-${THESEUS_VERSION}-${CP_STR}-${CP_STR}-linux_x86_64.whl"
+        HOST_WHL="theseus_ai-${THESEUS_VERSION}+cu${CUDA_SUFFIX}-${CP_STR}-${CP_STR}-manylinux_2_17_x86_64.whl"
+        sudo docker cp "${DOCKER_NAME}:${DOCKER_WHL}" ${DOCKER_DIR}/${HOST_WHL}
+
+        sudo docker rm ${DOCKER_NAME}
+        sudo docker image rm ${DOCKER_NAME}_img
+    done
+done
```
mhmukadam, line 55: Missing new line.

build_scripts/build_wheel.sh
```diff
@@ -0,0 +1,55 @@
...
+    echo >&2 "$@"
+    exit 1
+}
+[ "$#" -eq 1 ] || die "1 argument required, $# provided"
+ROOT_DIR=$1
+
+# This needs to match the version in main on Github
+THESEUS_VERSION=$(head theseus/__init__.py -n 6 | tail -n 1 | awk -F'"' '{print $2}')
+
+for PYTHON_VERSION in 3.9; do
```
mhmukadam, line 14: Maybe we have a file with the list of supported versions for python, cuda, etc where they can be pulled from instead of hardcoded in this script?

mhmukadam: We could add them to `theseus/__init__.py`.

luisenp: Not sure if this is necessary, since we are currently not using this information in any other script. At least not in a way where simply adding a new version to this file would automatically change some workflow.

mhmukadam (approved):

PR #266, opened 2022-08-08
Refactor SO3 and SE3 to be consistent with functorch

fantaosha:
## Types of changes

## Checklist


theseus/geometry/functorch.py
```diff
@@ -0,0 +1,52 @@
...
+        _FunctorchContext.set_context(mode)
+
+    def __enter__(self) -> None:
+        pass
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
+        _FunctorchContext.set_context(self.prev)
+
+
+class enable_functorch(object):
```
fantaosha, line 36: add a underscore to make it private

theseus/geometry/functorch.py
```diff
@@ -0,0 +1,52 @@
...
+    def __enter__(self) -> None:
+        self.prev = _FunctorchContext.get_context()
+        _FunctorchContext.set_context(True)
+
+    def __exit__(self, typ, value, traceback) -> None:
+        _FunctorchContext.set_context(self.prev)
+
+
+class no_functorch(_FunctorchContext):
```
fantaosha, line 45: add an underscore to make it private

theseus/geometry/functorch.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import threading
+from typing import Any
+
+
+class _FunctorchContext(object):
```
mhmukadam, line 10: Is `functorch.py` only used in `geometry` or should we move this to `core`?

mhmukadam, line 10: I don't think we are following the convention to inherit from `object`.

mhmukadam: I guess since this context is being renamed as checks for groups so this place is fine. We should do the renaming in this PR though.

theseus/geometry/manifold.py
```diff
@@ -40,7 +41,12 @@ def __init__(
         if tensor is None and dtype is None:
             dtype = torch.get_default_dtype()
         if tensor is not None:
-            tensor = self._check_tensor(tensor, strict)
+            if not _FunctorchContext.get_context():
+                tensor = self._check_tensor(tensor, strict)
+            else:
+                warnings.warn(
+                    "functorch is enabled and tensor is not checked for validness."
```
mhmukadam, line 48: How does a user 'enable' functorch use. Does this happen under the hood when you use for instance vmap from functorch? We should make the warning more descriptive. Also 'tensor' -> 'Manifold type'?

mhmukadam: This warning also might be too spammy if we raise this for every new Manifold object. Is there a way to raise this only once (or for the user to turn it off)?

theseus/geometry/se3.py
```diff
@@ -176,15 +174,22 @@ def _hat_matrix_check(matrix: torch.Tensor):
...
-            raise ValueError("The last row of hat matrices of SE3 can only be zero.")
-
-        if (
-            matrix[:, :3, :3].transpose(1, 2) + matrix[:, :3, :3]
-        ).abs().max().item() > HAT_EPS:
-            raise ValueError(
-                "The 3x3 top-left corner of hat matrices of SE3 can only be skew-symmetric."
+        if _FunctorchContext.get_context():
+            warnings.warn(
+                "functorch is enabled and the skew-symmetry of hat matrices are not checked."
```
mhmukadam, line 179: Same comment.

theseus/geometry/functorch.py
luisenp: This is only needed for compatibility with Python 2, AFAIK, so we don't need this.

theseus/geometry/se3.py
```diff
@@ -623,6 +625,165 @@ def to(self, *args, **kwargs):
         super().to(*args, **kwargs)
         self._resolve_eps()
 
+    def _deprecated_log_map_impl(
```
mhmukadam, line 628: I thought we are only making a deprecated version of the 'else' block and making the 'if' block the main version. Why is the whole thing deprecated?

theseus/geometry/so3.py
```diff
@@ -514,6 +529,180 @@ def unrotate(
 
         return ret
 
+    def _deprecated_log_map_impl(
```
mhmukadam, line 532: Same comment.

theseus/geometry/so3.py
```diff
@@ -514,6 +529,180 @@ def unrotate(
...
+            jac[:, 2, 1] += half_ret[:, 0]
+
+            diag_jac = torch.diagonal(jac, dim1=1, dim2=2)
+            diag_jac += a.view(-1, 1)
+
+            jacobians.append(jac)
+
+        return ret
+
+    def _deprecated_to_quaternion(self) -> torch.Tensor:
```
mhmukadam, line 635: Same comment.

mhmukadam (commented):
Looks good so far. Left a few comments.

theseus/geometry/manifold.py
luisenp: The `warnings` package suppresses repetitions from the same source. 

image

luisenp: BTW, we should add a warning category. I think `RuntimeWarning` is appropriate.

mhmukadam: Oh that's cool! Agreed on adding the category.

theseus/geometry/se3.py
luisenp: +1. At the very least we don't need the whole `if/else` block here, only the branch that is no longer available in the new code. But I think it's better if we encapsulate just the part that we are removing. Something like:

```python
def _log_map_impl(
        self, jacobians: Optional[List[torch.Tensor]] = None
    ) -> torch.Tensor:
    ...
    stuff = self.compute_things_for_log_map(some_inputs)
    ...
    # use stuff to compute more stuff


def _deprecated_compute_things_for_log_map(self, some_inputs):
    # the old way to do this
```

theseus/geometry/so3.py
luisenp: +1.

theseus/geometry/so3.py
luisenp: +1.

theseus/geometry/functorch.py
fantaosha: There are two PRs for functorch. I plan to create a new PR after #266 and #268 are merged.

fantaosha: Removed.

theseus/geometry/so3.py
fantaosha: Refactored.

theseus/geometry/so3.py
fantaosha: Refactored.

theseus/geometry/se3.py
fantaosha: Refactored.

theseus/geometry/functorch.py
fantaosha: Refactored.

theseus/geometry/se3.py
fantaosha: Refactored.

theseus/geometry/manifold.py
fantaosha: Refactored.

theseus/geometry/functorch.py
```diff
@@ -33,7 +33,7 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
         _FunctorchContext.set_context(self.prev)
 
 
-class enable_functorch(object):
+class enable_functorch(_FunctorchContext):
```
mhmukadam, line 36: Is deriving from `_FunctorchContext` intentional since before it was derived from `object`?

theseus/geometry/functorch.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import threading
+from typing import Any
+
+
+class _FunctorchContext:
```
mhmukadam, line 10: Given the name intended name change for the context we should likely change the name for this file too.

mhmukadam (commented):
Besides the context name update looks ready to me.

theseus/geometry/manifold.py
```diff
@@ -40,7 +41,13 @@ def __init__(
         if tensor is None and dtype is None:
             dtype = torch.get_default_dtype()
         if tensor is not None:
-            tensor = self._check_tensor(tensor, strict)
+            if not _FunctorchContext.get_context():
+                tensor = self._check_tensor(tensor, strict)
+            else:
+                warnings.warn(
+                    "functorch is enabled and tensor is not checked for {self.__class__.__name__}.",
```
luisenp, line 48: Missing `f""`

theseus/geometry/functorch.py
fantaosha: Correct.

theseus/geometry/manifold.py
fantaosha: Ah, good catch. Thanks!

mhmukadam (approved):

theseus/geometry/manifold.py
```diff
@@ -40,7 +41,14 @@ def __init__(
         if tensor is None and dtype is None:
             dtype = torch.get_default_dtype()
         if tensor is not None:
-            tensor = self._check_tensor(tensor, strict)
+            if _LieGroupCheckContext.get_context():
+                tensor = self._check_tensor(tensor, strict)
+            else:
+                warnings.warn(
+                    f"functorch is enabled and tensor is not checked "
```
luisenp, line 48: We should change this message as well, since this now has nothing to do with `functorch`. Maybe "Lie group consistency checks are disabled."? Feel free to change in #268 if that's easier.

theseus/geometry/se3.py
```diff
@@ -176,14 +174,23 @@ def _hat_matrix_check(matrix: torch.Tensor):
...
+
+            if (
+                matrix[:, :3, :3].transpose(1, 2) + matrix[:, :3, :3]
+            ).abs().max().item() > HAT_EPS:
+                raise ValueError(
+                    "The 3x3 top-left corner of hat matrices of SE3 can only be skew-symmetric."
+                )
+        else:
+            warnings.warn(
+                "functorch is enabled and the skew-symmetry of hat matrices is "
```
luisenp, line 191: Need to change this message as well. How about "Lie group checks are disabled, so the skew-symmetry... etc."?

theseus/geometry/so2.py
```diff
@@ -134,6 +136,24 @@ def _check_tensor_impl(tensor: torch.Tensor) -> bool:
...
+    def _hat_matrix_check(matrix: torch.Tensor):
+        _check = matrix.ndim == 3 and matrix.shape[1:] == (2, 2)
+
+        if _LieGroupCheckContext.get_context():
+            _check &= matrix[:, 0, 0].abs().max().item() < theseus.constants.EPS
+            _check &= matrix[:, 1, 1].abs().max().item() < theseus.constants.EPS
+            _check &= torch.allclose(matrix[:, 0, 1], -matrix[:, 1, 0])
+        else:
+            warnings.warn(
+                "functorch is enabled and the skew-symmetry of hat matrices is "
```
luisenp, line 149: Also change this message.

luisenp (approved):
LGMT, but need to change some of the warning messages. We can add this to #268 to make it easier.

theseus/geometry/manifold.py
mhmukadam: Oh yeah, good catch!

PR #268, opened 2022-08-11
Add functorch support for AutoDiffCostFunction

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,7 @@ setup_cuda10_libs: &setup_cuda10_libs
       working_directory: ~/
       command: |
         # ubuntu's default gcc9.3 is too recent for cuda10.2
+        sudo apt-get install -y gcc-7 g++-7
         sudo apt-get install -y gcc-8 g++-8
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 10
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 20
@@ -111,9 +112,17 @@ install_functorch: &install_functorch
       command: |
         pip install functorch
 
-setup_project: &setup_project
+setup_project_cuda10: &setup_project_cuda10
   - run:
-      name: Setting up project
+      name: Setting up project for cuda10
+      working_directory: ~/project
+      command: |
+        CXX=g++-7 CC=gcc-7 LD=g++-7 pip install --progress-bar off -e .
+        CXX=g++-7 CC=gcc-7 LD=g++-7 pip install -r requirements/dev.txt
+
+setup_project_cuda11: &setup_project_cuda11
+  - run:
+      name: Setting up project for cuda11
       working_directory: ~/project
       command: |
         pip install --progress-bar off -e .
@@ -199,6 +208,7 @@ jobs:
       - <<: *setup_cuda11_libs
       - <<: *setup_environment
       - <<: *install_torch_cuda11
+      - <<: *install_functorch
       - <<: *setup_project
       - <<: *run_tests
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,7 +94,7 @@ install_torch_cuda10: &install_torch_cuda10
       name: Install Torch for cuda10
       working_directory: ~/project
       command: |
-        pip install --progress-bar off torch==1.10.0+cu102 torchvision==0.11.1+cu102 torchaudio==0.10.0+cu102 -f https://download.pytorch.org/whl/cu102/torch_stable.html
+        pip install --progress-bar off torch==1.12.1+cu102 torchvision==0.13.1+cu102 torchaudio==0.12.1+cu102 -f https://download.pytorch.org/whl/cu102/torch_stable.html
         python -c 'import torch; print("Torch version:", torch.__version__); assert torch.cuda.is_available()'
 
 install_torch_cuda11: &install_torch_cuda11
@@ -102,7 +102,7 @@ install_torch_cuda11: &install_torch_cuda11
       name: Install Torch for cuda11
       working_directory: ~/project
       command: |
-        pip install --progress-bar off torch==1.10.0+cu113 torchvision==0.11.1+cu113 torchaudio==0.10.0+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
+        pip install --progress-bar off torch==1.12.1+cu113 torchvision==0.13.1+cu113 torchaudio==0.12.1+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
         python -c 'import torch; print("Torch version:", torch.__version__); assert torch.cuda.is_available()'
 
 install_functorch: &install_functorch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,7 +204,6 @@ jobs:
       - checkout
       - <<: *update_and_install_python
       - <<: *install_suitesparse
-      - <<: *install_functorch
       - <<: *setup_cuda11_libs
       - <<: *setup_environment
       - <<: *install_torch_cuda11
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -158,6 +167,7 @@ jobs:
     steps:
       - checkout
       - <<: *install_suitesparse
+      - <<: *install_functorch
```
luisenp, line 170: Need to add functorch to the gpu tests also. For the cpu test we are using `nox` which runs `pip install -e .`, so we also need to add `functorch` to the main requirements file.

theseus/core/cost_function.py
```diff
@@ -187,11 +203,42 @@ def _compute_autograd_jacobian(
...
         )
 
-    # Returns (jacobians, error)
+    def _make_jac_fn_functorch(
+        self, tmp_optim_vars: Tuple[Manifold, ...], tmp_aux_vars: Tuple[Variable, ...]
+    ):
+        def jac_fn(optim_vars_tensors_, aux_vars_tensors_):
+            assert len(optim_vars_tensors_) == len(tmp_optim_vars)
+
+            with enable_functorch():
```
luisenp, line 212: Lots going on here, it would be useful to have some comments explaining what's happening. This is what I understand:

- Need to disable Lie Group checks that are not compatible with `functorch`, which is why we use the context.
- `functorch` passes tensors that already indexed within the batch, which means we need to add back the batch dimension with `unsqueeze(0)`.
- The error function will return a batch dimension, but `functorch` expects no batch dimension, hence we return `[0]`. 
- Unlike the other methods, we need to update the `aux_vars` every jacobian call so that the variable contains **only** the current batch element being evaluated. 

Is this correct? If so, can. you add this explanations at the top of `_make_jac_fn_functorch`?

theseus/geometry/tests/common.py
```diff
@@ -21,13 +23,79 @@ def check_exp_map(tangent_vector, group_cls, atol=TEST_EPS, enable_functorch=Fal
             atol=atol,
         )
 
+    if enable_functorch:
```
luisenp, line 26: Move and refactor tests as discussed offline.

luisenp (changes requested):
Left some small comments, but LGTM!

theseus/core/cost_function.py
fantaosha: From Luis "you can just have a single function body for this test, like this one, and just loop over the LieGroup class and group function you want to test"

theseus/geometry/tests/test_se2.py
```diff
@@ -34,55 +34,72 @@ def create_random_se2(batch_size, rng, dtype=torch.float64):
...
+            th.SE2,
+            atol=ATOL,
+            enable_functorch=enable_functorch,
+        )
 
 
 @pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
-def test_log_map(dtype):
+@pytest.mark.parametrize("enable_functorch", [True, False])
+def test_log_map(dtype, enable_functorch):
```
mhmukadam, line 60: I am wondering what the usage looks like from the user side. Does the user pass `enable_functorch` to a `AutoDiffCostFunction` or is the flag internally set when you pass `autograd_functorch = True`.

We should also add the end-to-end or at least the `AutoDiffCostFunction` unit test for when using `autograd_functorch` and also `autograd_loop_over_batch` if we haven't already.

Finally I think it would be useful setup a benchmark script `/evaluations` to quickly run a comparison of the variant with dense, loop over batch, and functorch, so that we can evaluate future updates to our code or functorch more easily.

mhmukadam (commented):
Looks good so far. Let's wrap the lie group refactor PR and then address the comments here.

theseus/core/cost_function.py
```diff
@@ -187,11 +203,42 @@ def _compute_autograd_jacobian(
...
+
+    def _compute_autograd_jacobian_functorch(
+        self,
+        optim_tensors: Tuple[torch.Tensor, ...],
+        aux_tensors: Tuple[torch.Tensor, ...],
+        jac_fn: Callable,
+    ) -> Tuple[torch.Tensor, ...]:
+        return vmap(jacrev(jac_fn, argnums=0))(optim_tensors, aux_tensors)
+
+        # Returns (jacobians, error)
```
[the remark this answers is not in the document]
fantaosha: Removed.

theseus/core/cost_function.py
```diff
@@ -205,17 +252,21 @@ def jacobians(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
                 )
                 jacobians_raw_loop.append(jacobians_n)
 
-            jacobians_full = [
-                torch.cat([jac[k][:, :, 0, :] for jac in jacobians_raw_loop], dim=0)
-                for k in range(len(optim_vars))
-            ]
+            jacobians_full = tuple(
```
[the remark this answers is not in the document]
fantaosha: Removed. Thanks for catching this.

theseus/core/cost_function.py
```diff
@@ -133,11 +138,21 @@ def __init__(
...
         self._tmp_aux_vars_for_loop = None
 
-        if autograd_loop_over_batch:
+        self._autograd_loop_over_batch = autograd_loop_over_batch
+        self._autograd_functorch = autograd_functorch
+
+        if autograd_functorch and self._autograd_loop_over_batch:
+            self._autograd_loop_over_batch = False
+            warnings.warn(
+                "autograd_loop_over_batch has been reset to be false due to "
```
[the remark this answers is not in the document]
fantaosha: Replaced. Thanks!

.circleci/config.yml
fantaosha: Now it works for CircleCI.

theseus/geometry/tests/test_se2.py
fantaosha: This is internally set. I'm not sure if we still another test for AutoDiffCostFunction with `autograd_functorch=True` since these tests in Lie groups have already covered this. 

There are tests for `autograd_loop_over_batch=True`.

theseus/core/cost_function.py
```diff
@@ -133,11 +138,21 @@ def __init__(
...
         self._tmp_aux_vars_for_loop = None
 
-        if autograd_loop_over_batch:
+        self._autograd_loop_over_batch = autograd_loop_over_batch
+        self._autograd_functorch = autograd_functorch
+
+        if autograd_functorch and self._autograd_loop_over_batch:
+            self._autograd_loop_over_batch = False
+            warnings.warn(
+                "autograd_loop_over_batch has been reset to be False due to "
```
luisenp, line 151: I don't love this warning message, but I'm struggling to find a good alternative. Any suggestions from @mhmukadam?

I am thinking something like "autograd_use_functorch=True overrides given autograd_loop_over_batch=True, so the latter will be set to False".

luisenp (approved):
LGTM. Left a minor comment.

theseus/core/cost_function.py
mhmukadam: Yeah what you suggest sounds clearer.

mhmukadam (approved):
Looks good. Don't forget to add the profiling script comparing time and memory between dense, loop over batch, and vmap versions for a simple AutoDiffCostFcuntion example.

theseus/geometry/manifold.py
```diff
@@ -45,7 +45,7 @@ def __init__(
                 tensor = self._check_tensor(tensor, strict)
             else:
                 warnings.warn(
-                    f"functorch is enabled and tensor is not checked "
+                    f"Lie group consistency checks are disabled "
```
mhmukadam, line 48: Lie group -> Manifold? Since the class is Manifold.

fantaosha: Good catch. Fixed.

mhmukadam (approved):

PR #272, opened 2022-08-18
Added an option to pass extra costs to MotionPlanner.

luisenp:
Need to add some unit tests before merging this. We need a callback because after creating the optimizer our code doesn't allow more costs to be added.

### DIFF
--- a/theseus/utils/examples/motion_planning/motion_planner.py
+++ b/theseus/utils/examples/motion_planning/motion_planner.py
@@ -24,9 +24,9 @@ class MotionPlanner:
         max_optim_iters: int,
         step_size: float = 1.0,
         use_single_collision_weight: bool = True,
+        obj_modifier_callback: Optional[Callable[[th.Objective], None]] = None,
         device: str = "cpu",
         dtype: torch.dtype = torch.double,
-        extra_costs_callback: Optional[Callable[[th.Objective], None]] = None,
     ):
         self.map_size = map_size
         self.epsilon_dist = epsilon_dist
@@ -173,8 +173,8 @@ class MotionPlanner:
                 )
             )
         # Add any extra user-provided cost functions before creating optimizer
-        if extra_costs_callback is not None:
-            extra_costs_callback(objective)
+        if obj_modifier_callback is not None:
+            obj_modifier_callback(objective)
 
         # Finally, create the Nonlinear Least Squares optimizer for this objective
         # and wrap both into a TheseusLayer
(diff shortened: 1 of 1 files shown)


theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -171,6 +172,9 @@ def __init__(
                     )
                 )
             )
+        # Add any extra user-provided cost functions before creating optimizer
+        if extra_costs_callback is not None:
+            extra_costs_callback(objective)
```
luisenp, line 177: We could add a warning if no extra costs were added (i.e., num of cost functions is the same before and after)?

cpaxton: Should we also be disabling the goal cost if we do this?

cpaxton (approved):

theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -171,6 +172,9 @@ def __init__(
                     )
                 )
             )
+        # Add any extra user-provided cost functions before creating optimizer
+        if obj_modifier_callback is not None:
```
mhmukadam, line 176: Looks good. Agreed on merging after adding some unit test.

Can you briefly explain the intended behavior for when the callback is triggered?

luisenp: I realized later that none of this code has any unit tests in place, since this is mostly meant to be taken as examples of how to use the library. The add on in this PR seems to be working well on our private experiment repo, so I propose merging as is. 

For the explanation, do you mean in the comments? Or here?

mhmukadam: Sounds good on the tests in that case.

For explanation, actually, yes both :)

luisenp: This lets users add other cost functions to the motion planner or even remove some of the default ones. We need to do it this way because the optimizer gets created inside the MotionPlanner, which locks the objective from allowing further changes.

mhmukadam: Ah I see. Hmm another alternative could be that a MotionPlannerObjective class gives you a base objective you can modify as you see fit. Then you can pass to a wrapper class MotionPlannerLayer that gives you back a TheseusLayer version of your modified objective. Basically splitting the current functionality in to two steps.

luisenp: I like that suggestion, I think it's cleaner. Let's go with that.

luisenp: See #277

PR #277, opened 2022-08-24
Refactored MotionPlanner so that objective can be passed separately.

luisenp:
## Motivation and Context
Motion planner objective can be created separately, allowing one to modify the objective before wrapping in an optimizer and `TheseusLayer`. Note that this is a breaking change, old `MotionPlanner` code won't work anymore. 

Closes #272

### DIFF
--- a/theseus/utils/examples/motion_planning/motion_planner.py
+++ b/theseus/utils/examples/motion_planning/motion_planner.py
@@ -11,7 +11,7 @@ import torch
 import theseus as th
 
 
-class MotionPlanner:
+class MotionPlannerObjective(th.Objective):
     def __init__(
         self,
         map_size: int,
@@ -20,25 +20,27 @@ class MotionPlanner:
         collision_weight: float,
         Qc_inv: List[List[int]],
         num_time_steps: int,
-        optim_method: str,
-        max_optim_iters: int,
-        step_size: float = 1.0,
         use_single_collision_weight: bool = True,
-        device: str = "cpu",
         dtype: torch.dtype = torch.double,
     ):
+        for v in [
+            map_size,
+            epsilon_dist,
+            total_time,
+            collision_weight,
+            Qc_inv,
+            num_time_steps,
+        ]:
+            assert v is not None
+
+        super().__init__(dtype=dtype)
         self.map_size = map_size
         self.epsilon_dist = epsilon_dist
         self.total_time = total_time
         self.collision_weight = collision_weight
         self.Qc_inv = copy.deepcopy(Qc_inv)
         self.num_time_steps = num_time_steps
-        self.optim_method = optim_method
-        self.max_optim_iters = max_optim_iters
-        self.step_size = step_size
         self.use_single_collision_weight = use_single_collision_weight
-        self.device = device
-        self.dtype = dtype
 
         self.trajectory_len = num_time_steps + 1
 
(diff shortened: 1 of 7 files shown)


theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -11,7 +11,7 @@
 import theseus as th
 
 
-class MotionPlanner:
+class MotionPlannerObjective(th.Objective):
```
mhmukadam, line 14: Nice this does look more streamlined. We can potentially do this for other example later.

Now we are starting to really have something like downloadable pre-trained models that can be then further customized!

theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -20,25 +20,27 @@ def __init__(
...
+            map_size,
+            epsilon_dist,
+            total_time,
+            collision_weight,
+            Qc_inv,
+            num_time_steps,
+        ]:
+            assert v is not None
+
+        super().__init__(dtype=dtype)
```
mhmukadam, line 36: Did we miss `device`?

theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -172,25 +171,64 @@ def __init__(
                 )
             )
 
+
+class MotionPlanner:
```
mhmukadam, line 175: Since this abstraction is the same as before does this PR cause a breaking change in the API?

tutorials/05_differentiable_motion_planning.ipynb
```diff
@@ -17,8 +19,37 @@
...
    "execution_count": 1,
+   "id": "48c45eca",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The Zen of Python, by Tim Peters\n",
```
mhmukadam, line 29: Same

mhmukadam (commented):
Looks great!

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: We don't have a kwarg `device=` in `th.Objective` (maybe we should add later?). In the `MotionPlanner` the objective is pushed to the given device after the objective is created. The code does assume that all inputs are on the same device.

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: It does because the order of args changed in the constructor.

tutorials/04_motion_planning.ipynb
```diff
@@ -22,8 +24,37 @@
...
    "execution_count": 1,
+   "id": "51ef8453",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The Zen of Python, by Tim Peters\n",
```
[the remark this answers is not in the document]
luisenp: wth, this must have been some record breaking autocomplete by vscode lol. This is a pretty famous set of principles, though. Maybe vscode is passive-aggressively telling me get better at coding.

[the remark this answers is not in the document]
luisenp: Removed.

tutorials/05_differentiable_motion_planning.ipynb
luisenp: Removed.

tutorials/04_motion_planning.ipynb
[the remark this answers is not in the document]
luisenp: So, turns out if you write `import this` in python, it prints out the Zen of Python. So, I guess vscode autocompleted "th" as "this" as some point and I didn't notice.

theseus/utils/examples/motion_planning/motion_planner.py
mhmukadam: Okay. Hmm might be worth having a device in Objective, later.

mhmukadam (approved):

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: Created #281

PR #278, opened 2022-08-25
Add support for SE2 poses in Collision2D

luisenp:
There is a breaking change in this PR, since `SE2.xy()` now returns a `Point2` object instead of a tensor. 

This should be merged after #277

### DIFF
--- a/theseus/embodied/collision/tests/test_collision_factor.py
+++ b/theseus/embodied/collision/tests/test_collision_factor.py
@@ -76,18 +76,14 @@ def test_collision2d_copy():
     assert cost_function2.name == "new_name"
 
 
-@pytest.mark.parametrize("pose_cls", [th.Vector, th.SE2])
+@pytest.mark.parametrize("pose_cls", [th.Point2, th.SE2])
 def test_collision2d_jacobians(pose_cls):
     rng = torch.Generator()
     rng.manual_seed(0)
     for _ in range(10):
         for batch_size in [1, 10, 100]:
             cost_weight = th.ScaleCostWeight(torch.ones(1).squeeze().double())
-            if pose_cls == th.Vector:
-                pose = th.Point2.randn(batch_size, generator=rng, dtype=torch.float64)
-            else:
-                pose = th.SE2.randn(batch_size, generator=rng, dtype=torch.float64)
-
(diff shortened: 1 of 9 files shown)


theseus/embodied/collision/tests/test_collision_factor.py
```diff
@@ -76,13 +76,18 @@ def test_collision2d_copy():
     assert cost_function2.name == "new_name"
 
 
-def test_collision2d_jacobians():
+@pytest.mark.parametrize("pose_cls", [th.Vector, th.SE2])
```
mhmukadam, line 79: Why not make th.Vector -> th.Point2 directly?

theseus/geometry/se2.py
```diff
@@ -136,7 +136,7 @@ def theta(self, jacobians: Optional[List[torch.Tensor]] = None) -> torch.Tensor:
     def translation(self) -> Point2:
         return Point2(tensor=self[:, :2])
 
-    def xy(self, jacobians: Optional[List[torch.Tensor]] = None) -> torch.Tensor:
+    def xy(self, jacobians: Optional[List[torch.Tensor]] = None) -> Point2:
```
mhmukadam, line 139: I guess now it is same as translation but also gives jacobian. Maybe we add jacobian to translation and alias xy (or vice-versa)?

mhmukadam (approved):
Looks good.

theseus/embodied/collision/tests/test_collision_factor.py
luisenp: Ah, sorry, when I first wrote this I wasn't sure if there was a `randn` method for `th.Point2`. I'll fix.

luisenp: Fixed.

theseus/geometry/se2.py
luisenp: Good catch. I'll refactor a bit.

luisenp: I made `translation` an alias for `xy()`.

theseus/embodied/collision/tests/test_collision_factor.py
mhmukadam: No worries. Thanks for the fix.

mhmukadam (approved):

PR #279, opened 2022-08-25
Add cost function that computes Local(se2.xy(), point_2_target)

luisenp:
## Motivation and Context
Planning to use this for SE2 motion planning with an (x, y) goal but free angle.

### DIFF
--- a/theseus/embodied/misc/local_cost_fn.py
+++ b/theseus/embodied/misc/local_cost_fn.py
@@ -8,7 +8,7 @@ from typing import List, Optional, Tuple
 import torch
 
 from theseus import CostFunction, CostWeight
-from theseus.geometry import LieGroup
+from theseus.geometry import LieGroup, Point2, SE2
 
 
 class Local(CostFunction):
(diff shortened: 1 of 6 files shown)


theseus/embodied/misc/local_cost_fn.py
```diff
@@ -48,3 +53,45 @@ def _copy_impl(self, new_name: Optional[str] = None) -> "Local":
         return Local(  # type: ignore
             self.var.copy(), self.target.copy(), self.weight.copy(), name=new_name
         )
+
+
+class XYDifference(CostFunction):
```
mhmukadam, line 58: I am wondering if we can streamline the name/convention.

In a sense this is basically Local but on a subset of dims. Maybe we can have Local take in optionally a list of dims to apply cost to and when empty it operates on all dims?

If we want to stick to this version then the name should have SE2 somewhere. Maybe `XYDifferenceSE2`?

mhmukadam (commented):
Looks good aside from the one comment.

theseus/embodied/misc/local_cost_fn.py
luisenp: I've been thinking a lot about this, and my current thought is that what we really want is something like this:
```python
   cf1 = th.Transform(th.Difference(se2_1, target, ...), [th.xy])
   cf2 = th.Transform(th.Between(se2_1, point2_1, meas, ...), [th.xy, None]). # for 2 optim vars
```
where `th.Transform` is a cost function wrapper that applies some optional transformations to the optimization variables before evaluating the error.  Note that for `th.Difference` a list of dims is not enough because we need jacobians for the transformation (e.g., for `xy()` the jacobian is the rotation matrix). So what we really want is some composition of functions, and this can be made more general with something like the transform above. 

I see a way to implement this that will maintain backward compatibility, but requires some refactoring of the `CostFunction` abstraction and subclasses, so it might be a bit painful. But if you like this idea, my proposal would be to move the currently named  to `XYDifference ` to the motion planning module as a private function, so we can make progress on the SE2 motion planning, and I can work on this other thing in parallel.

mhmukadam: Oh excellent point! I like this idea.

This is also relevant to costs that involve kinematics: for instance, you do FK to go from joint space vector to end effector pose SE3 and want to do a diff there with a desired end effector SE3 pose, [see here](https://github.com/gtrll/gpmp2/blob/main/gpmp2/kinematics/GaussianPriorWorkspacePose.h#L64). Similarly diff for end effector Point3 or SO3.

Let's go with the planning only version you propose to unblock for now. The ideal version requires more discussion, cc @fantaosha.

PR #282, opened 2022-08-26
Add SE2 support in MotionPlanner 

luisenp:


mhmukadam (approved):

theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -337,36 +339,58 @@ def get_variable_values_from_straight_line(
...
     ) -> Dict[str, torch.Tensor]:
         # Returns a dictionary of variable names to values, so that values
-        # are assigned with the data from the given trajectory. Trajectory should be a
+        # are assigned with the data from the given trajectory.
+        # For Point2 trajectories, trajectory should be a
         # tensor of shape (batch_size, 4, planner.trajectory_len).
-        assert trajectory.shape[1:] == (4, self.objective.trajectory_len)
+        # For SE2 trajectories, it should be a tensor of shape
+        # tensor of shape (batch_size, 7, planner.trajectory_len).
+        pose_numel = 2 if self.objective.pose_type == th.Point2 else 4
```
mhmukadam, line 383: you could use the Manifold.numel() available in main now?

mhmukadam (approved):

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
     rev: v0.931
     hooks:
       - id: mypy
-        additional_dependencies: [torch==1.9.0, tokenize-rt==3.2.0, types-PyYAML, types-mock]
+        additional_dependencies: [torch, tokenize-rt==3.2.0, types-PyYAML, types-mock]
```
luisenp, line 16: Had to add this because `so3` was giving me a mypy error.  `torch.pi` was only added in torch 1.10, but pre-commit hooks was using `1.9.0`. @fantaosha you never saw this error?

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: Changed.

.pre-commit-config.yaml
mhmukadam: Strange though since all the CI checks passed in 286.

luisenp: I think this only affects pre-commit hooks, not CI (which uses `nox`).

mhmukadam: Ah okay. If this is fixed here then we are set. I suspect some of these will need updating based on the vmap PRs.

PR #293, opened 2022-09-07
Fixed bug in visualization of SE2 motion plans.

luisenp:


theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -349,7 +349,7 @@ def get_variable_values_from_straight_line(
                 cur_pos = start[:, :2] + unit_trajectory_len * i
                 input_dict[f"pose_{i}"] = torch.cat([cur_pos, start[:, 2:]], dim=1)
                 input_dict[f"vel_{i}"] = torch.cat(
-                    [avg_vel, torch.zeros(avg_vel.shape[0], 1)], dim=1
+                    [avg_vel, torch.zeros_like(avg_vel[:, :1])], dim=1
```
mhmukadam, line 352: Should we just replace all torch.zeros to torch.zeros_like everywhere in the repo?

mhmukadam (approved):

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: I don't think that's necessary. In this case I used `zeros_like` to have the same device and dtype w/o passing it explicitly. For `functorch` we use `new_zeros` because that's compatible with `vmap`. I do think both of those make the code cleaner than using `torch.zeros(..., device=device, dtype=dtype)`, but probably not worth the effort to go and change it all over the code if there is no associated bug.

PR #298, opened 2022-09-16
Changed unit tests so that the batch sizes to tests are defined in a central import

luisenp:
## Motivation and Context
Faster unit tests.

### DIFF
--- a/theseus/embodied/measurements/tests/test_between.py
+++ b/theseus/embodied/measurements/tests/test_between.py
@@ -130,7 +130,7 @@ def test_error_between_se2():
 
 
 def test_jacobian_between_se3():
-    for batch_size in [1, 20, 100]:
+    for batch_size in [1, 10]:
         aux_id = torch.arange(batch_size)
         se3_1 = th.SE3.rand(batch_size, dtype=torch.float64)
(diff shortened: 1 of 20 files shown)


theseus/embodied/measurements/tests/test_between.py
```diff
@@ -130,7 +130,7 @@ def test_error_between_se2():
 
 
 def test_jacobian_between_se3():
-    for batch_size in [1, 20, 100]:
+    for batch_size in [1, 10]:
```
mhmukadam, line 133: This [1, 10] actually shows up so often that I am wondering if we should add it to constants.py?

mhmukadam (approved):

theseus/embodied/measurements/tests/test_between.py
luisenp: Added a constant to `theseus.core.tests.common` and replaced other similar instances with this. Now all tests use `BATCH_SIZES_TO_TEST = (1, 10)`.

mhmukadam (approved):

fantaosha (approved):
LGTM.

PR #302, opened 2022-09-20
Cleanup Baspacho setup

luisenp:
To build wheels with Baspacho, users can now do 

`BASPACHO_ROOT_DIR=/my/path/ python -m build --no-isolation`

I'm still unclear how does this interact with `pip install -e .`.


maurimo:
Tested. Works, thanks a lot, merging!

Issue #301, opened 2022-09-20
The efficiency of objective.add(CostFunction)


Christopher6488:
## ❓ Questions and Help
I want to construct 20000+ optimization problem and call objective.add(CostFunction) for each problem in a for loop. I found that the loop velocity is slower and slower, and objective.add(CostFunction) functions are increasingly time-consuming. The total time is about 200+ seconds. How can I fix this problem?

Christopher6488:
I have found the reason and fixed it. I will launch a PR.

mhmukadam:
@Christopher6488 thanks for raising this. Yes the PR would be great to understand the details and also great that you found a fix too!

Christopher6488:
> 

Thanks for replying! I have two questions:I have run `python -m pytest theseus -m "not cudaext"` to conduct unit test after fixing the issue. However, the terminal logs `theseus/tests/test_theseus_layer.py:366: AssertionError`.  This error seems to have nothing to do with my changes(even if I don't make any changes, the error still exists). Thus, how will I fill out the Checklist in PULL_REQUEST_TEMPLATE.MD? The changes in Objective.add(CostFunction) are only to improve the execution speed and  does not bring about functional changes. Do

Christopher6488:
Thanks for replying! I have two questions:
1. I have run python -m pytest theseus -m "not cudaext" to conduct unit test after fixing the issue. However, the terminal logs theseus/tests/test_theseus_layer.py:366: AssertionError. This error seems to have nothing to do with my changes(even if I don't make any changes, the error still exists). Thus, how will I fill out the Checklist in PULL_REQUEST_TEMPLATE.MD?

2. The changes in Objective.add(CostFunction) are only to improve the execution speed and does not bring about functional changes. Do I need to add test? The test in test_objective.py has tested the safety of Objective.add(CostFunction).

luisenp:
Hi @Christopher6488, thanks for the contribution! If you open a PR with your current version I'm happy to take a look at give some feedback. It'd be easier to answer your questions if I can look at the code.

PR #303, opened 2022-09-21
enhance the efficiency of Objectve.add()

Christopher6488:
## Motivation and Context

Issue: https://github.com/facebookresearch/theseus/issues/301

Motivation: Objective.add(CostFunction) execution time will increase with the number of optimization problems. If I want to construct 20000+ odometry constraints, it will need 200+ seconds. It is too slow. The reason is that time consuming of https://github.com/facebookresearch/theseus/blob/main/theseus/core/objective.py#L198 increases with the number of problems.

Solution: We only need to judge whther the newly appended aux/optim variables are in the  optim/aux variables. Not need to compare all  variables.

## How Has This Been Tested

Safety: The modificated Objective has passed the unit test in test_objective.py. It means that the Objective.add(CostFunction) could detect optim variables and aux variables with duplicate names.

Time consuming: I constructed 29312 odometry constraints(correspond to 29312 optimization problems). The time consuming before changes is about 200+ seconds. The time consuming after changes is about 0.5 seconds.

## Types of changes

## Checklist

### DIFF
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import warnings
+import itertools
 from collections import OrderedDict
 from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Union
 
@@ -101,6 +102,7 @@ class Objective:
             function_vars = function.aux_vars  # type: ignore
             self_var_to_fn_map = self.functions_for_aux_vars  # type: ignore
             self_vars_of_this_type = self.aux_vars  # type: ignore
+
         for variable in function_vars:
             # Check that variables have name and correct dtype
             if variable.name is None:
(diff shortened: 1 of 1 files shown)


theseus/core/objective.py
```diff
@@ -195,11 +211,18 @@ def add(self, cost_function: CostFunction):
 
         self.cost_functions_for_weights[cost_function.weight].append(cost_function)
 
-        if self.optim_vars.keys() & self.aux_vars.keys():
-            raise ValueError(
-                "Objective does not support a variable being both "
-                "an optimization variable and an auxiliary variable."
-            )
+        for optim_name in self._new_optim_var_name:
```
luisenp, line 214: Good catch and thanks for the fix! The overall logic is good. I think it would be better to do this w/o introducing any new member variables. You can compute two sets before line 214, in pseudocode:

optim_vars := cost_function.optim_vars + cost_function.weight.optim_vars
aux_vars := cost_function.aux_vars + cost_function.weight.aux_vars

WIth this you can do the two checks you introduced.

theseus/core/objective.py
```diff
@@ -195,11 +211,18 @@ def add(self, cost_function: CostFunction):
...
+            if self.has_aux_var(optim_name):
+                raise ValueError(
+                    "Objective does not support a variable being both "
```
luisenp, line 217: Since you are reusing this message, better to define a variable called `dual_var_err_msg` and reference it in both `ValueError`.

luisenp (changes requested):
Thanks a lot for this fix! I left a couple of comments. I don't think we need to introduce new unit tests, since it's existing functionality. 

BTW, you don't need to modify the pull request template in the code, you should do it directly in the pull request form when submitting the PR, otherwise you are changing the template for all future PRs.

Also, all CIs tests passed. The error you were seeing might have been some flaky unit tests using random numbers.

mhmukadam (commented):
Thanks for the fix, @Christopher6488, 200+ -> 0.5s speedup is awesome! Overall looks good and we can merge once you address @luisenp's comments.

> error you were seeing might have been some flaky unit tests using random numbers

Over time we have been fixing seeds to handle such flaky tests. Can we figure out which test was failing so we can fix that too.

theseus/core/objective.py
```diff
@@ -195,11 +197,28 @@ def add(self, cost_function: CostFunction):
 
         self.cost_functions_for_weights[cost_function.weight].append(cost_function)
 
-        if self.optim_vars.keys() & self.aux_vars.keys():
-            raise ValueError(
-                "Objective does not support a variable being both "
-                "an optimization variable and an auxiliary variable."
+        optim_vars_name = [
```
luisenp, line 200: nit: `optim_vars_names` and `aux_var_names`.

luisenp (approved):
LGTM! Just left one nitpicky comment. Thanks a lot!

mhmukadam (approved):

PR #305, opened 2022-09-26
Rename BackwardMode.FULL --> UNROLL and simplify backward mode config

luisenp:
We can now pass strings to specify backward modes, instead of having to pass the specific enum type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ outer_optimizer = torch.optim.RMSprop([phi], lr=0.001)
 for epoch in range(10):
     solution, info = layer.forward(
         input_tensors={"x": phi.clone(), "v": torch.ones(1, 1)},
-        optimizer_kwargs={"backward_mode": th.BackwardMode.IMPLICIT})
+        optimizer_kwargs={"backward_mode": "implicit"})
     outer_loss = torch.nn.functional.mse_loss(solution["v"], v_true)
     outer_loss.backward()
     outer_optimizer.step()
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -141,7 +141,7 @@ outer_optimizer = torch.optim.RMSprop([phi], lr=0.001)
 for epoch in range(10):
     solution, info = layer.forward(
         input_tensors={"x": phi.clone(), "v": torch.ones(1, 1)},
-        optimizer_kwargs={"backward_mode": th.BackwardMode.IMPLICIT})
+        optimizer_kwargs={"backward_mode": "implicit"})
```
mhmukadam, line 144: We should update this example in the paper and the website following this.

theseus/optimizer/nonlinear/tests/test_backwards.py
```diff
@@ -111,7 +111,7 @@ def fit_x(data_x_np):
         optimizer_kwargs={
             "track_best_solution": True,
             "verbose": False,
-            "backward_mode": th.BackwardMode.TRUNCATED,
+            "backward_mode": "TRUNCATED",
```
mhmukadam, line 114: nit: TRUNCATED -> truncated

mhmukadam (approved):
Looks great!

theseus/optimizer/nonlinear/tests/test_backwards.py
luisenp: Oh, I was just adding a call with uppercase to add coverage for this case.

PR #309, opened 2022-09-27
update readme and bump version

mhmukadam:


luisenp (approved):
LGTM. Do you mind changing `RMSProp` for `Adam` in the code snippet? It's weird to me that I left it as `RMSProp`, feels outdated. I'll change in the paper also.

PR #313, opened 2022-09-30
Set vmap as the default autograd mode for autodiff cost function.

luisenp:


mhmukadam (approved):
Looks good. CI is failing though.

theseus/tests/test_theseus_layer.py
```diff
@@ -125,6 +125,7 @@ def error_fn(optim_vars, aux_vars):
                 th.Vector(cost_weight_dim, name="learnable_err_param", tensor=cw_data)
             ],
             autograd_vectorize=True,
+            autograd_mode="dense",
```
luisenp, line 128: @fantaosha In [this](https://github.com/facebookresearch/theseus/blob/main/theseus/core/cost_function.py#L243) line, `functorch` throws an error when the batch size is not the same for all variables. We could consider automatically broadcasting before passing to `vmap`. Roughly something like the code below (not a clean version)

image

Any concerns about adding something along these lines in a separate PR?

theseus/embodied/kinematics/tests/test_inverse_kinematics.py
```diff
@@ -71,7 +71,12 @@ def ee_pose_err_fn(optim_vars, aux_vars):
     aux_vars = (ee_pose_target,)
 
     cost_function = th.AutoDiffCostFunction(
-        optim_vars, ee_pose_err_fn, 6, aux_vars=aux_vars, name="ee_pose_err_fn"
+        optim_vars,
+        ee_pose_err_fn,
+        6,
+        aux_vars=aux_vars,
+        name="ee_pose_err_fn",
+        autograd_mode="dense",
```
mhmukadam, line 79: This does not work with loop_batch or vmap?

mhmukadam: DRM dependency is incompatible.

theseus/tests/test_theseus_layer.py
fantaosha: LGTM.

PR #324, opened 2022-10-10
Add support for BaSpaCho sparse solver

maurimo:
## Motivation and Context

We would like to have the batched sparse solver BaSpaCho be supported by mainline Theseus.

## How Has This Been Tested

Tests added to the CI testing the solver in a simple example, a wider set of cases, the autograd, and the linear solver class on a wise set of examples and verifying that torch gradient checks. Tests target both CPU and GPU.

## Types of changes

- [X] Docs change / refactoring / dependency upgrade
- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have completed my CLA (see **CONTRIBUTING**)
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


README.md
```diff
@@ -107,11 +107,18 @@ pip install -e ".[dev]"
 ```
 and follow the more detailed instructions in [CONTRIBUTING](https://github.com/facebookresearch/theseus/blob/main/CONTRIBUTING.md).
 
+Compilation from source supports is affected by the following enviroment variables:
+* `BASPACHO_ROOT_DIR`: root dir of [BaSpaCho](https://github.com/facebookresearch/baspacho) sparse solver, which must be already
+  built with binaries in the subdirectory `build`.
+* `THESEUS_ENABLE_CUDA` (1 or 0): if set, forces enabling or disabling Cuda support regardless of system's `torch.cuda.is_available()`.
```
mhmukadam, line 113: Can we make this easier for users?

Can we do `pip install -e .` like before and cpu baspacho is installed while gpu baspacho is installed if cuda is available just like the current extlib stuff.

For these env variables can we add reasonable defaults in setup.py that could overridden from command line with a flag after `pip install -e .` if needed.

README.md
```diff
@@ -107,11 +107,18 @@ pip install -e ".[dev]"
...
+
 ### Running unit tests (requires `dev` installation)
 ```bash
 python -m pytest theseus
 ```
-By default, unit tests include tests for our CUDA extensions. You can add the option `-m "not cudaext"` to skip them when installing without CUDA support.
+By default, unit tests include tests for our CUDA extensions, and for BaSpaCho sparse solver. You can add the option `-m "not cudaext"`
+to skip them when installing without CUDA support, or `-m "not baspacho"` installing without BaSpaCho (or
+`-m "(not cudaext) and (not baspacho)"` to disable both).
```
mhmukadam, line 121: Should we have the cpu baspacho alwasy included? Then we can subsume cuda baspacho under all cuda stuff with cudaext and avoid having multiple flags.

setup.py
```diff
@@ -25,6 +38,50 @@ def parse_requirements_file(path):
...
+        return None
+    ext_cls = torch_cpp_ext.CUDAExtension if has_cuda else torch_cpp_ext.CppExtension
+    sources = ["theseus/extlib/baspacho_solver.cpp"]
+    define_macros = [("NO_BASPACHO_CHECKS", "1")]
+    if has_cuda:
+        sources.append("theseus/extlib/baspacho_solver_cuda.cu")
+        define_macros.append(("THESEUS_HAVE_CUDA", "1"))
+        # extra_compile_args = {"cxx": ["-std=c++17"]}  # , "nvcc": ["-std=c++17"]}
+        # else:
+        # extra_compile_args = ["-std=c++17"]
```
mhmukadam, line 70: remove commented stuff?

theseus/extlib/baspacho_solver.cpp
```diff
@@ -0,0 +1,377 @@
+
```
mhmukadam, line 1: Add license header to all new files.

theseus/extlib/baspacho_solver_cuda.cu
```diff
@@ -0,0 +1,305 @@
...
+        factorPtrs.push_back(pFactor + factorSize * i);
+        xPtrs.push_back(pPermX + order * i);
+    }
+    dec->solver->solve(&factorPtrs, &xPtrs, order, 1);
+
+    // un-scramble
+    unscramble_kernel<<<numBlocks, wgs>>>(accessor, pParamSize, pParamStart,
+                                   order, pX, pPermX, nParams, batchSize);
+    cuCHECK(cudaDeviceSynchronize());
+}
```
mhmukadam, line 305: Missing new line

theseus/extlib/baspacho_solver.h
```diff
@@ -0,0 +1,68 @@
...
+#ifdef THESEUS_HAVE_CUDA
+    bool isCuda;
+#endif
+    torch::Tensor toParamIndex;
+    torch::Tensor paramStart;
+    torch::Tensor paramSize;
+};
+
+// Numeric decomposition, CPU version
+class NumericDecomposition {
```
mhmukadam, line 19: Maybe helpful to leave a brief comment next to all class/functions about what they do.

theseus/extlib/cusolver_lu_solver.cpp
```diff
@@ -88,21 +88,21 @@ CusolverLUSolver::CusolverLUSolver(int batchSize,
     const int *pA_rowPtr_cpu = A_rowPtr_cpu.data_ptr<int>();
     const int *pA_colInd_cpu = A_colInd_cpu.data_ptr<int>();
 
-    // we compute the permutation Q which allows 
+    // we compute the permutation Q which allows
```
mhmukadam, line 91: Seems like all changes to this file are just adding extra space to a few lines. If this is not automatic from a linter then we should remove this file from the PR.

theseus/extlib/tests/test_baspacho.py
```diff
@@ -0,0 +1,179 @@
...
+@pytest.mark.baspacho
+def test_baspacho_cpu_0():
+    torch.manual_seed(0)
+    check_baspacho(batch_size=2, num_rows=20, num_cols=10, fill=0.3)
+
+
+@pytest.mark.baspacho
+def test_baspacho_cpu_1():
+    torch.manual_seed(1)
+    check_baspacho(batch_size=5, num_rows=50, num_cols=30, fill=0.2)
```
mhmukadam, line 102: These and other test below can be streamlined into a single function and the settings can be passed as options like in other unit tests with `@pytest.mark.parametrize`.

mhmukadam, line 102: These and other test below can be streamlined into a single function and the settings can be passed as options like in other unit tests with `@pytest.mark.parametrize`.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
```diff
@@ -0,0 +1,199 @@
...
+
+import torch
+
+from ..linear_system import SparseStructure
+
+from scipy.sparse import csr_matrix, csc_matrix
+import numpy as np
+
+
+def mat_vec_cpu(batch_size, num_cols, A_rowPtr, A_colInd, A_val, v):
```
mhmukadam, line 14: I just realized that many files like these in `theseus/optimizer/` do not have type hints (input types, return type) included when writing functions.

mhmukadam (commented):
Look good, awesome work @maurimo! Left some minor comments and one about streamlining installation.

README.md
maurimo: it is included in tests by default, indeed, and that's why I added a marker to run tests not depending on it.

README.md
maurimo: Not entirely sure about this - there are some points which are a bit delicate, in particular how to compile baspacho so that it takes advantage of system BLAS if available, and reverts to a fallback if not available. Also, it may fail if CMake is not recent enough, and if GCC/NVCC are not compatible versions. So I would consider reasonable to leave the user control over how it's installed, for the time being. Maybe we can start like this and then when confident that there is no installation issues we make setup.py install it automatically (it would be at least a bit of work required to achieve this too I think  right? @luisenp )

theseus/extlib/cusolver_lu_solver.cpp
maurimo: actually I was cleaning extraspaces, and ended up doing so with a few extra files.

theseus/extlib/baspacho_solver.h
maurimo: +1.

theseus/extlib/baspacho_solver_cuda.cu
maurimo: +1.

theseus/extlib/baspacho_solver.cpp
maurimo: +1.

setup.py
maurimo: +1.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
maurimo: +1.

evaluations/pose_graph_synthetic.sh
```diff
@@ -6,11 +6,14 @@
...
        inner_optim.solver=sparse \
        loop_closure_ratio=0.2 \
        solver_device=cuda \
-       dataset_size=256 \
-       batch_size=128 \
-       num_poses=256
+       dataset_size=32 \
+       batch_size=32 \
+       num_poses=4096 \
+       solver_type=baspacho
```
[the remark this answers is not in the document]
luisenp: I don't think there is some clear default for these settings, except perhaps using one that's likely to work for any  user (e.g., small batch/problem size, and CHOLMOD).

setup.cfg
```diff
@@ -6,6 +6,7 @@ max-line-length = 100
 ignore=E203, W503
 per-file-ignores =
     *__init__.py:F401
+    theseus/extlib/tests/test_baspacho_simple.py:E261,E241,E121,E131,E201,E202
```
luisenp, line 9: Can you extend the comment above (lines 3-5) with the errors that these codes are referring to?

setup.py
luisenp: +2. Also in line 78.

setup.py
```diff
@@ -37,8 +94,21 @@ def parse_requirements_file(path):
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
-if torch.cuda.is_available():
+# Add C++ and CUDA extensions
+compile_cuda_flag = os.environ.get("THESEUS_ENABLE_CUDA")
```
luisenp, line 98: Following up on the discussion in #304, is there a reason why the currently named `THESEUS_ENABLE_CUDA` shouldn't be the default behavior? Does it interact weirdly with `pip install -e .` for non-CUDA machines?

Otherwise, why not make this default, remove the flag completely, and just add another flag that disables all extensions if needed?

theseus/extlib/tests/test_baspacho.py
```diff
@@ -0,0 +1,179 @@
...
+
+import numpy as np
+import pytest  # noqa: F401
+import torch  # needed for import of Torch C++ extensions to work
+from scipy.sparse import csr_matrix, tril
+
+from theseus.utils import random_sparse_binary_matrix
+
+
+# ideally we would like to support batch_size <= init_batch_size, but
```
luisenp, line 14: Is this also true for Baspacho or is it remains of a copy paste from LUCuda?

theseus/extlib/tests/test_baspacho.py
luisenp: Agreed. [This](https://github.com/facebookresearch/theseus/blob/main/theseus/tests/test_theseus_layer.py#L377) might be a useful example.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
luisenp: Good catch. Let's make an issue for this. Definitely not to be solved in this PR.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
```diff
@@ -0,0 +1,199 @@
...
+
+        A_args = ctx.sparse_structure.num_cols, ctx.A_rowPtr, ctx.A_colInd, ctx.A_val
+        AH = mat_vec(batch_size, *A_args, H)
+        b_Ax = ctx.b - mat_vec(batch_size, *A_args, ctx.x)
+
+        # now we fill values of a matrix with structure identical to A with
+        # selected entries from the difference of tensor products:
+        #   b_Ax (X) H - AH (X) x
+        # NOTE: this row-wise manipulation can be much faster in C++ or Cython
+        A_colInd = ctx.sparse_structure.col_ind
```
luisenp, line 179: From a quick look the code below this line seems to be equal to the one in `lu_cuda_sparse_autograd.py`. If so, can we encapsulate this into a function that both of these call?

theseus/optimizer/linear/baspacho_sparse_solver.py
```diff
@@ -0,0 +1,123 @@
...
+
+        self._num_solver_contexts: int = num_solver_contexts
+
+        if self.linearization.structure().num_rows:
+            self.reset(dev)
+
+    def reset(self, dev="cpu"):
+        if dev == "cuda" and not torch.cuda.is_available():
+            raise RuntimeError(
+                "Cuda not available, BaspachoSparseSolver cannot be used"
```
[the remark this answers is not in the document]
luisenp: Something like "The CUDA version of BaspachoSparseSolver cannot be used."?

README.md
luisenp: I think @mhmukadam is making a different point, which is to have only the one flag for CUDA (which would include LUCuda, but also CUDA+Baspacho), and then **always** install CPU-Baspacho, so that we don't need the other flag at all. Is this correct?

I'm not sure about this change because BASPACHO support requires a manual installation of a dependency (the core BASPACHO), so it seems a bit too much user work for it to be considered the default behavior. 

I would actually propose to remove the "baspacho" marker, and just change unit tests to `try` catch the Baspacho import, and only emit a warning that Baspacho is not installed, instead of crashing.

luisenp (changes requested):
Awesome work @maurimo, looking forward to merging this soon! I took an initial pass and left some comments, let me know if you have any questions.

theseus/extlib/tests/test_baspacho.py
mhmukadam: Oh yeah good catch. I thought this applied here.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
mhmukadam: I think we may have an issue for this already?

README.md
mhmukadam: Yes, one flag for CUDA is what I meant.

theseus/optimizer/linear/baspacho_sparse_solver.py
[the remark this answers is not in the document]
maurimo: yeah will improve the error message!

setup.cfg
maurimo: just the fact that the simple test has matrices in the python code that I formatted in order to make the sparse pattern visible, and therefore and I don't want to be reformatted (or getting error for the format being edited manually). Will comment.

setup.py
maurimo: I would leave `THESEUS_ENABLE_CUDA`, no weird interaction AFAIK, it just overrides the result of `torch.cuda.is_available()`

theseus/optimizer/autograd/baspacho_sparse_autograd.py
maurimo: Yes, definitely to be addressed. only objection is that the code has several similar parts but also subtle differences, so unifying and making it generic might end up being harder said than done. (For instance, data saved in the ctx are quite different for all the autograd classes.) If ok I propose solving this in a future PR, as it might require iterating several times before I get it right!

README.md
maurimo: Not sure if I'm understanding. If BaSpaCho is installed, there is just one flag indeed `-m cudaext` runs Cuda tests (including the BaSpaCho ones), and `-m "not cudaext` runs the CPU tests (including BaSpaCho ones).
Now, if you decide to install without BaSpaCho (which is currently still possible) you might want to run tests skipping the BaSpaCho ones, and `-m "not baspacho"` allows you to do exactly that. If BaSpaCho is installed, the only relevant marker is `cudaext`.

setup.py
luisenp: Thanks @maurimo for clarifying offline. Current proposal is to rename as `THESEUS_FORCE_CUDA`.

README.md
maurimo: renamed.

evaluations/pose_graph_synthetic.sh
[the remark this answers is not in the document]
maurimo: So, values here are just a set of example for the values iterated on in the eval (and explained above in the comments). marking as resolved.

setup.py
maurimo: cleaned.

theseus/extlib/baspacho_solver.cpp
maurimo: done

theseus/extlib/baspacho_solver_cuda.cu
maurimo: done

theseus/extlib/baspacho_solver.h
maurimo: done!

theseus/optimizer/linear/baspacho_sparse_solver.py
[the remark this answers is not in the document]
maurimo: improved.

setup.cfg
maurimo: added comment to explain.

setup.py
maurimo: renamed to THESEUS_FORCE_CUDA.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
maurimo: Ok to resolve? Then getting batch to this in a future PR. Doesn't look like the best thing to do delaying this PR to solve an issue already present in the code...

theseus/extlib/tests/test_baspacho.py
luisenp: Remove comment then?

theseus/optimizer/autograd/baspacho_sparse_autograd.py
luisenp: Yes, OK to resolve.

README.md
maurimo: done, now tests are skipped with a "SKIPPED" message if `hasattr(theseus.extlib, "baspacho_solver")` fails.

theseus/extlib/tests/test_baspacho.py
maurimo: cleared.

theseus/extlib/tests/test_baspacho.py
maurimo: done, got to really like `parametrize` and now a couple hundreds sparse matrices are generated for thorough testing. Looking forward to do the same for the solver and autograd classes. Tell me if I went too far :-p

maurimo:
Addressed all comments, let's see what the CI says. Might still parametrize solver class and autograd tests for more thorough testing.

maurimo:
parametrize solver and autograd tests too. autograd is tested with several possibilities but not as many, as it's pretty slow (because of computation of numerical derivatives for all involved params).

theseus/extlib/tests/test_baspacho.py
mhmukadam: Nice. Maybe 100s is overkill and might slow down CI?

evaluations/pose_graph_synthetic.sh
mhmukadam, line 19: Like @luisenp mentioned we should swap to a more reasonable default that will always work out of the box?

mhmukadam (commented):
All updates look good!

Just two comments to resolve: [this](https://github.com/facebookresearch/theseus/pull/324#discussion_r997155253) and [this](https://github.com/facebookresearch/theseus/pull/324#discussion_r997214315).

Once the install instructions are streamlined we should be ready to merge.

theseus/extlib/tests/test_baspacho.py
luisenp: We can give it a shot and adjust later if it's too slow?

evaluations/pose_graph_synthetic.sh
maurimo: oh sounds good.

maurimo:
added support for CPU-only baspacho in build_wheel.sh. I don't have time to test now but should work, the option to be tweaked is BASPACHO_CUDA_OPTIONS variable in build_wheel.sh.

evaluations/pose_graph_synthetic.sh
```diff
@@ -16,4 +16,4 @@ python examples/pose_graph/pose_graph_synthetic.py \
        dataset_size=32 \
        batch_size=32 \
        num_poses=4096 \
-       solver_type=baspacho
+       solver_type=lucuda
```
mhmukadam, line 19: hmm maybe cholmod is a better default since it works on both CPU and GPU?

mhmukadam (approved):
Looks good! Once we merge #330 and all CI pass, we should ask someone from the team to try installs from wheels and source (CPU only and CPU+GPU). If everything goes smoothly then we are ready to merge.

evaluations/pose_graph_synthetic.sh
maurimo: when solver_device=cpu then the solver_type is ignored. If I add baspacho as a cpu option it will not (currently) work because of the extra parameter dev="cpu" needed to the solver class, not supported by default! let's not forget this is an example script which was already there...

theseus/extlib/tests/common.py
```diff
@@ -0,0 +1,17 @@
+_BASPACHO_NOT_INSTALLED_MSG = "Baspacho solver not in theseus extension library."
```
mhmukadam, line 1: Add license header here and to the two new empty init files in extlib.

mhmukadam (approved):

luisenp (approved):
Ready to merge!

PR #328, opened 2022-10-13
Adaptive damping for Levenberg-Marquardt

luisenp:
This PR adds adaptive damping for LM, based on Section 4 of [this](https://people.duke.edu/~hpgavin/ce281/lm.pdf) reference. It includes a number of features:

- Allow independent step sizes for each batch element (not used in this PR).
- Include a `reset()` method for nonlinear optimizers that's called at the start of each optimization. We could also have it call the linear solver's reset method, although in some of the current sparse solvers this is costly (@maurimo to confirm).
- Include a `update_optimizer_step()` method for nonlinear optimizers. 
- Add an implementation of the above for LM to allow adaptive damping.

This feature is currently only compatible with dense solvers. @maurimo how likely is that we could add support for this in the sparse solvers?

As an added bonus of this PR, once merged it should be easier to implement dogleg, learnable damping, and more sophisticated line search.

### DIFF
--- a/theseus/extlib/mat_mult.cu
+++ b/theseus/extlib/mat_mult.cu
@@ -14,6 +14,8 @@
 #include <ATen/cuda/detail/DeviceThreadHandles.h>
 #include <ATen/cuda/CUDAContext.h>
 
+#include "utils.h"
+
 __device__ int bisect_index(const int* values, int len, int needle) {
 
 	int a = 0, b = len;
@@ -86,16 +88,10 @@ torch::Tensor mult_MtM(int batchSize,
 	int64_t M_numRows = M_rowPtr.size(0) - 1;
 	int64_t M_nnz = M_colInd.size(0);
 
-	TORCH_CHECK(M_rowPtr.device().is_cuda());
-	TORCH_CHECK(M_colInd.device().is_cuda());
-	TORCH_CHECK(Ms_val.device().is_cuda());
-	TORCH_CHECK(M_rowPtr.dtype() == torch::kInt);
-	TORCH_CHECK(M_colInd.dtype() == torch::kInt);
-	TORCH_CHECK(Ms_val.dtype() == torch::kDouble); // TODO: add support for float
-	TORCH_CHECK(M_rowPtr.dim() == 1);
-	TORCH_CHECK(M_colInd.dim() == 1);
-	TORCH_CHECK(Ms_val.dim() == 2);
-	TORCH_CHECK(Ms_val.size(0) == batchSize);
+	THESEUS_TENSOR_CHECK_CUDA(M_rowPtr, 1, M_rowPtr.size(0), torch::kInt);
+	THESEUS_TENSOR_CHECK_CUDA(M_colInd, 1, M_colInd.size(0), torch::kInt);
+	// TODO: add support for float
+	THESEUS_TENSOR_CHECK_CUDA(Ms_val, 2, batchSize, torch::kDouble);
 	TORCH_CHECK(Ms_val.size(1) == M_nnz);
 
 	int64_t MtM_numRows = MtM_rowPtr.size(0) - 1;
@@ -177,20 +173,13 @@ torch::Tensor mat_vec(int batchSize,
 	int64_t M_nnz = M_colInd.size(0);
 
 	TORCH_CHECK(M_rowPtr.device().is_cuda());
-	TORCH_CHECK(M_colInd.device().is_cuda());
-	TORCH_CHECK(Ms_val.device().is_cuda());
-	TORCH_CHECK(M_rowPtr.dtype() == torch::kInt || M_rowPtr.dtype() == torch::kInt64);
-	TORCH_CHECK(M_colInd.dtype() == M_rowPtr.dtype());
-	TORCH_CHECK(Ms_val.dtype() == torch::kDouble); // TODO: add support for float
 	TORCH_CHECK(M_rowPtr.dim() == 1);
-	TORCH_CHECK(M_colInd.dim() == 1);
-	TORCH_CHECK(Ms_val.dim() == 2);
-	TORCH_CHECK(Ms_val.size(0) == batchSize);
+	TORCH_CHECK(M_rowPtr.dtype() == torch::kInt || M_rowPtr.dtype() == torch::kInt64);
+	THESEUS_TENSOR_CHECK_CUDA(M_colInd, 1, M_colInd.size(0), M_rowPtr.dtype());
+	// TODO: add support for float
+	THESEUS_TENSOR_CHECK_CUDA(Ms_val, 2, batchSize, torch::kDouble);
 	TORCH_CHECK(Ms_val.size(1) == M_nnz);
-	TORCH_CHECK(vec.device().is_cuda());
-	TORCH_CHECK(vec.dim() == 2);
-	TORCH_CHECK(vec.size(0) == batchSize);
-	TORCH_CHECK(vec.size(1) == M_numCols);
+	THESEUS_TENSOR_CHECK_CUDA(vec, 2, batchSize, vec.dtype());
 	
 	auto xOptions = torch::TensorOptions().dtype(torch::kDouble).device(Ms_val.device());
 	torch::Tensor retv = torch::empty({(long)batchSize, (long)M_numRows}, xOptions);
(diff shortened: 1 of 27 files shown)


theseus/optimizer/linear/lu_cuda_sparse_solver.py
```diff
@@ -90,17 +90,19 @@ def reset(self, batch_size: int = 16):
...
     ) -> torch.Tensor:
+        if damping is not None and not isinstance(damping, float):
+            raise ValueError("LUCudaSparseSolver only supports scalar damping.")
         if self._auto_reset:
             if self._solver_contexts[0].batch_size != self._objective.batch_size:
                 self.reset(self._objective.batch_size)
         if not isinstance(self.linearization, SparseLinearization):
             raise RuntimeError(
-                "CholmodSparseSolver only works with theseus.optimizer.SparseLinearization."
+                "LUCudaSparseSolver only works with theseus.optimizer.SparseLinearization."
```
mhmukadam, line 105: Good catch!

theseus/optimizer/linear/tests/test_dense_solver.py
```diff
@@ -47,9 +52,10 @@ def test_dense_solvers():
         # Add damping (using loop to make it more visually explicit)
         for i in range(batch_size):
             for j in range(n):
-                AtA_ellipsoidal_damp[i, j, j] *= 1 + damping
+                damp = damping if isinstance(damping, float) else damping[i]
```
mhmukadam, line 55: In the case when damping is scalar/float would it make sense to do a `as_tensor` on it and then allow it to be broadcasted? Would that make it more complicated than simplifying things?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -460,3 +467,22 @@ def _optimize_impl(
     @abc.abstractmethod
     def compute_delta(self, **kwargs) -> torch.Tensor:
         pass
+
+    # Resets any internal state needed by the optimizer for a new optimization
+    # problem. Optimizer loop will pass all optimizer kwargs to this method.
+    def reset(self, **kwargs) -> None:
```
mhmukadam, line 473: Should this be an abstract method?

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -2,23 +2,42 @@
...
+    LUCudaSparseSolver,
+    LUDenseSolver,
+)
 
 from .nonlinear_least_squares import NonlinearLeastSquares
 
-_LM_ALLOWED_ELLIPS_DAMP_SOLVERS = [DenseSolver, LUCudaSparseSolver]
+_LM_ALLOWED_ELLIPS_DAMP_SOLVERS: List[Type[LinearSolver]] = [
+    CholeskyDenseSolver,
+    LUDenseSolver,
```
mhmukadam, line 23: Since all `DenseSolver` in general are supported is it necessary to qualify the two current dense solvers explicitly?

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -54,31 +73,87 @@ def __init__(
...
+        damping = (
+            self._damping.view(-1, 1)
+            if isinstance(self._damping, torch.Tensor)
+            else self._damping
+        )
+        den = (delta * (damping * delta + linearization.Atb.squeeze(2))).sum(dim=1)
+        rho = (last_err - new_err) / den
+        good_idx = rho > self._ACCEPT_EPS
+        self._damping = torch.where(good_idx, self._damping / 9, self._damping * 11)
+        self._damping = self._damping.clamp(1.0e-7, 1.0e7)
```
mhmukadam, line 159: Are 9, 11, 1e-7, 1e7 fixed numbers of the algorithm or hyperparameters? If latter then maybe we shouldn't hardcode them.

mhmukadam (approved):
Looks great, nice work! Also love that this sets us up for other items on the roadmap.

If the addition to sparse solvers involves minor changes then maybe we should do it in this PR so the feature is merged with complete support.

theseus/optimizer/nonlinear/levenberg_marquardt.py
mhmukadam: Ah looking from the doc they are fixed numbers in the algorithm for 'good' convergence.

luisenp: Yeah, but also not a bad idea to make them named constants instead of magic numbers.

theseus/optimizer/nonlinear/levenberg_marquardt.py
luisenp: This was to avoid some `mypy` annoyance, because apparently `Type` is meant to be used only for non-abstract  classes. I found multiple `mypy` issues complaining about this limitation, and some workarounds that didn't seem to quite apply to our case.

theseus/optimizer/linear/tests/test_dense_solver.py
luisenp: Do you mean for the unit test?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: I prefer not, since I imagine there might be solvers where this is not needed (this is indeed the current case for `GaussNewton`), then we are forcing the class to unnecessarily implement a stub just to comply with the base class.

maurimo:
Yes LUCuda's reset is expected to be a bit slow. This can be easily implemented in all sparse solvers (without having to really recreate the sparse solver) apart from Cholmod, where we use the builting At*A+beta and the damping option is just adding a constant value.

maurimo:
If anyone is keen to implement this for sparse solver, this is the function that has to be extended for LuCuda:
https://github.com/facebookresearch/theseus/blob/main/theseus/extlib/mat_mult.cu#L317
and here (CPU/CUDA) for the upcoming sparse solver BaSpaCho (taking care that a permutation was applied in the factor, and have to do the right thing):
https://github.com/facebookresearch/theseus/blob/mau.baspacho_revamp/theseus/extlib/baspacho_solver.cpp#L178
https://github.com/facebookresearch/theseus/blob/mau.baspacho_revamp/theseus/extlib/baspacho_solver_cuda.cu#L202

theseus/optimizer/linear/tests/test_dense_solver.py
mhmukadam: I meant in the main implementation (I just realized I flagged this from a test file).

theseus/optimizer/nonlinear/levenberg_marquardt.py
mhmukadam: Agreed. Then we can easily edit them in the future if needed. Potentially add them to constants.py.

rmurai0610: I think we should make these configurable parameters as some applications uses different values. 
For example GTSAM and ceres uses different parameters (as ceres is tuned more for SfM)
https://github.com/borglab/gtsam/blob/develop/gtsam/nonlinear/LevenbergMarquardtParams.h#L85

mhmukadam: Good point @rmurai0610, then we need these to be kwargs rather than in constants.py.

theseus/optimizer/linear/tests/test_dense_solver.py
luisenp: Changed it in the main implementation. Prefer to leave as is in the test, to make sure passing a float also works.

theseus/optimizer/nonlinear/levenberg_marquardt.py
luisenp: Good idea. Added!

mhmukadam (commented):
LGTM so far. Based on offline discussion we'll wait to add support for sparse solvers as well before merging.

theseus/extlib/mat_mult.cu
```diff
@@ -264,19 +253,13 @@ torch::Tensor tmat_vec(int batchSize,
...
-	TORCH_CHECK(Ms_val.device().is_cuda());
 	TORCH_CHECK(M_rowPtr.dtype() == torch::kInt || M_rowPtr.dtype() == torch::kInt64);
-	TORCH_CHECK(M_colInd.dtype() == M_rowPtr.dtype());
-	TORCH_CHECK(Ms_val.dtype() == torch::kDouble); // TODO: add support for float
 	TORCH_CHECK(M_rowPtr.dim() == 1);
-	TORCH_CHECK(M_colInd.dim() == 1);
-	TORCH_CHECK(Ms_val.dim() == 2);
-	TORCH_CHECK(Ms_val.size(0) == batchSize);
+	THESEUS_TENSOR_CHECK_CUDA(M_colInd, 1, M_colInd.size(0), M_rowPtr.dtype());
+	// TODO: add support for float
```
mhmukadam, line 259: raise issue? comment could get lost

mhmukadam (approved):

theseus/extlib/mat_mult.cu
luisenp: Created #348

PR #330, opened 2022-10-18
Cleaned up Baspacho install, tests and README

luisenp:
Main changes:

* Changed build script so that it always does the equivalent of `THESEUS_FORCE_CUDA=1`. This argument no longer needs to be passed.
* Hardcoded CUDA archs in the build script for. I'm guessing this  might be controversial, but couldn't get this to compile otherwise.
* Fixed build script so that it works with CPU-only install.
* Removed pytest baspacho marker. I don't think it's needed because tests are skipped automatically if BaSpaCho is not found.
* Added clearer README instructions. 
* Added a preliminary version bump (0.1.2rc1) as the build script requires a tag to work properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Our implementation provides an easy to use interface to build custom optimizatio
 - [Second-order nonlinear optimizers](https://github.com/facebookresearch/theseus/tree/main/theseus/optimizer/nonlinear)
     - Gauss-Newton, Levenberg–Marquardt
 - [Linear solvers](https://github.com/facebookresearch/theseus/tree/main/theseus/optimizer/linear)
-    - Dense: Cholesky, LU; Sparse: CHOLMOD, LU
+    - Dense: Cholesky, LU; Sparse: CHOLMOD, LU, [BaSpaCho](https://github.com/facebookresearch/baspacho)
 - [Commonly used costs](https://github.com/facebookresearch/theseus/tree/main/theseus/embodied), [AutoDiffCostFunction](https://github.com/facebookresearch/theseus/blob/main/theseus/core/cost_function.py), [RobustCostFunction](https://github.com/facebookresearch/theseus/blob/main/theseus/core/robust_cost_function.py)
 - [Lie groups](https://github.com/facebookresearch/theseus/tree/main/theseus/geometry)
 - [Robot kinematics](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/kinematics/kinematics_model.py)
@@ -97,6 +97,7 @@ For other CUDA versions, consider installing from source or using our
 [build script](https://github.com/facebookresearch/theseus/blob/main/build_scripts/build_wheel.sh).
 
 #### **From source**
+The simplest way to install Theseus from source is by running
 ```bash
 git clone https://github.com/facebookresearch/theseus.git && cd theseus
 pip install -e .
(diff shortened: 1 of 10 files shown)


build_scripts/build_wheel.sh
```diff
@@ -119,24 +111,19 @@ for PYTHON_VERSION in 3.9; do
     echo $(pwd)
     DOCKER_NAME=theseus_${PYTHON_VERSION}
     sudo docker build -t "${DOCKER_NAME}_img" .
-    sudo docker run ${GPU_ARGS} --name ${DOCKER_NAME} ${DOCKER_NAME}_img
+    echo "sudo docker run --name ${DOCKER_NAME} ${DOCKER_NAME}_img"
```
maurimo, line 114: probably debug message that can be cleaned up

maurimo (changes requested):
All looks great, will merge as soon as the CI succeeds!

theseus/extlib/tests/test_baspacho_simple.py
```diff
@@ -121,14 +112,11 @@ def check_simple(verbose=False, dev="cpu"):
...
 
 
-@pytest.mark.baspacho
-@requires_baspacho
+@run_if_baspacho()
 def test_simple_cpu():
     check_simple(dev="cpu")
 
 
-@pytest.mark.cudaext
```
maurimo, line 130: found! restore the cudaext mask and CI will succeed ;)

luisenp: Added, thanks!

build_scripts/build_wheel.sh
luisenp: Removed.

luisenp (commented):
Fixed the comments and rebased. I also changed the batch sizes to [1, 32] to be close to what we are doing for other unit tests.

theseus/constants.py
```diff
@@ -54,3 +54,21 @@ def __getitem__(self, dtype):
 _SE3_NEAR_ZERO_EPS = EPSDict(float32_eps=1e-2, float64_eps=5e-3)
 
 _SE3_HAT_EPS = EPSDict(float32_eps=5e-6, float64_eps=5e-7)
+
+_BASPACHO_NOT_INSTALLED_MSG = "Baspacho solver not in theseus extension library."
+
+
+def run_if_baspacho():
+    # Not sure what's a better place to put this in
```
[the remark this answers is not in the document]
mhmukadam: Or extlib/tests/common.py ?

README.md
```diff
@@ -97,6 +97,7 @@ For other CUDA versions, consider installing from source or using our
 [build script](https://github.com/facebookresearch/theseus/blob/main/build_scripts/build_wheel.sh).
 
 #### **From source**
+The simplest way to install Theseus from source is by running
```
mhmukadam, line 100: could also add:

... from source is by running (see below to also include BaSpaCho)

README.md
```diff
@@ -107,19 +108,27 @@ pip install -e ".[dev]"
 ```
 and follow the more detailed instructions in [CONTRIBUTING](https://github.com/facebookresearch/theseus/blob/main/CONTRIBUTING.md).
 
-Compilation from source supports is affected by the following enviroment variables:
-* `BASPACHO_ROOT_DIR`: root dir of [BaSpaCho](https://github.com/facebookresearch/baspacho) sparse solver, which must be already
-  built with binaries in the subdirectory `build`.
-* `THESEUS_FORCE_CUDA` (1 or 0): if set, forces enabling or disabling Cuda support regardless of system's `torch.cuda.is_available()`.
+**Installing BaSpaCho extensions from source**
```
mhmukadam, line 111: For readability I would suggest making the following inner headings as bullets

- #### **pypi**
- #### **From source**
- #### **Installing BaSpaCho extensions from source**

mhmukadam (approved):
This is great! Nice work @luisenp and @maurimo on streamlining the installation and reducing complexity with env vars and flags. A lot more intuitive now.

README.md
```diff
@@ -62,7 +62,7 @@ Our implementation provides an easy to use interface to build custom optimizatio
 - [Second-order nonlinear optimizers](https://github.com/facebookresearch/theseus/tree/main/theseus/optimizer/nonlinear)
     - Gauss-Newton, Levenberg–Marquardt
 - [Linear solvers](https://github.com/facebookresearch/theseus/tree/main/theseus/optimizer/linear)
-    - Dense: Cholesky, LU; Sparse: CHOLMOD, LU
+    - Dense: Cholesky, LU; Sparse: CHOLMOD, LU, [BaSpaCho](https://github.com/facebookresearch/baspacho)
```
mhmukadam, line 65: Should we add next to Sparsse LU (GPU only), so it's clear all others support both CPU and GPU?

PR #332, opened 2022-10-20
Deprecated FULL backward mode (now UNROLL).

luisenp:


examples/backward_modes.py
```diff
@@ -4,7 +4,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 #
-# This example illustrates the three backward modes (FULL, IMPLICIT, and TRUNCATED)
+# This example illustrates the four backward modes
+# (UNROLL, IMPLICIT, TRUNCATED, and DLM)
```
mhmukadam, line 8: This is the only file with all caps for the modes. Should we makes these all lower case to be consistent with the rest?

mhmukadam (approved):

PR #336, opened 2022-10-22
Small vectorization improvements

luisenp:
This PR commits two changes:

- Don't expand tensors in vectorization if the objective's batch size = 1
- Retract all variables now updates vectorization immediately, which avoids some cases were vectorization had to be run twice with the same data (because of torch.no_grad() uses). 

TODO: check if this changes times in the PGO benchmark. In a small BA script it looks like this leads to some improvements in running time (~10-20%).

### DIFF
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -509,9 +509,7 @@ class Objective:
         self._batch_size = _get_batch_size(batch_sizes)
 
     def _vectorization_needs_update(self):
-        num_updates = dict(
-            (name, v._num_updates) for name, v in self._all_variables.items()
-        )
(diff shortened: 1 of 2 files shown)


theseus/core/objective.py
```diff
@@ -575,6 +573,11 @@ def retract_optim_vars(
         self._retract_method(
             delta, ordering, ignore_mask=ignore_mask, force_update=force_update
         )
+        # Updating immediately is useful, since it will keep grad history if
+        # needed. Otherwise, with lazy waitng we can be in a situation were
```
mhmukadam, line 577: were -> where

mhmukadam (approved):

PR #337, opened 2022-10-22
Fixed vmap related bug breaking homography ex. with sparse solvers.

luisenp:
There was an error that was causing failures in code using AutodiffCostFunction with "vmap" in some circumstances. For example, if one wished to move tensors to another device after having run vmap at least once. 

This was causing the homography estimation example to break when using sparse solvers and vectorization at the same time.

I also added some code to easily change the linear solver used in the homography example.


theseus/core/cost_function.py
```diff
@@ -263,11 +272,20 @@ def _expand_all(
...
-            )
+            # functorch doesn't allow BatchedTensors created inside their transforms
+            # to be accessed outside that context. Since our tmp containers get
+            # populated inside the vmap(jacrev) calculation, this would result in
+            # errors when calling functions like copy() or to() after
+            # vmap has been run at least once.
+            # The _tmp_tensors context managers stores the (regular) tensors the vars
+            # had before entering and restores them on exit, thus dereferencing
+            # the temporary BatchedTensors.
+            with _tmp_tensors(self._tmp_optim_vars), _tmp_tensors(self._tmp_aux_vars):
```
mhmukadam, line 283: This is a nice fix! There is likely very little overhead I am assuming?

mhmukadam (approved):

theseus/core/cost_function.py
fantaosha: So functorch essentially changes the types of `self._tmp_optim_vars` and `self._tmp_aux_vars`?

fantaosha (approved):
LGTM.

theseus/core/cost_function.py
luisenp: Looks like functorch's vmap replaces all tensor inputs with copies of a class called `BatchedTensor` and returns tensors of this type that get converted to regular tensors when the function returns. If you have some global reference to a `BatchedTensor` created inside but never returned (e.g., by storing in `self._tmp_optim_vars`), then it never gets converted to a regular tensor and we get an error if we try to access it outside.

luisenp: @mhmukadam I would assume that the overhead of this is not too large compared to the costly operation will happen inside, but I can run some checks just in case.

fantaosha: > Looks like functorch's vmap replaces all tensor inputs with copies of a class called `BatchedTensor` and returns tensors of this type that get converted to regular tensors when the function returns. If you have some global reference to a `BatchedTensor` created inside but never returned (e.g., by storing in `self._tmp_optim_vars`), then it never gets converted to a regular tensor and we get an error if we try to access it outside.

Great catch! Not sure if this will raise other bugs since the reference types are essentially changed. BTW, what is the type of `jacobian`? Is it a regular `Tensor`?

luisenp: Yeah, `jacobian` should be a regular tensor, since we get it as the return value from `vmap`. 

Looking at the code below, the only potential source of errors I can see is if `_err_fn` stores internal data during the error computation. We should maybe document that `err_fn` should be a [pure function](https://en.wikipedia.org/wiki/Pure_function). 

image

luisenp: I'll add a comment at the top of `AutoDiffCostFunction`. Also, my last statement wasn't correct. I think it's fine if the function uses some outside information to compute error (e.g., some model that gets updated periodically), but what's not allowed is that the function stores the tensors used as input to be reused elsewhere.

luisenp: Added comment. Should I merge? I ran some small tests locally and I didn't notice any overhead.

mhmukadam (approved):

PR #340, opened 2022-10-31
Add tensor damping support to sparse solvers

luisenp:


theseus/extlib/baspacho_solver.cpp
```diff
@@ -175,13 +175,23 @@ void NumericDecomposition::add_MtM(const torch::Tensor& val,
...
 #ifdef THESEUS_HAVE_CUDA
     if (dec->isCuda) {
-        damp_cuda(alpha, beta);
+        TORCH_CHECK(alpha.device().is_cuda());
+        TORCH_CHECK(beta.device().is_cuda());
+        damp_cuda(alpha.data_ptr<double>(), beta.data_ptr<double>());
         return;
     }
 #endif
+    TORCH_CHECK(alpha.device().is_cpu());
```
mhmukadam, line 194: No similar beta cpu check?

theseus/extlib/tests/test_cusolver_lu_solver.py
```diff
@@ -23,8 +23,10 @@ def check_lu_solver(
         return
     from theseus.extlib.cusolver_lu_solver import CusolverLUSolver
 
+    rng = torch.Generator()
+    rng.manual_seed(0)
```
mhmukadam, line 27: Any reason the manual_seed is hash based from params in baspacho test but set to default in lu test?

theseus/extlib/baspacho_solver.cpp
luisenp: Added some macros to clean up all the tensor checks. See 919b8783b030a0be89e4033966b855440482c102.

theseus/extlib/tests/test_cusolver_lu_solver.py
luisenp: We could do the same thing here. I was just maintaining consistency with the original code. I'm guessing @maurimo hadn't thought of the hash version before baspacho.

theseus/extlib/baspacho_solver.h
```diff
@@ -57,7 +79,7 @@ class NumericDecomposition {
                       const torch::Tensor& inds);
 
     // apply damping, ie on factor diag x -> x*(1+alpha) + beta (Cuda)
-    void damp_cuda(double alpha, double beta);
+    void damp_cuda(double* alpha, double* beta);
```
mhmukadam, line 82: Hmm alpha/beta here are double while elsewhere are torch::Tensor. Should we add something to the name to reduce ambiguity: alpha_double or alpha_tensor or something? Is this a concern?

theseus/optimizer/linear/baspacho_sparse_solver.py
```diff
@@ -110,10 +111,13 @@ def solve(
         if damping is None:
             damping_alpha_beta = None
         else:
-            # See Nocedal and Wright, Numerical Optimization, pp. 260 and 261
-            # https://www.csie.ntu.edu.tw/~r97002/temp/num_optimization.pdf
```
mhmukadam, line 114: Was this comment relevant to keep?

theseus/optimizer/linear/utils.py
```diff
@@ -0,0 +1,26 @@
+from typing import Tuple, Union
```
mhmukadam, line 1: License header.

mhmukadam: Also add to the new init file in this PR.

mhmukadam (approved):
Looks great. Nice work on adding this feature to the sparse solvers, especially given the cpp/cuda backends.

theseus/extlib/baspacho_solver.h
luisenp: This is the CUDA kernel for damping, so I think this is fine. It should be the only place where this happens.

luisenp:
@maurimo waiting for you review before merging.

maurimo (approved):
Looks great to me! Can merge!

PR #341, opened 2022-11-01
Update continuous integration (torch 1.13, cuda 10 dropped)

maurimo:
## Motivation and Context

Since torch-1.13 is now used in a fundamental way (functorch), this diff upgrades the ci to use torch-1.13.
Support for Cuda 10.2 is dropped, and the reference Cuda 11 version is upgraded to 11.7 from 11.3 (as recommended by torch too).

## How Has This Been Tested

Ci itself should now succeed.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] My code follows the code style of this project.


mhmukadam (approved):

maurimo:
@mhmukadam @luisenp @fantaosha thank you for the approval, but not merging yet maybe? as in the chat we discussed that FAIR cluster only has CUDA 11.4 so perhaps it's best to stick with the previous setup and just lock the version of functorch so it doesn't trigger update of torch to 1.13.

maurimo:
I propose to modify the CI (in another PR) so that functorch in installed together with torch, so we can make sure nothing goes wrong there and it's all in one place.

mhmukadam:
Okay so we can close this PR? Or should we make it a draft and revive it when we are ready to upgrade the versions.

luisenp:
Let's wait until I write the other PR, I think this one will be superseded by that one. Then we can close.

PR #349, opened 2022-11-04
Added a few other CUDA versions to build script.

luisenp:


build_scripts/build_wheel.sh
```diff
@@ -54,7 +55,9 @@ else
         BASPACHO_CUDA_ARCHS="${BASPACHO_CUDA_ARCHS};80"
         TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST};8.0"
     fi
-    if [[ ${CUDA_VERSION} = "10.2" ]] || [[ ${CUDA_VERSION} = "13.3" ]]
+    DEPRECATED_TORCH_CUDA="10.2 11.1 11.3"
```
mhmukadam, line 58: Does this conflict with SUPPORTED_CUDA_VERSIONS?

mhmukadam (approved):

build_scripts/build_wheel.sh
luisenp: These are deprecated by torch not by us. Should be fine as long as we use torch < 1.13.

mhmukadam: Should we manually fix pytorch 1.12 too? In case any other dependency later forces pytorch 1.13+

luisenp: We don't include pytorch in requirements, user has to install manually. Even if we manually fix it, if a dependency requires newer torch, it would override our requirement (this is what in fact what caused the CI and pypi errors). Also, using torch 1.13+ is only a problem if you are trying to use older CUDA.

mhmukadam: Oh yeah, totally blanked out on us installing torch separately first! Thanks for the explanation.

PR #351, opened 2022-11-04
Added option for making LieGroup checks silent.

luisenp:
There were too many warnings in our unit tests. I think this removed something like 650 warnings (this is not hyperbole).


mhmukadam (approved):

fantaosha (approved):
LGTM.

PR #352, opened 2022-11-04
Moved all unit tests to a separate folder

luisenp:
- Reduces bloat in the namespace theseus and the installed package
- Makes it easier to test different installed versions of Theseus


mhmukadam (approved):
Looks great, thanks for updating this. Much cleaner tests now!

mhmukadam:
Cuda CI needs fixing. In fact, maybe the separate command for just test_theseus_layer.py could be combined with the rest of the tests folder?

luisenp:
> Cuda CI needs fixing. In fact, maybe the separate command for just test_theseus_layer.py could be combined with the rest of the tests folder?

That separation is deliberate so that we only run CUDA-specific tests (end-to-end + marked as cudaext). We are not currently running the full suite of tests in GPU.

PR #354, opened 2022-11-07
Add tests for wheels in CI

luisenp:


luisenp:
@maurimo Do you have any suggestions for fixing the CUDA 10 wheel test? The script works for me locally, not sure why it's failing in CI.

mhmukadam (approved):
Looks good. License header for new init file is missing :P

Issue #355, opened 2022-11-07
Add device parameter to `UrdfRobotModel` in `theseus/embodied/kinematics/kinematics_model.py`


thomasweng15:
## 🚀 Feature
Add device parameter to `UrdfRobotModel` in `theseus/embodied/kinematics/kinematics_model.py`

## Motivation

Placing the robot model on gpu would enable faster operations

## Pitch

Update __init__ of UrdfRobotModel to pass a device flag to self.drm_model

## Alternatives

Keep the status quo and only perform operations on cpu

## Additional context

See the pre-existing device flag in the differentiable-robot-model repo that is not exposed in UrdfRobot Model: https://github.com/facebookresearch/differentiable-robot-model/blob/d7bd1b3b8ef1d6dabe9b68474a622185c510e112/differentiable_robot_model/robot_model.py#L94

PR #356, opened 2022-11-07
#355 Add device parameter to UrdfRobotModel

thomasweng15:
## Motivation and Context

Existing Feature Request: https://github.com/facebookresearch/theseus/issues/355

## How Has This Been Tested

Updated tests in `tests/embodied/kinematics/test_urdf_model.py`. 
The test currently defaults to `device=cuda:0` if cuda exists: https://github.com/thomasweng15/theseus/blob/3faa926334584596684eaadf0f082c4950ee763e/tests/embodied/kinematics/test_urdf_model.py#L24. 
But I have also tested with `device=cpu`. 
In both cases the tests pass. 

## Types of changes

## Checklist

### DIFF
--- a/tests/embodied/kinematics/test_urdf_model.py
+++ b/tests/embodied/kinematics/test_urdf_model.py
@@ -21,10 +21,13 @@ class VectorType(Enum):
     TH_VECTOR = 2
 
 
+device = "cuda:0" if torch.cuda.is_available() else "cpu"
+
+
 @pytest.fixture
 def robot_model():
     urdf_path = os.path.join(os.path.dirname(__file__), URDF_REL_PATH)
-    return th.eb.UrdfRobotModel(urdf_path)
+    return th.eb.UrdfRobotModel(urdf_path, device=device)
 
 
 @pytest.fixture(params=[VectorType.TORCH_TENSOR, VectorType.TH_VECTOR])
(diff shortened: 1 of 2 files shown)


tests/embodied/kinematics/test_urdf_model.py
```diff
@@ -21,10 +21,13 @@ class VectorType(Enum):
     TH_VECTOR = 2
 
 
+device = "cuda:0" if torch.cuda.is_available() else "cpu"
```
mhmukadam, line 24: Since cuda is set explicitly should we also update the CI config to handle this just like test_theseus_layer.py? I am wondering if there is an easier way to set this up say with pytest.mark, than running each file individually from CI that have explicit cuda related tests. @luisenp

mhmukadam (commented):
Thanks for adding this. Looks good so far. We should add `to()` as discussed offline.

tests/embodied/kinematics/test_urdf_model.py
mhmukadam: Based on offline discussion, we can leave this as is. CI for these tests on cuda as well on top of current cpu tests would not be necessary.

thomasweng15:
from offline discussion -- `to()` requires changes to DRM upstream, current changes in PR are sufficient for running NGDF and can be merged.

mhmukadam (approved):

theseus/embodied/kinematics/kinematics_model.py
```diff
@@ -36,7 +36,7 @@ def forward_kinematics(self, robot_pose: RobotModelInput) -> Dict[str, LieGroup]
 
 
 class UrdfRobotModel(KinematicsModel):
-    def __init__(self, urdf_path: str):
+    def __init__(self, urdf_path: str, device: str = None):
```
exhaustin, line 39: Typing should be `Optional[str]` if the argument defaults to None.
Out of curiosity, what does `device=None` do? Does it use the gpu if available and cpu otherwise?

exhaustin (approved):
One minor comment, looks good in general.
Thanks for adding!

theseus/embodied/kinematics/kinematics_model.py
thomasweng15: yes, gpu if available and cpu otherwise: https://github.com/facebookresearch/differentiable-robot-model/blob/d7bd1b3b8ef1d6dabe9b68474a622185c510e112/differentiable_robot_model/robot_model.py#L101

thomasweng15: Fixed typing to use Optional[str] here: https://github.com/facebookresearch/theseus/pull/356/commits/272a4de77113abbae09b5d621662de4eb11cb644

mhmukadam: Good catch @exhaustin!

luisenp (approved):
LGTM. Let's merge as is. I can add the changes discussed offline later today.

PR #357, opened 2022-11-08
Added th.device to represent both str and torch.device.

luisenp:
This PR also adds a file `py.typed` so that `mypy` knows how to use our type annotations in other projects.

### DIFF
--- a/theseus/__init__.py
+++ b/theseus/__init__.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 __version__ = "0.1.2"
 
+from .constants import _TheseusDevice as device
 
 from .core import (  # usort: skip
     AutoDiffCostFunction,
--- a/theseus/__init__.py
+++ b/theseus/__init__.py
@@ -4,7 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 __version__ = "0.1.2"
 
-from .constants import _TheseusDevice as device
+from .constants import DeviceType as DeviceType
 
 from .core import (  # usort: skip
     AutoDiffCostFunction,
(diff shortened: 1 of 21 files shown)


theseus/geometry/se2.py
```diff
@@ -44,7 +44,7 @@ def rand(
         *size: int,
         generator: Optional[torch.Generator] = None,
         dtype: Optional[torch.dtype] = None,
-        device: Optional[torch.device] = None,
+        device: Optional[theseus.constants._TheseusDevice] = None,
```
mhmukadam, line 47: Some files instead do `from theseus.constants import _TheseusDevice`. We could pick one for consistency.

luisenp: I was going to change all of them to the `from` version, but the files that have the full import are using things like `PI` and `EPS` and I thought it would just make the code really confusing for people to figure out where these are coming from. I thought clarity was better than consistency in this case.

theseus/__init__.py
```diff
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 __version__ = "0.1.2"
 
+from .constants import _TheseusDevice as device
```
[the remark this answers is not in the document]
luisenp: Oh, yes, I like this better.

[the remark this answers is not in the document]
luisenp: Changed.

theseus/geometry/se2.py
mhmukadam: Yeah makes sense. Should we switch others to `theseus.constants._TheseusDevice` also?

mhmukadam (approved):

theseus/geometry/se2.py
luisenp: Too verbose, I think? I don't think it's a big deal if we are not fully consistent.

PR #360, opened 2022-11-14
Add sparse solvers support to LM adaptive damping

luisenp:
This PR adds a method to compute `Atb` in the sparse linearization, which we can use for the adaptive damping computation. 

I also sneaked in a change to the PGO example, to make it easier to pass optimizer kwargs.

### DIFF
--- a/theseus/optimizer/sparse_linearization.py
+++ b/theseus/optimizer/sparse_linearization.py
@@ -150,6 +150,8 @@ class SparseLinearization(Linearization):
                 self.objective.device
             )
             A_colInd = A_rowPtr.new_tensor(self.A_col_ind)
+
+            # unsqueeze at the end for consistency with DenseLinearization
             self._Atb = tmat_vec(
                 self.objective.batch_size,
                 self.num_cols,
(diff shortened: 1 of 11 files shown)


theseus/optimizer/sparse_linearization.py
```diff
@@ -20,7 +21,7 @@ def __init__(
         self,
         objective: Objective,
         ordering: Optional[VariableOrdering] = None,
-        **kwargs
+        **kwargs,
```
mhmukadam, line 24: Comma not needed.

mhmukadam (approved):

theseus/optimizer/sparse_linearization.py
luisenp: Comma at the end of multiline arguments is the standard in `black`.

maurimo (approved):
LGTM!
As a very minor note for the future, it would be great to separate the PRs about refactoring and those with new functionality, here most of the code differences are the renaming of `AtA` to `_AtA` and `tmat_vec/mat_vec` refactor.

PR #363, opened 2022-11-16
Change Objective.error() so it can be vectorized w/o changing vectorized cache

### LINKED ISSUE #231: RobustCostFunction weighted error is different when vectorization is on/off.
## 🐛 Bug
Reported by @joeaortiz 

When using `RobustCostFunction` we get different results in `Objective.error()` if vectorization is turned on/off. The reason is that when vectorization is off we get the result from `cost_function.weighted_error()`, and when it's turned on we get it (indirectly) from `cost_function.weighted_jacobians_and_error()`. As explained in the code (and to be documented), `RobustCostFunction` has different behavior for each of this methods; the jacobians version of the error is rescaled so that it can be used by the optimizers, while the other one is the actual error value (not rescaled). 

My proposal (using input from @joeaortiz) is to find a way for the Vectorization to serve the result of `weighted_error` when `_CostFunctionWrapper` is called, rather than the one from `weighted_jacobians_and_error`.

luisenp:
This will be useful for improving the accept/reject step of LM, and for more complex line search later. 

Closes #231. 

Note that as a consequence of this, this PR changes PGO optimization numerical results because now Objective.error() returns the non-scaled weighted error for `RobustCostFunction`, while the previous version was returning the scaled one.

### DIFF
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -66,8 +66,14 @@ class Objective:
         self.current_version = 0
 
         # ---- Callbacks for vectorization ---- #
-        # This gets replaced when cost function vectorization is used
-        self._cost_functions_iterable: Optional[Iterable[CostFunction]] = None
+        # This gets replaced when cost function vectorization is used.
+        # Normally it returns an iterator over `self.cost_functions.values()`, so
+        # that calling error() or jacobians() on the yielded cost functions
+        # computes these quantities on demand.
+        # But when vectorization is on, it returns separate cost function
+        # containers that serve cached jacobians and errors that have been
+        # previously by the vectorization computed.
+        self._jacobians_iter: Optional[Iterable[CostFunction]] = None
 
         # Used to vectorize cost functions error + jacobians after an update
         # The results are cached so that the `self._get_jacobians_iter()` returns
@@ -543,9 +549,10 @@ class Objective:
 
     def _get_jacobians_iter(self) -> Iterable:
         self.update_vectorization_if_needed()
-        if self._cost_functions_iterable is None:
+        if self._jacobians_iter is None:
+            # No vectorization is used, just serve from cost functions
             return iter(cf for cf in self.cost_functions.values())
-        return iter(cf for cf in self._cost_functions_iterable)
+        return iter(cf for cf in self._jacobians_iter)
 
     # Applies to() with given args to all tensors in the objective
     def to(self, *args, **kwargs):
@@ -591,7 +598,7 @@ class Objective:
 
     def _enable_vectorization(
         self,
-        cost_fns_iter: Iterable[CostFunction],
+        jacobians_iter: Iterable[CostFunction],
         vectorization_run_fn: Callable,
         vectorized_to: Callable,
         vectorized_retract_fn: Callable,
@@ -603,7 +610,7 @@ class Objective:
             enabler.__module__ == "theseus.core.vectorizer"
             and enabler.__class__.__name__ == "Vectorize"
         )
-        self._cost_functions_iterable = cost_fns_iter
+        self._jacobians_iter = jacobians_iter
         self._vectorization_run = vectorization_run_fn
         self._vectorization_to = vectorized_to
         self._retract_method = vectorized_retract_fn
@@ -612,7 +619,7 @@ class Objective:
 
     # Making public, since this should be a safe operation
     def disable_vectorization(self):
-        self._cost_functions_iterable = None
+        self._jacobians_iter = None
         self._vectorization_run = None
         self._vectorization_to = None
         self._retract_method = Objective._retract_base
@@ -622,7 +629,7 @@ class Objective:
     def vectorized(self):
         assert (
             (not self._vectorized)
-            == (self._cost_functions_iterable is None)
+            == (self._jacobians_iter is None)
             == (self._vectorization_run is None)
             == (self._vectorization_to is None)
             == (self._retract_method is Objective._retract_base)
--- a/theseus/core/objective.py
+++ b/theseus/core/objective.py
@@ -623,6 +623,7 @@ class Objective:
         self._vectorization_run = None
         self._vectorization_to = None
         self._retract_method = Objective._retract_base
+        self._get_error_iter = Objective._get_error_iter_base
         self._vectorized = False
 
     @property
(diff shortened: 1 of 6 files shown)


theseus/core/objective.py
```diff
@@ -532,11 +538,14 @@ def update_vectorization_if_needed(self):
     def __iter__(self):
         return iter([cf for cf in self.cost_functions.values()])
 
-    def _get_iterator(self):
+    def _get_error_iter_base(self) -> Iterable:
+        return iter(cf for cf in self.cost_functions.values())
+
+    def _get_jacobians_iter(self) -> Iterable:
```
mhmukadam, line 544: Trying to understand the difference between error iter and jacobians iter since they both return on cost_functions.values().

theseus/core/vectorizer.py
```diff
@@ -33,6 +34,12 @@ def _varinfo(var) -> str:
     )
 
 
+class _VectorizationMode(Enum):
+    ERROR = 0
+    WEIGHTED_ERROR = 1
+    FULL = 2
```
mhmukadam, line 40: Might be helpful to leave some brief comments about each mode and what it does.

theseus/core/objective.py
luisenp: The ones that return `self.cost_functions.values()` are the non-vectorized versions, which should be the same in both cases. When vectorized is on, they return different iterators: the jacobians iter will point to containers that have a permanent cache of errors and jacobians, the error iter will point to a tmp container that only stores errors computed on demand. I added some comments at the top explaining `_get_error_iter`, but maybe the variable names are also non-descriptive enough.

theseus/core/vectorizer.py
luisenp: I'll add some comments.

mhmukadam (approved):
Looks great!

theseus/core/objective.py
mhmukadam: Okay that makes sense. Thanks for the explanation and comment will be useful.

theseus/core/objective.py
```diff
@@ -66,8 +66,14 @@ def __init__(self, dtype: Optional[torch.dtype] = None):
...
         # ---- Callbacks for vectorization ---- #
-        # This gets replaced when cost function vectorization is used
-        self._cost_functions_iterable: Optional[Iterable[CostFunction]] = None
+        # This gets replaced when cost function vectorization is used.
+        # Normally it returns an iterator over `self.cost_functions.values()`, so
+        # that calling error() or jacobians() on the yielded cost functions
+        # computes these quantities on demand.
+        # But when vectorization is on, it returns separate cost function
+        # containers that serve cached jacobians and errors that have been
+        # previously by the vectorization computed.
```
mhmukadam, line 75: typo: ... computed by the vectorization.

mhmukadam:
Do any of the PGO example scripts need updating?

theseus/core/objective.py
```diff
@@ -532,11 +544,15 @@ def update_vectorization_if_needed(self):
...
+
+    def _get_jacobians_iter(self) -> Iterable:
         self.update_vectorization_if_needed()
-        if self._cost_functions_iterable is None:
-            return iter([cf for cf in self.cost_functions.values()])
-        return iter([cf for cf in self._cost_functions_iterable])
+        if self._jacobians_iter is None:
+            # No vectorization is used, just serve from cost functions
+            return iter(cf for cf in self.cost_functions.values())
+        return iter(cf for cf in self._jacobians_iter)
```
fantaosha, line 555: Do we need to cache `self._jacobians_iter` here? For example, 
```
if self._jacobians_iter is None:
     self._jacobians_iter = iter(cf for cf in self.cost_functions.values())
return iter(cf for cf in self._jacobians_iter)
```

theseus/core/vectorizer.py
```diff
@@ -33,6 +34,18 @@ def _varinfo(var) -> str:
     )
 
 
+# Identifies what quantities vectorizaton computes for all cost functions
+class _VectorizationMode(Enum):
```
fantaosha, line 38: Are these three models selected by the user? Looks not?

theseus/core/vectorizer.py
```diff
@@ -328,16 +360,35 @@ def _handle_schema_vectorization(
...
         Vectorize._compute_error_and_replace_wrapper_caches(
-            vectorized_cost_fn, cost_fn_wrappers, batch_size
+            vectorized_cost_fn, cost_fn_wrappers, batch_size, mode
         )
 
-    def _vectorize(self):
-        self._clear_wrapper_caches()
-        for schema, cost_fn_wrappers in self._schema_dict.items():
+    def _vectorize(
+        self, mode: _VectorizationMode = _VectorizationMode.FULL
```
fantaosha, line 367: How to select different vectorization mode? According to memory usage?

fantaosha (approved):
LGTM. Only minor comments.

theseus/core/objective.py
luisenp: I'm using `None` as a surrogate for `self.vectorized == True`, but maybe this is not ideal. I see if I can replace with 
```python
is self.vectorized:
    return iter(cf for cf in self._jacobians_iter)
else:
    return iter(cf for cf in self.cost_functions.values())
```
I should probably also rename this iterator to `self._vectorized_jacobian_iter`, seems more clear.

theseus/core/vectorizer.py
luisenp: Currently we select it manually in two places:

1.) when requesting to compute `Objective.error()` we select `WEIGHTED_ERROR`, so that we compute the vectorized error w/o modifying the permanent caches. 
2.) when Objective requests that the vectorization is updated, then we use `FULL` so that we compute jacobians and replace the permanent caches. 

It's good that you asked because I just realized that running vectorization for 1.) only makes sense if `error` receives `input_tensors`. But if this is `None` or `empty`, then there is no point in computing vectorization again, we can just return the cached values.

theseus/core/vectorizer.py
luisenp: No, they are currently private, as all of this happens under the hood.

theseus/core/vectorizer.py
luisenp: Oh nevermind. I was about to fix this, but if we return the jacobians cache we go back to the issue were the error under robust costs is the scaled error instead of the actual error. I guess we could add some logic to prevent recompute if there are no robust costs, but this seems a bit messy and prone to error. I'll leave as is for now.  

FWIW, in code about to be merged in other PRs the optimizers always compute error with `input_tensors` set to the latest values, so I don't think this optimization will save much time.

luisenp: Added.

PR #364, opened 2022-11-16
Add proper accept/reject logic for LM optimizer

luisenp:
The previous version would adjust the damping depending on whether the step should be accepted or not, but it would nevertheless always take the step. This could result in divergence if the down damping ratio was too aggressive. 

In the new version, if the step should be rejected, it's indeed rejected (and this is computed independently for each batch element).

### DIFF
--- a/theseus/optimizer/nonlinear/levenberg_marquardt.py
+++ b/theseus/optimizer/nonlinear/levenberg_marquardt.py
@@ -2,7 +2,7 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Sequence, Type, Union
 
 import torch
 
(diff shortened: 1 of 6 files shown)


theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -268,6 +270,10 @@ def _merge_infos(
             M & (grad_loop_info.status == NonlinearOptimizerStatus.MAX_ITERATIONS)
         ] = -1
 
+    def _update_tmp_optim_vars(self):
```
mhmukadam, line 273: Might be helpful to leave a brief comment. Are the tmp vars still in the compute graph?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -460,30 +466,55 @@ def _optimize_impl(
     def compute_delta(self, **kwargs) -> torch.Tensor:
         pass
 
+    # Given descent directions and step sizes, updates the optimization
+    # variables.
+    # Batch indices indicated by `converged_indices` mask are ignored
+    # unless `force_update = True`.
+    # Returns the total error tensor after the update
+    def step(
```
mhmukadam, line 474: Should this be also a private function `_step`?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -460,30 +466,55 @@ def _optimize_impl(
...
+        self,
+        delta: torch.Tensor,
+        previous_err: torch.Tensor,
+        converged_indices: torch.Tensor,
+        force_update: bool,
+        **kwargs,
+    ) -> torch.Tensor:
+        # makes sure tmp cotainers are up to date with current variables
```
mhmukadam, line 482: typo: containers

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -460,30 +466,55 @@ def _optimize_impl(
...
+        self._update_tmp_optim_vars()
+        # stores the result of the retract step in `self._tmp_optim_vars`
+        self.objective.retract_vars_sequence(
```
mhmukadam, line 485: Slightly confused, would this update the optim vars of the objective with the delta retract on the tmp optim vars? Or does this happen below when calling objective.update.

mhmukadam (approved):
Looks good! Do the current unit tests cover the new step logic?

luisenp (commented):
Thanks for the comments! I didn't think of adding new tests, since the optimizer needs to go through all the new step logic. I guess it's possible to add some tests specifically for accept/reject, but that seems a bit cumbersome, since I think I'd need to mock a lot of things to be able to test it.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
mhmukadam: Okay. Maybe a brief comment block at the top might be helpful to write more optimizers in the future: describing a typical flow of what functions get called, in what order, and what they do.

luisenp: Good idea, I'll add it.

luisenp: Added.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Added comments at the top of this method.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Fixed.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -80,6 +80,32 @@ def resolve(key: Union[str, "BackwardMode"]) -> "BackwardMode":
...
+#
+#   - `compute_delta`: returns a descent direction given the current values
+#     of the objective's optimization vars.
+#
+# Optionally, they can also provide the following methods:
+#
+#   - `reset`: resets any internal state needed by the optimizer.
+#   - `_complete_step`: called at the end of an optimization step, but before
+#     optimization variables are updated. Returns batch indices that should not
+#     any be updated (e.g., if the step is to be rejected).
```
mhmukadam, line 94: typo: should not be updated

luisenp: Will fix this in the Dogleg PR.

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -133,37 +133,62 @@ def compute_delta(
...
             self._damping.view(-1, 1)
             if isinstance(self._damping, torch.Tensor)
             else self._damping
         )
+        # Deliberately using Atb before updating the variables, according to
+        # the LM reference above
         den = (delta * (damping * delta + linearization.Atb.squeeze(2))).sum(dim=1)
-        rho = (last_err - new_err) / den
-        good_idx = rho > damping_accept
+        rho = (previous_err - err) / den
```
fantaosha, line 184: Wondering if we should rename `previous_err` and `err` to `previous_cost` and `cost`. In our convention, we assume `err` is vector.

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -133,37 +133,62 @@ def compute_delta(
...
+        reject_indices = rho <= damping_accept
         self._damping = torch.where(
-            good_idx,
-            self._damping / down_damping_ratio,
+            reject_indices,
             self._damping * up_damping_ratio,
+            self._damping / down_damping_ratio,
```
fantaosha, line 189: Maybe we should change the `_damping` update policy. In Ceres,  `up_damping_ratio`/`down_damping_ratio` are also dynamically [updated](https://github.com/ceres-solver/ceres-solver/blob/c8493fc366950d1b6d52c2bcc6ec874775c77bd6/internal/ceres/levenberg_marquardt_strategy.cc#L149).

fantaosha: In LM method, the `self._damping` might be unchanged if the the `rho` is neither good nor bad. Therefore, there should `good_indices`, `bad_indices` and `accepted_indices`, where `self._damping` is reduced, increased and unchanged, respectively.

fantaosha (commented):
LGTM. There are some comments which seem difficult to be addressed in this PR.

theseus/optimizer/nonlinear/levenberg_marquardt.py
luisenp: Created #384

PR #365, opened 2022-11-17
Add  theseus.geometry.functional.so3 module with exp() and jexp() implementations

fantaosha:
## Types of changes

## Checklist


theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,197 @@
...
+    def forward(
+        cls,
+        ctx,
+        tangent_vector,
+        jacobians=None,
+    ):
+        tangent_vector: torch.Tensor = cast(torch.Tensor, tangent_vector)
+        ret = cls.call(tangent_vector, jacobians)
+        ctx.save_for_backward(tangent_vector, ret)
+        ctx.jacobians = jacobians
```
fantaosha, line 173: It should be `ctx.jacobians = None if jacobians is None else jacobians[0].clone()`

theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,202 @@
...
+
+            jacobians.append(jac)
+
+        return ret
+
+    @classmethod
+    def jacobian(cls, tangent_vector: torch.Tensor) -> torch.Tensor:
+        if not check_tangent_vector(tangent_vector):
+            raise ValueError("Tangent vectors of SO3 should be 3-D vectors.")
+        tangent_vector = tangent_vector.view(-1, 3)
```
luisenp, line 122: I see a few lines of duplicated code between `error` and `jacobian`. This should be extracted out into some common function.

theseus/geometry/functions/tests/test_so3.py
```diff
@@ -0,0 +1,18 @@
...
+from theseus.constants import TEST_EPS
+
+import torch
+
+
+@pytest.mark.parametrize("batch_size", [1, 20, 100])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
+def test_exp_map(batch_size: int, dtype: torch.dtype):
+    tangent_vector = torch.rand(batch_size, 3, dtype=dtype)
+    check_lie_group_function(SO3Function, "ExpMap", TEST_EPS, tangent_vector)
```
luisenp, line 18: We should probably write new tests for these functions that better suited to the structure of this code, instead of relying on tests based on the old logic.

theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,202 @@
...
+
+def check_tangent_vector(tangent_vector: torch.Tensor) -> bool:
+    _check = tangent_vector.ndim == 3 and tangent_vector.shape[1:] == (3, 1)
+    _check |= tangent_vector.ndim == 2 and tangent_vector.shape[1] == 3
+    return _check
+
+
+class ExpMap(LieGroupExpMap):
+    @classmethod
+    def call(
```
luisenp, line 49: The list of jacobians is something we did in `theseus` as a way to avoid having multiple return types for Lie Group operations, so that we could do things like `x.compose(x.inverse())`, even if we wanted to return the jacobians. For functions with `forward()`, and `backward()` we don't need this syntax anymore, which always felt a bit clunky. 

I would prefer a structure like this:

```python
def _so3_exp_map_impl(tangent_vector: torch.Tensor) -> Tuple[torch.Tensor, ...]:
    # Returns the matrix result, as well as any intermediate tensor needed to compute the jacobian

def _so3_jac_exp_map_impl(tangent_vector) -> Tuple[torch.Tensor, torch.Tensor]:
    exp_map, other_tensors = _so3_exp_map_impl(tangent_vector)
    return _so3_jac_exp_map_aux(exp_map, other_tensors)

def _so3_jac_exp_map_aux(exp_map: torch.Tensor, *other_tensors) -> torch.Tensor:
   # Returns the jacobian given the exp_map and any pre-computed intermediate tensors 

class ExpMap(LieGroupExpMap):
    @staticmethod
    def forward(ctx, tangent_vector):
        exp_map, other_tensors = _so3_jac_exp_map_impl(tangent_vector)
        ctx.save_for_backward(exp_map, *other_tensors)

   @staticmethod
   def backward(ctx, grad_output):
        # Compute jacobian, use a cache, multiply by grad_output, etc. 
        # Here we just call _so3_jac_exp_map_aux(ctx.exp_map, *ctx.other_tensors)
```

Of course, the function to compute the jacobians can also be public, since we'll probably use it in some cost functions.

luisenp (changes requested):
Thanks for getting started on this @fantaosha! I left some comments, in addition to the discussion that we had offline.

theseus/geometry/functions/so3_function.py
fantaosha: Thanks for the detailed structure. If I understand correctly, jacobians should be one of the returns of `forward()`. Thus, we have to provide an analytical `backward()` for the jacobians as well, which can be very complicated.

luisenp: Do you mean in my proposal? If so, then no, I wasn't suggesting jacobians are returned by forward, only the exp map. But I just noticed there was a typo, the first line in my forward was supposed to be 
```python
exp_map, other_tensors = _so3_exp_map_impl(tangent_vector)
```
The jacobians are only computed in backward.

fantaosha: Got it. But the `jacobians` got from `backward()` are `Euclidean gradients`. How can people get access to `Lie tangent jacobians`?

luisenp: Could be a separate function inside so3.py like
```python
def jexp_tangent(tangent_vector: torch.Tensor) -> torch.Tensor:
    ...
```
or if the logic allows it, we could do
```python
def jexp(tangent_vector: torch.Tensor, is_euclidean: bool = False) -> torch.Tensor:
     pass
```
BTW, I'm thinking that maybe we don't need the word "map" for log map, or exp map? I think short names would make sense to be in the spirit of a lightweight library.

fantaosha: I would like the jacobians to be computed with `exp_map`, which some of the computation can be reused.

luisenp: I already proposed one way to reuse computation, by returning intermediate tensors as a separate output in the private `_exp_map_impl`. This lets you defer the final computatio of the jacobians to only be done if backward is called.

theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,197 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import theseus
```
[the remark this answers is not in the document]
mhmukadam: Same goes for the unit tests.

theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,202 @@
...
+    return "SO3"
+
+
+def dim() -> int:
+    return 3
+
+
+def check_group_tensor(tensor: torch.Tensor) -> bool:
+    with torch.no_grad():
+        if tensor.ndim != 3 or tensor.shape[1:] != (3, 3):
```
mhmukadam, line 25: Should we allow any batch dims (..., 3, 3) rather than just one so things are more torch like?

mhmukadam, line 25: Related to Luis' comment about support arbitrary batch dims to be more torch like, here it would mean (..., 3, 3).

theseus/geometry/functions/so3_function.py
mhmukadam: I agree that we want to move away from the cpp like calls that fill jacobians passed as a reference in the function call.

For a python/pytorch user we need to support say
```
y = x.compose(x.inverse())
y.backward()
```

But we also need the ability to get function jacobians for manual cost functions (not based on AutoDiffCostFunction). Would we just call a partial backward here? For this to be efficient maybe we need to support both forward and backward mode differentiation?

theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,202 @@
...
+    @classmethod
+    def call(
+        cls,
+        tangent_vector: torch.Tensor,
+        jacobians: Optional[List[torch.Tensor]] = None,
+    ) -> torch.Tensor:
+        if not check_tangent_vector(tangent_vector):
+            raise ValueError("Tangent vectors of SO3 should be 3-D vectors.")
+        tangent_vector = tangent_vector.view(-1, 3)
+        ret = tangent_vector.new_zeros(tangent_vector.shape[0], 3, 3)
```
mhmukadam, line 57: I am assuming the math is same as out current implementation and therefore functorch compatible? We might want to include some unit tests related to vectorization as well for this api.

mhmukadam (commented):
Looked at this so far. We might need some more discussion before locking the api.

theseus/geometry/functions/so3_function.py
```diff
@@ -0,0 +1,197 @@
...
+            torch.matmul(tensor, tensor.transpose(1, 2))
+            - torch.eye(3, 3, dtype=tensor.dtype, device=tensor.device)
+        ).abs().max().item() < MATRIX_EPS
+        _check &= (torch.linalg.det(tensor) - 1).abs().max().item() < MATRIX_EPS
+
+    return _check
+
+
+def check_tangent_vector(tangent_vector: torch.Tensor) -> bool:
+    _check = tangent_vector.ndim == 3 and tangent_vector.shape[1:] == (3, 1)
```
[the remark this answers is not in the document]
fantaosha: The support of arbitrary dimension should be done in 2023 H1. Considering there are lots of uncertainties, I think we should first start with one batch dimension, which is faster and will not block differentiable FK.

theseus/geometry/functional/constants.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
fantaosha, line 1: This file is copied from https://github.com/facebookresearch/theseus/blob/main/theseus/constants.py

theseus/geometry/functions/so3_function.py
fantaosha: Yes, it is functorch compatible.

theseus/geometry/functions/so3_function.py
fantaosha: We will support this in the 2023 H1.

theseus/geometry/functions/tests/test_so3.py
fantaosha: I wrote new tests.

theseus/geometry/functional/lie_group.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ExpMap(torch.autograd.Function):
+    @classmethod
+    @abc.abstractmethod
+    def forward(cls, ctx, tangent_vector):
+        pass
+
+
+def ExpMapFactory(module):
```
luisenp, line 20: Why not pass the two functions directly? `exp_autograd_fn` and `jexp_impl`? No need to be fancy here.

theseus/geometry/functional/so3.py
```diff
@@ -0,0 +1,158 @@
...
+                dR[:, 1, 0] - dR[:, 0, 1],
+            ),
+            dim=1,
+        ).view(-1, 3, 1)
+        return grad_input.view(-1, 3)
+
+
+_module = get_module(__name__)
+
+exp_map, jexp_map = LieGroup.ExpMapFactory(_module)
```
luisenp, line 158: If you pass `ExpMap` and `_j_exp_map_impl` as arguments, no need to use this `get_module`.

theseus/geometry/functional/tests/common.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+
+def check_lie_group_function(module, method: str, atol: float, *args):
```
luisenp, line 9: Here it is also clearer to directly pass the functions you are testing. Also, I don't quite understand the difference between `func_call` and `func_apply`. I'm guessing this names are remnants from the initial implementation?

luisenp (changes requested):
Thanks for iterating so much on these. I think this is almost there! Left some additional comments.

theseus/geometry/functional/lie_group.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class ExpMap(torch.autograd.Function):
+    @classmethod
+    @abc.abstractmethod
+    def forward(cls, ctx, tangent_vector):
+        pass
+
+
+class LieGroup:
```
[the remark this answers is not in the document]
fantaosha: No `LieGroup` any more.

theseus/geometry/functional/lie_group.py
```diff
@@ -0,0 +1,40 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import abc
+
+from typing import List, Optional, Tuple
+from .utils import get_module, check_jacobians_list
+
+
+class ExpMap(torch.autograd.Function):
```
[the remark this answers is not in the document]
fantaosha: I think this is a standard way to do this. For some functions like `Inverse`, we might need this function.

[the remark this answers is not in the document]
luisenp: It's standard if we are using inheritance, but we are not doing this here, since we retrieve the attribute directly from each specific module, no?

theseus/geometry/functional/constants.py
fantaosha: I think we will use all of them very shortly. If everything goes well, `SO3` and `SE3` should be done very soon.

theseus/geometry/functional/so3.py
```diff
@@ -0,0 +1,157 @@
...
+import torch
+
+from . import constants
+from . import lie_group as LieGroupModule
+from .lie_group import LieGroup
+
+from typing import cast
+
+
+def name() -> str:
```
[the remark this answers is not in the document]
fantaosha: Will be used in error messages.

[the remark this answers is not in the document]
fantaosha: > These don't need to be functions, you can use constants here. Something like `NAME` and `DIM`. Also, where are these used? Is that in another PR?

Looks like we will not implement `Vector`. If so, we can use `NAME` and `DIM` as constants here.

theseus/geometry/functional/tests/common.py
fantaosha: Rename some of the variables, and the differences are explained in https://github.com/facebookresearch/theseus/blob/86ea13facd1acfb46b522f0aa6ddfcb4de71662d/theseus/geometry/functional/lie_group.py#L12

theseus/geometry/functional/lie_group.py
```diff
@@ -0,0 +1,39 @@
...
+import abc
+
+from typing import List, Optional
+from .utils import check_jacobians_list
+
+# There are four functions associated with each Lie group operator xxx.
+# _xxx_impl: mathematical implementation of the operator
+# _j_xxx_impl: mathematical implementation of the operator jacobian
+# _xxx_base: a torch.autograd.Function wrapper of _xxx_impl
+# _j_xxx_base: simply equivalent to _j_xxx_impl
```
mhmukadam, line 16: wondering if `_xxx_wrapper` is a better name since `base` makes me think of the low level math which is actually in `impl`.

Also the comment for `_j_xxx_base` should also include torch.autograd.Function in the text similar to `_xxx_base`.

theseus/geometry/functional/lie_group.py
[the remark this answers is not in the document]
fantaosha: For some functions such as `Inverse`, we might need inheritance to reuse the same code for different code. We can definitely remove `ExpMap` now.

theseus/geometry/functional/lie_group.py
```diff
@@ -0,0 +1,39 @@
...
+
+def UnaryOperatorFactory(module, op_name):
+    op_base = getattr(module, "_" + op_name + "_base")
+    j_op_base = getattr(module, "_j_" + op_name + "_base")
+
+    def op(
+        input: torch.Tensor,
+        jacobians: Optional[List[torch.Tensor]] = None,
+    ) -> torch.Tensor:
+        if jacobians is not None:
```
mhmukadam, line 34: I guess we are still discussing whether to keep the cpp style jacobian in function call, around for the time being. We should condier how messy it would be to phase out later.

theseus/geometry/functional/lie_group.py
fantaosha: The implementation of `torch.autograd.Function` for `_j_xxx_impl` is planned for 2023 H1.

mhmukadam (commented):
Nice work @fantaosha and @luisenp, this code reads so much better!

theseus/geometry/functional/lie_group.py
fantaosha: My plan is to keep the old API now to be compatible with differentiable FK. Then, we might refactor the API as a whole together with Theseus.

theseus/geometry/functional/lie_group.py
mhmukadam: Based on offline discussion here are functions we need

```python
_exp_impl  # analytic op
_jexp_impl  # analytic jop
_exp_wrapper  # torch autograd op
_jexp_wrapper = _jexp_impl # later torch autograd jop

exp = _exp_wrapper
jexp = _jexp_wrapper
```

theseus/geometry/functional/so3.py
```diff
@@ -0,0 +1,171 @@
...
+            ),
+            dim=1,
+        ).view(-1, 3, 1)
+        return grad_input.view(-1, 3)
+
+
+_module = get_module(__name__)
+
+_exp_map_autograd_fn = ExpMap.apply
+_jexp_map_autograd_fn = _jexp_map_impl
```
luisenp, line 169: Add a comment here that says we are using this as a placeholder while we add an actual implementation for this.

luisenp (approved):
LGTM. Only one minor comment to address. Let's remove the use of `_map` also.

theseus/geometry/functional/so3.py
fantaosha: Fixed.

luisenp (approved):
LGTM!

theseus/geometry/functional/so3.py
```diff
@@ -0,0 +1,172 @@
...
+    jac[:, 1, 0] -= jac_temp[:, 2]
+    jac[:, 0, 2] -= jac_temp[:, 1]
+    jac[:, 2, 0] += jac_temp[:, 1]
+    jac[:, 1, 2] += jac_temp[:, 0]
+    jac[:, 2, 1] -= jac_temp[:, 0]
+
+    return [jac], ret
+
+
+class ExpMap(LieGroup.UnaryOperator):
```
mhmukadam, line 139: Should we drop the 'Map' from `ExpMap` here as well?

PR #371, opened 2022-11-21
Dogleg

### LINKED ISSUE #15: Dogleg nonlinear optimizer

luisenp:
Closes #15 

This is a partial implementation, that doesn't have support for `SparseLinearization`. Planning to add that right after, but we need to add differentiable jacobian vector product for our sparse operations.


luisenp:
The test should pass once we merge #372

theseus/optimizer/nonlinear/__init__.py
```diff
@@ -13,3 +13,4 @@
     NonlinearOptimizerParams,
     NonlinearOptimizerStatus,
 )
+from .trust_region import TrustRegionOptimizer
```
mhmukadam, line 16: We could drop `Optimizer` from name to `TrustRegion` for consistency with the other names.

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,133 @@
...
+
+        linearization = self.linear_solver.linearization
+        # TODO: For sparse I'll add a method like linearization.Avp() for Jacobian
+        # vector product. The issue is that for the sparse linearization we need
+        # to make this differentiable, because we are not using torch's
+        # native functions
+        assert isinstance(linearization, DenseLinearization)
+
+        neg_grad = linearization.Atb.squeeze(2)
+        with torch.no_grad():
```
fantaosha, line 75: Why `with torch.no_grad()`? I think we still need `backward()` for differentiable optimization

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,133 @@
...
+        if not isinstance(self.linear_solver.linearization, DenseLinearization):
+            # Since I will implement for sparse soon after,
+            # I'll avoid fancier error handling
+            # I expect this method to work with all our current solvers
+            raise NotImplementedError
+
+    def _compute_delta_impl(self) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+        trust_region_2 = self._trust_region**2
+        delta_gn = self.linear_solver.solve()
+        delta_gn_norm_2 = TrustRegion._detached_squared_norm(delta_gn)
```
fantaosha, line 57: The squared norm is usually scaled with the diagonals of the hessian, which practically has better convergence.

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,133 @@
...
+        assert isinstance(linearization, DenseLinearization)
+
+        neg_grad = linearization.Atb.squeeze(2)
+        with torch.no_grad():
+            Adelta_sd = linearization.A.bmm(neg_grad.unsqueeze(2)).squeeze(2)
+            Adelta_sd_norm_2 = TrustRegion._detached_squared_norm(Adelta_sd)
+            grad_norm_2 = TrustRegion._detached_squared_norm(neg_grad)
+            t = grad_norm_2 / (Adelta_sd_norm_2 + Dogleg.EPS)
+            delta_sd = neg_grad * t
+            delta_sd_norm_2 = TrustRegion._detached_squared_norm(delta_sd)
```
fantaosha, line 81: Maybe diagonal-scaled squared norm here as well. More details can be found [here](https://www.mathworks.com/help/optim/ug/equation-solving-algorithms.html).

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,182 @@
...
+        if not good_params:
+            raise ValueError(
+                "Invalid parameters for TrustRegionMethod. "
+                "Values must satisfy <accept/shrink>_threshold < expand_threshold, "
+                "shrink_ratio in (0, 1], and expand_ratio > 1.0."
+            )
+        pred_err = self._predicted_error(previous_err, delta)
+        rho = ((previous_err - new_err) / (previous_err - pred_err)).view(-1, 1)
+        shrink_idx = rho < shrink_threshold
+        self._trust_region = torch.where(
```
fantaosha, line 172: For LM method, the trust-region should be scaled by the diagonals.

luisenp: While LM with adaptive damping is a trust region method, it doesn't inherit from `TrustRegion` because the current `_complete_step` logic is a bit different. I think it's def worth considering unifying these, but I'd prefer to leave for a separate PR to keep things simple.

theseus/optimizer/nonlinear/dogleg.py
luisenp: I was a bit concerned about potential outer loss instabilities due to numerical issues, since there is a lot of branching involved in these operations, and things get messy when `delta` or `t` are near zero. But let me play around a bit with removing this context. 

Do note that we can still backpropagate through this `compute_delta()`, since only some auxiliary operations are detached; all descent directions are tensors attached to the compute graph.

theseus/optimizer/nonlinear/dogleg.py
luisenp: I guess the main question is if the current version is incorrect. Based on the reference I followed, it seems that the vanilla formulation uses non-scaled squared norm (scaling is introduced one section later). So, I think it's fine to leave as is to get a working version in, and introduce this improvement in a separate PR.

theseus/optimizer/nonlinear/dogleg.py
luisenp: See comment below.

theseus/optimizer/nonlinear/dogleg.py
fantaosha: Agreed.

theseus/optimizer/nonlinear/trust_region.py
fantaosha: Ah, sorry. I think it might be better for the LM method to rescale the Hessian matrix as `H + damping_factor * diag(H)`.

luisenp: This is the `elliptical` damping option. We already have that.

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # ---------------------------------------------------------------------
+        # Some Gauss-Newton steps are outside trust-region,
+        # need to compute dogleg step
+        # ---------------------------------------------------------------------
+
+        linearization = self.linear_solver.linearization
+
+        neg_grad = linearization.Atb.squeeze(2)
+        Adelta_sd = linearization.Av(neg_grad)
```
fantaosha, line 67: So this is the computation of the Cauchy step?

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,121 @@
...
+            # Need to solve a quadratic || sd + tau * (gn - sd)|| == tr**2
+            # This can be written as
+            # a * tau^2 + b * tau + c, with a, b, c given below
+            diff = delta_gn - delta_sd
+            a = TrustRegion._squared_norm(diff)
+            b = (2 * delta_sd * diff).sum(dim=1, keepdim=True)
+            c = delta_sd_norm_2 - trust_region_2
+            disc = ((b**2) - 4 * a * c).clamp(0.0)
+            # By taking min(tau, 1), this also covers the case when ||d_gn|| < TR
+            tau = ((-b + disc.sqrt()) / (2 * a + Dogleg.EPS)).minimum(disc.new_ones(1))
```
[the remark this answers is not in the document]
fantaosha: Also `(disc + Dogleg.EPS).sqrt()` to avoid `NaN` error for the backward around zero.

fantaosha, line 101: `(disc + Dogleg.EPS).sqrt()` to avoid `NaN` error for the backward around zero.

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,121 @@
...
+            delta_dogleg = torch.where(
+                sd_within_region_idx,
+                delta_sd + tau * diff,
+                delta_dogleg,
+            )
+
+        # Finally, when the steepest descent direction is too close to zero, just use
+        # the Gauss-Newton direction truncated at the trust region, to avoid
+        # numerical errors.
+        if not not_near_zero_sd_idx.all():
```
fantaosha, line 111: What will happen if `||delta_gn||<self._trust_region`? Do you have a reference for this step?

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,167 @@
...
+        self, previous_error: torch.Tensor, delta: torch.Tensor
+    ) -> torch.Tensor:
+        linearization = self.linear_solver.linearization
+
+        # Note that B = |A^t@ A| so that p^tBp := delta^t A^t A delta = ||Adelta||^2
+        Adelta = linearization.Av(delta)
+        grad = -linearization.Atb.squeeze(2)
+        delta_dot_grad = (delta * grad).sum(dim=1)
+
+        return (
```
fantaosha, line 124: Looks like the predicted error does not apply to LM method

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,167 @@
...
+    def compute_delta(self, **kwargs) -> torch.Tensor:
+        # Storing the indices at the trust boundary so that we can update the
+        # trust region inside self.complete_step()
+        delta, self._trusted_step_idx = self._compute_delta_impl()
+        if self._trusted_step_idx is None:
+            self._trusted_step_idx = (
+                TrustRegion._squared_norm(delta) <= self._trust_region**2
+            )
+
+        # Storing the indices of delta exactly at the trust boundary.
```
fantaosha, line 98: Not sure if we need to store the indices of delta exactly at the trust boundary since the expansion and shrinking of the trust region usually only depends on whether the current linearized model gives a good approximation.

theseus/optimizer/nonlinear/dogleg.py
fantaosha: Why not use `clamp(1.0)`?

fantaosha (changes requested):
Only minor comments. Feel free to merge after fixing the `(disc + Dogleg.EPS).sqrt()` issue.

theseus/optimizer/nonlinear/dogleg.py
luisenp: Yes. Oh, this reminds, me I was going to rename all of these variables. `delta_sd` --> `delta_c` and `t` --> `caucy_step_size` (and also rename the norms accordingly).

theseus/optimizer/nonlinear/dogleg.py
luisenp: I tried that first and didn't work as expected, but I just realized that the right call should have been `torch.clamp(max=1.0)`. I'll fix.

luisenp: I don't think we need to add `EPS` because I'm clamping to zero in the line above. Adding EPS instead of clamping wasn't always working because in some cases |disc| > EPS, so we would still get nans. 

This code probably introduces some amount of error, but during testing I had an assert that norm(delta_dogleg) = TR for `~good_gn_idx` which I ran in our unit tests and also in PGO and never failed (tol of 1e-4), so I think this seems decent enough.

theseus/optimizer/nonlinear/dogleg.py
luisenp: Mmmm, I actually think this is a bug. There was a line I removed that filtered for that case so that we would choose the GN step. Maybe we don't need this at all if we do the right thing when computing `tau`. I'll take a look.

theseus/optimizer/nonlinear/trust_region.py
luisenp: Yeah, I noticed it's a bit different in the current code. Will need to look into how to unify these later.

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,167 @@
...
+                "Values must satisfy <accept/shrink>_threshold < expand_threshold, "
+                "shrink_ratio in (0, 1], and expand_ratio > 1.0."
+            )
+        pred_err = self._predicted_error(previous_err, delta)
+        rho = ((previous_err - new_err) / (previous_err - pred_err)).view(-1, 1)
+        shrink_idx = rho < shrink_threshold
+        self._trust_region = torch.where(
+            shrink_idx, self._trust_region * shrink_ratio, self._trust_region
+        )
+        expand_idx = (rho > expand_threshold) & self._at_trust_boundary_idx
```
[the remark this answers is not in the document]
luisenp: I'm following Nocedal (see below). I assumed the rationale for this was to be conservative, since your predicted error might be in agreement with the actual error just because the step was actually smaller than the trust region. 

image

theseus/optimizer/nonlinear/trust_region.py
luisenp: See comment below.

luisenp (commented):
Left some responses. I also think you caught a bug :)

theseus/optimizer/nonlinear/dogleg.py
fantaosha: I mean `disc.sqrt()` in line 101. What will happen to backward if `disc = 0` even after clamping?

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,106 @@
...
+
+        linearization = self.linear_solver.linearization
+
+        delta_sd = linearization.Atb.squeeze(2)
+        Adelta_sd = linearization.Av(delta_sd)
+        Adelta_sd_norm_2 = TrustRegion._squared_norm(Adelta_sd)
+        grad_norm_2 = TrustRegion._squared_norm(delta_sd)
+        cauchy_step_size = grad_norm_2 / (Adelta_sd_norm_2 + Dogleg.EPS)
+        delta_c = delta_sd * cauchy_step_size
+        delta_c_norm_2 = TrustRegion._squared_norm(delta_c)
```
fantaosha, line 71: Maybe simplify the computation as `delta_c_norm_2 = grad_norm_2*Cauchy_step_size**2`?

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,147 @@
...
+        objective: Objective,
+        linear_solver_cls: Optional[Type[LinearSolver]] = None,
+        vectorize: bool = False,
+        linearization_cls: Optional[Type[Linearization]] = None,
+        linearization_kwargs: Optional[Dict[str, Any]] = None,
+        linear_solver_kwargs: Optional[Dict[str, Any]] = None,
+        abs_err_tolerance: float = 1e-10,
+        rel_err_tolerance: float = 1e-8,
+        max_iterations: int = 20,
+        step_size: float = 1.0,
```
fantaosha, line 50: Add `self.max_trust_region` and `self.min_trust_region` here.

theseus/optimizer/nonlinear/dogleg.py
luisenp: Good catch!

theseus/optimizer/nonlinear/dogleg.py
```diff
@@ -0,0 +1,106 @@
...
+        # All Gauss-Newton step are within trust-region, can return
+        if (delta_gn_norm_2 < trust_region_2).all():
+            # Return a False mask since no indices are at the boundary
+            return delta_gn
+
+        # ---------------------------------------------------------------------
+        # Some Gauss-Newton steps are outside trust-region,
+        # need to compute dogleg step
+        # ---------------------------------------------------------------------
+
```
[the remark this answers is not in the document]
luisenp: Removed.

theseus/optimizer/nonlinear/trust_region.py
luisenp: Added.

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,147 @@
...
+# Iff the ratio between the actual error reduction after the step and the
+# predicted one is greater than `accept_threshold`, the step is applied.
+# Iff it's also greater than `expand_threshold`, the trust region is also
+# increased by `expand_ratio`.
+# Iff the ratio is lower than `shrink_threshold`,
+# then the trust region is reduced by `shrink_ratio`.
+# The trust region is initialized to `trust_region_init`.
+class TrustRegion(NonlinearLeastSquares, abc.ABC):
+    _MIN_TRUST_REGION = 1.0e-5
+    _MAX_TRUST_REGION = 1.0e5
```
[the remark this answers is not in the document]
luisenp: Changed.

theseus/optimizer/sparse_linearization.py
```diff
@@ -152,12 +158,33 @@ def _atb_impl(self) -> torch.Tensor:
...
-                self.A_val,
-                self.b,
+                self.A_val.double(),
+                self.b.double(),
             ).unsqueeze(2)
-        return self._Atb
+        return self._Atb.to(dtype=self.A_val.dtype)
+
+    def Av(self, v: torch.Tensor) -> torch.Tensor:
+        A_row_ptr = torch.tensor(self.A_row_ptr, dtype=torch.int32).to(
```
fantaosha, line 171: Why `dtype=torch.int32`? How about `self.A_row_ptr.clone()` or `v.new_tensor(self.A_row_ptr)`?

theseus/optimizer/sparse_linearization.py
```diff
@@ -152,12 +158,33 @@ def _atb_impl(self) -> torch.Tensor:
...
             # unsqueeze at the end for consistency with DenseLinearization
-            self._Atb = tmat_vec(
-                self.objective.batch_size,
+            self._Atb = sparse_mtv(
                 self.num_cols,
                 A_row_ptr,
                 A_col_ind,
-                self.A_val,
-                self.b,
+                self.A_val.double(),
```
fantaosha, line 165: So our sparse solvers only supports `float64`?

theseus/utils/sparse_matrix_utils.py
```diff
@@ -97,6 +99,97 @@ def tmat_vec(
...
+        start = A_row_ptr[row]
+        end = A_row_ptr[row + 1]
+        columns = A_col_ind[start:end].long()
+        if is_tmat:
+            A_grad[:, start:end] = v[:, row].view(-1, 1) * grad_output[:, columns]
+            v_grad[:, row] = (grad_output[:, columns] * A_val[:, start:end]).sum(dim=1)
+        else:
+            A_grad[:, start:end] = v[:, columns] * grad_output[:, row].view(-1, 1)
+            v_grad[:, columns] += grad_output[:, row].view(-1, 1) * A_val[:, start:end]
+    return A_grad, v_grad
```
fantaosha, line 140: Please make sure `A_grad = grad_output @ v^T` or `v_grad = A^T @ grad_output`.

theseus/utils/sparse_matrix_utils.py
```diff
@@ -97,6 +99,97 @@ def tmat_vec(
...
+class _SparseMvPAutograd(torch.autograd.Function):
+    @staticmethod
+    def forward(  # type: ignore
+        ctx: Any,
+        num_cols: int,
+        A_row_ptr: torch.Tensor,
+        A_col_ind: torch.Tensor,
+        A_val: torch.Tensor,
+        v: torch.Tensor,
+    ) -> torch.Tensor:
```
fantaosha, line 152: Add `ctx.mark_non_differentiable(num_cols)`, `ctx.mark_non_differentiable(A_row_ptr)`, and `ctx.mark_non_differentiable(A_col_ind)`.

theseus/optimizer/nonlinear/trust_region.py
fantaosha: I'm still a little confused. Where the `self.max_trust_region` and `self.min_trust_region`. Also I wonder if we need `self.max_trust_region` since it does not matter if the `trust_reigon` is very large as long as the linearized model is correct.

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,146 @@
...
+                "Invalid parameters for TrustRegionMethod. "
+                "Values must satisfy <accept/shrink>_threshold < expand_threshold, "
+                "shrink_ratio in (0, 1], and expand_ratio > 1.0."
+            )
+        pred_err = self._predicted_error(previous_err, delta)
+        rho = ((previous_err - new_err) / (previous_err - pred_err)).view(-1, 1)
+        shrink_idx = rho < shrink_threshold
+        self._trust_region = torch.where(
+            shrink_idx, self._trust_region * shrink_ratio, self._trust_region
+        )
```
fantaosha, line 138: Should we add a stopping condition here if `self._trust_region` is too small?

fantaosha: Maybe we need to add another vector `stopped` to indicate whether an optimization problem is still active or not.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
```diff
@@ -46,22 +46,23 @@ def forward(  # type: ignore
```
fantaosha, line 46: Add `ctx.make_non_differentiable()` here?

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
```diff
@@ -46,22 +46,23 @@ def forward(  # type: ignore
...
         x = Atb.clone()
         solver_context.solve(x)  # solve in place
 
         ctx.b = b
         ctx.x = x
-        ctx.A_val = A_val
+        ctx.A_val_double = A_val_double
         ctx.A_row_ptr = A_row_ptr
         ctx.A_col_ind = A_col_ind
         ctx.sparse_structure = sparse_structure
```
fantaosha, line 68: If I remember correctly, it is better to use `ctx.save_for_backward()`.

theseus/optimizer/autograd/baspacho_sparse_autograd.py
```diff
@@ -34,27 +34,30 @@ def forward(  # type: ignore
         numeric_decomposition = symbolic_decomposition.create_numeric_decomposition(
```
fantaosha, line 34: Add `ctx.mark_non_differentiable()` here?

fantaosha (changes requested):
Some minor changes requested.

theseus/optimizer/sparse_linearization.py
luisenp: We can ignore this, this is already merged into main. It was showing up because I hadn't rebased yet.

theseus/optimizer/sparse_linearization.py
luisenp: That's correct. This part of the code is also already merged.

theseus/utils/sparse_matrix_utils.py
luisenp: This part is also already merged. The expression above encodes that matrix product (but for our sparse data structure), and we have gradcheck for this in our unit tests.

theseus/utils/sparse_matrix_utils.py
luisenp: Oh, good point. I can add this in a separate PR. It doesn't belong to Dogleg feature.

theseus/optimizer/autograd/lu_cuda_sparse_autograd.py
luisenp: This is very old code that Maurizio added. I'm going to refactor this in a future PR.

theseus/utils/sparse_matrix_utils.py
luisenp: Ah, sorry. I think `mark_non_differentiable` is meant for outputs, not for inputs. See [docs](https://pytorch.org/docs/stable/generated/torch.autograd.function.FunctionCtx.mark_non_differentiable.html).

theseus/optimizer/nonlinear/trust_region.py
luisenp: I think our current `converged_indices` is functionally equivalent to `stopped`.

theseus/utils/sparse_matrix_utils.py
fantaosha: Ah, correct.

fantaosha (approved):
LGTM. Feel free to merge.

tests/optimizer/nonlinear/test_trust_region.py
```diff
@@ -0,0 +1,29 @@
...
+import torch
+
+
+@pytest.mark.parametrize("dof", [1, 8])
+def test_rho(dof):
+    # This test checks that rho = 1 for a simple linear problem min (x - t) ** 2
+    x = th.Vector(tensor=torch.zeros(1, dof), name="x")
+    t = th.Vector(tensor=torch.ones(1, dof), name="t")
+    o = th.Objective()
+    o.add(th.Difference(x, t, th.ScaleCostWeight(1.0), name="cf"))
```
fantaosha, line 13: Can we add multiple factors with different `ScaleCostWeight`? In the future, we might create a factor to be something like `A*X-b`?

fantaosha (approved):
LGTM only one minor comment.

tests/optimizer/nonlinear/test_trust_region.py
```diff
@@ -0,0 +1,29 @@
...
+    # This is testing TrustRegion base class rather than Dogleg's
+    # Using Dogleg because it's the only subclass of TrustRegion atm
+    opt = th.Dogleg(o)
+    o._resolve_batch_size()
+    opt.linear_solver.linearization.linearize()
+    previous_err = opt._error_metric()
+
+    # Check rho = 1. Predicted error by TrustRegion method should
+    # match actual error after step for a linear problem
+    for _ in range(10):  # repeat a few times with random delta
```
fantaosha, line 23: Not sure if it is better to first compute `d_gn` direction and then random scale the trust region with `scale * \|d_gn\|`

tests/optimizer/nonlinear/test_trust_region.py
```diff
@@ -0,0 +1,58 @@
...
+import theseus as th
+import torch
+
+
+# Passing different linear_solver_cls to test with both dense/sparse linearization
+# Also test sparse [cpu/gpu] because they go through different custom backends
+@pytest.mark.parametrize("dof", [1, 8])
+@pytest.mark.parametrize(
+    "linear_solver_cls",
+    [th.CholeskyDenseSolver, th.CholmodSparseSolver, th.LUCudaSparseSolver],
```
fantaosha, line 12: Do we need to test `BaSpaCho`?

tests/optimizer/nonlinear/test_trust_region.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # Check rho = 1. Predicted error by TrustRegion method should
+        # match actual error after step for a linear problem
+        for _ in range(100):
+            delta = torch.randn(1, dof * n_vars, device=device, generator=rng)
+            _, new_err = opt._compute_retracted_tensors_and_error(
+                delta, torch.zeros_like(delta[:, 0]), False
+            )
+            rho = opt._compute_rho(delta, previous_err, new_err)
+            torch.testing.assert_close(rho, torch.ones_like(rho), atol=1e-3, rtol=1e-3)
```
fantaosha, line 58: Wondering if we could use even smaller `atol` than `1e-3`. I guess `1e-6` should also work.

fantaosha (approved):
LGTM. Feel free to merge.

tests/optimizer/nonlinear/test_trust_region.py
luisenp: No need for `Baspacho` since it uses the same sparse linearization class.

tests/optimizer/nonlinear/test_trust_region.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
mhmukadam, line 1: License header missing.

theseus/optimizer/nonlinear/trust_region.py
```diff
@@ -0,0 +1,151 @@
...
+    # Return the computed delta
+    @abc.abstractmethod
+    def _compute_delta_impl(self) -> torch.Tensor:
+        pass
+
+    def compute_delta(self, **kwargs) -> torch.Tensor:
+        return self._compute_delta_impl()
+
+    @staticmethod
+    def _squared_norm(tensor: torch.Tensor, keepdim: bool = True) -> torch.Tensor:
```
mhmukadam, line 87: This does not seem specific to TrustRegion or to Optimizer, maybe add to utils?

mhmukadam (commented):
Nice to see this feature land, great work @luisenp on the PR and @fantaosha on the multiple rounds of thorough reviews.

PR #372, opened 2022-11-22
Added missing square in `test_theseus_layer` loss.

luisenp:
This PR adds a few small fixes that I found when working on #371. Somehow none of us noticed that the loss in `test_theseus_layer` was actually missing a square! I also removed a test that's no longer needed for LM.

### DIFF
--- a/tests/test_theseus_layer.py
+++ b/tests/test_theseus_layer.py
@@ -191,7 +191,7 @@ def _run_optimizer_test(
     optimizer_kwargs,
     cost_weight_model,
     use_learnable_error=False,
-    verbose=True,
+    verbose=False,
     learning_method="default",
     force_vectorization=False,
 ):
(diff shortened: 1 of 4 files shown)


mhmukadam:
Good catch on
> the loss in test_theseus_layer was actually missing a square

I think @dishank-b also surfaced this bug in his CEM PR so he may want to rebase with this fix once it's merged.

tests/test_theseus_layer.py
```diff
@@ -353,12 +353,13 @@ def cost_weight_fn():
...
                 cost_weight_fn(), torch.zeros((1, num_points), device=device)
             )
-            loss = (cost_gt - cost_opt) + 10.0 * l2_reg
+            loss = (cost_gt - cost_opt) ** 2 + 10.0 * l2_reg
             loss = torch.mean(loss, dim=0)
         else:
             loss = mse_loss
 
         loss.backward()
+        print("Loss: ", loss.item())
```
mhmukadam, line 362: Does this print need to be inside `if verbose:`?

mhmukadam (approved):

tests/test_theseus_layer.py
luisenp: I think it's OK to leave as is, easier to see what's going on if it ever fails in CI. `verbose=True` was printing all the optimizer iterations which was way too much stuff.

PR #373, opened 2022-11-23
Add theseus.geometry.functional.so3.adjoint()

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/tests/geometry/functional/test_so3.py
+++ b/tests/geometry/functional/test_so3.py
@@ -18,3 +18,13 @@ def test_exp(batch_size: int, dtype: torch.dtype):
     rng.manual_seed(0)
     tangent_vector = torch.rand(batch_size, 3, dtype=dtype, generator=rng)
     check_lie_group_function(so3, "exp", TEST_EPS, tangent_vector)
+
+
+@pytest.mark.parametrize("batch_size", [1, 20, 100])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
+def test_adjoint(batch_size: int, dtype: torch.dtype):
+    rng = torch.Generator()
+    rng.manual_seed(0)
+    tangent_vector = torch.rand(batch_size, 3, dtype=dtype, generator=rng)
+    group = so3.exp(tangent_vector)
+    check_lie_group_function(so3, "adjoint", TEST_EPS, group)
--- a/theseus/geometry/functional/lie_group.py
+++ b/theseus/geometry/functional/lie_group.py
@@ -44,7 +44,7 @@ def UnaryOperatorFactory(module, op_name):
         return op, jop
     else:
 
-        def op_no_jacobians(input: torch.Tensor) -> torch.Tensor:
+        def op_no_jop(input: torch.Tensor) -> torch.Tensor:
             return op_autograd_fn(input)
 
-        return op_no_jacobians
+        return op_no_jop
--- a/theseus/geometry/functional/lie_group.py
+++ b/theseus/geometry/functional/lie_group.py
@@ -27,16 +27,24 @@ def UnaryOperatorFactory(module, op_name):
     op_autograd_fn = getattr(module, "_" + op_name + "_autograd_fn")
     jop_autograd_fn = getattr(module, "_j" + op_name + "_autograd_fn")
 
-    def op(
-        input: torch.Tensor,
-        jacobians: Optional[List[torch.Tensor]] = None,
-    ) -> torch.Tensor:
-        if jacobians is not None:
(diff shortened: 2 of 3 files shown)


theseus/geometry/functional/lie_group.py
```diff
@@ -27,16 +27,24 @@ def UnaryOperatorFactory(module, op_name):
     op_autograd_fn = getattr(module, "_" + op_name + "_autograd_fn")
```
luisenp, line 27: Let's add some comments at the top of this class describing what it does.

luisenp (commented):
LGTM! One minor comment.

theseus/geometry/functional/lie_group.py
fantaosha: I added the comments in `taoshaf.functional.so3.inverse`.

fantaosha: > Let's add some comments at the top of this class describing what it does.

I added the comments in [`taoshaf.functional.so3.inverse`](https://github.com/facebookresearch/theseus/blob/8f01664ba4a178169250b94062df4cc7e429b465/theseus/geometry/functional/lie_group.py#L34).

tests/geometry/functional/test_so3.py
```diff
@@ -18,3 +18,13 @@ def test_exp(batch_size: int, dtype: torch.dtype):
...
+
+
+@pytest.mark.parametrize("batch_size", [1, 20, 100])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
+def test_adjoint(batch_size: int, dtype: torch.dtype):
+    rng = torch.Generator()
+    rng.manual_seed(0)
+    tangent_vector = torch.rand(batch_size, 3, dtype=dtype, generator=rng)
+    group = so3.exp(tangent_vector)
+    check_lie_group_function(so3, "adjoint", TEST_EPS, group)
```
mhmukadam, line 30: This logic is same as above. Could we instead just do
`@pytest.mark.parametrize("op", ["exp", "adjoint"])`

theseus/geometry/functional/lie_group.py
```diff
@@ -27,16 +27,24 @@ def UnaryOperatorFactory(module, op_name):
...
+        def jop(input: torch.Tensor) -> Tuple[List[torch.Tensor], torch.Tensor]:
+            return jop_autograd_fn(input)
+
+        return op, jop
+    else:
+
+        def op_no_jop(input: torch.Tensor) -> torch.Tensor:
+            return op_autograd_fn(input)
+
+        return op_no_jop
```
mhmukadam, line 50: I am guessing this block will be removed once we have added `_jxxx_autograd_fn`?

mhmukadam (approved):

tests/geometry/functional/test_so3.py
luisenp: Good catch! Let's definitely do this. You can call it `test_unary_operator`.

mhmukadam: Looking at inverse I realized there is an extra line there calling exp to get group from tangent. So not sure if we can do this for all unary ops.

theseus/geometry/functional/lie_group.py
fantaosha: No, the jacobian of `adjoint` transformation has no meaning and is not well defined. Despite that, we still provide analytic backward.

tests/geometry/functional/test_so3.py
fantaosha: No, we can not do this for `adjoint` and `exp` since the inputs are different.

theseus/geometry/functional/lie_group.py
mhmukadam: Ah okay, it would useful to add this in comments so it's clear it's not something we will be adding later like the autograd for some other jacobians.

tests/geometry/functional/test_so3.py
luisenp: We can add an if to handle ops that require group inputs, I think that would still be preferable. For example,
```python
if op_name in ["adjoint", ..., ]:
     op_input = so3.exp(tangent_vector)
else:
     op_input = tangent_vector
check_lie_group_function(so3, op_name, TEST_EPS, op_input)
```

fantaosha: I will consider if doing this in the future since I'm planning to add more tests for `test_inverse` and `test_adjoint` which seem need specific tests like what we did in `theseus/geometry/lie_group`. But I created #377 for reminder.

PR #381, opened 2022-11-26
Add rand() and randn() to theseus.geometry.functional.so3

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/tests/geometry/functional/test_so3.py
+++ b/tests/geometry/functional/test_so3.py
@@ -4,7 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import pytest
-from .common import check_lie_group_function
+from .common import check_lie_group_function, left_project_func
 from theseus.geometry.functional.constants import TEST_EPS
 import theseus.geometry.functional.so3 as so3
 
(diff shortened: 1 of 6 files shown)


luisenp (approved):
LGTM.

tests/geometry/functional/test_so3.py
```diff
@@ -27,8 +27,7 @@ def test_exp(batch_size: int, dtype: torch.dtype):
 def test_adjoint(batch_size: int, dtype: torch.dtype):
     rng = torch.Generator()
     rng.manual_seed(0)
-    tangent_vector = torch.rand(batch_size, 3, dtype=dtype, generator=rng)
-    group = so3.exp(tangent_vector)
+    group = so3.rand(batch_size, generator=rng, dtype=dtype)
```
mhmukadam, line 30: We are testing `rand` indirectly as part of other tests which might be okay, but none of the tests use `randn` so it's currently not tested in CI.

mhmukadam (approved):
LGTM aside from one comment.

tests/geometry/functional/test_so3.py
fantaosha: `rand()` are tested offline. The only difference is how to generate the tangent vector. So I think it is is safe.

mhmukadam: Offline testing is always nice. I am wondering though whether it is good to add to the unit tests as well so that CI checks it regularly.

PR #387, opened 2022-11-29
Made torch's CUDA_GCC_VERSION check optional in setup.py.

### LINKED ISSUE #376: module 'torch.utils.cpp_extension' has no attribute 'CUDA_GCC_VERSIONS'
## ❓ Questions and Help
Hi author~ I follow your installation step :
1. I create conda env with python3.8 + torch1.9 + cu111 (because the original cuda installed in my machine is 11.1 by running **nvcc -V**)
2. Then I choose to build from source ：
image
but the following error occurs:
image

it seems like pytorch version is not right? So I upgrade to **torch 1.11 + cu113**，but the same error is still reported.

I would like to ask what the possible reason is and can you give some advice？ I'll thanks for your reply~

luisenp:
Closes #376

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,11 +19,12 @@ try:
     #   -ltorch_cuda_cu -ltorch_cuda_cpp
     torch_cpp_ext.BUILD_SPLIT_CUDA = False
 
-    # hack to be able to compile with gcc-8.4.0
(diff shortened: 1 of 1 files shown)


maurimo (approved):
LGTM!

mhmukadam (commented):
LGTM!

PR #388, opened 2022-11-30
Add a workaround for NonlinearOptimizer rejecting all batch steps

luisenp:
In this version we sometimes have to recompute the error norm if some steps are rejected, so optimizing error vectorization becomes important again. Will fix in a separate PR.

### DIFF
--- a/theseus/optimizer/nonlinear/levenberg_marquardt.py
+++ b/theseus/optimizer/nonlinear/levenberg_marquardt.py
@@ -96,7 +96,8 @@ class LevenbergMarquardt(NonlinearLeastSquares):
         if adaptive_damping and not self._allows_adaptive:
             raise NotImplementedError(
                 f"Adaptive damping is only supported by solvers with type "
-                f"[{_ADAPT_DAMP_SOLVERS_STR}]."
+                f"[{_ADAPT_DAMP_SOLVERS_STR}], but got solver of type "
+                f"{type(self.linear_solver)}."
             )
         if adaptive_damping:
             self._damping = damping * torch.ones(
--- a/theseus/optimizer/nonlinear/levenberg_marquardt.py
+++ b/theseus/optimizer/nonlinear/levenberg_marquardt.py
@@ -116,11 +116,17 @@ class LevenbergMarquardt(NonlinearLeastSquares):
         damping_eps: Optional[float] = None,
         **kwargs,
     ) -> torch.Tensor:
-        if ellipsoidal_damping and not self._allows_ellipsoidal:
-            raise NotImplementedError(
-                f"Ellipsoidal damping is only supported by solvers with type "
-                f"[{_EDAMP_SOLVERS_STR}]."
-            )
+        if ellipsoidal_damping:
+            if not self._allows_ellipsoidal:
+                raise NotImplementedError(
+                    f"Ellipsoidal damping is only supported by solvers with type "
+                    f"[{_EDAMP_SOLVERS_STR}]."
+                )
+            if isinstance(self._damping, torch.Tensor):
+                raise NotImplementedError(
+                    "Adaptive damping is not currently supported together with "
+                    "ellipsoidal_damping."
+                )
         if damping_eps is not None and not self._allows_ellipsoidal:
             raise NotImplementedError(
                 f"damping eps is only supported by solvers with type "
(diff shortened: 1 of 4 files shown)


theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -518,13 +513,43 @@ def _step(
...
+        tensor_dict, err = self._compute_retracted_tensors_and_error(
+            delta, converged_indices, force_update
+        )
         reject_indices = self._complete_step(delta, err, previous_err, **kwargs)
-        self.objective.update(tensor_map, batch_ignore_mask=reject_indices)
 
+        if reject_indices is not None and reject_indices.all():
+            # If all steps are marked as rejected, optimizer will think convergence
+            # has happened and stop. Taking a small step prevents this issue
+            tensor_dict, err = self._compute_retracted_tensors_and_error(
```
fantaosha, line 542: Not sure if the logic is reasonable. According to my experience, a small step might not be sufficient for the optimizer to jump out of the bad trap.

fantaosha (commented):
LGTM. Please make sure the predicted error for LM and dogleg is also divided by 2.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: I was afraid that an all reject would somehow break the compute graph, and I also wanted to keep things very simple. But I realized that there is an easy way to do this and it can be implemented w/o affecting backward pass. I implemented it in a8d3dc74158097e93c6783a0cb94e78017cc6467. Let me know if this looks OK.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -518,13 +513,43 @@ def _step(
             ignore_mask=converged_indices,
             force_update=force_update,
         )
-        tensor_map = {v.name: v.tensor for v in self._tmp_optim_vars}
+        tensor_dict = {v.name: v.tensor for v in self._tmp_optim_vars}
         with torch.no_grad():
-            err = self.objective.error_squared_norm(tensor_map, also_update=False)
+            err = self.objective.error_squared_norm(tensor_dict, also_update=False) / 2
```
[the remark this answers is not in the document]
luisenp: I added method to avoid having to manually divide by 2 everywhere.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: I also checked manually that the current end-to-end test in `test_theseus_layer` hits the all reject case, so backward pass does to work with this logic.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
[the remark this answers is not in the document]
fantaosha: Do you mean the predicted error from linearization? If so, feel free to merge.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: I mean this b80977dc72c3d2cd45efd31c35760a989cd96b4a

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -115,11 +116,17 @@ def compute_delta(
...
-                f"Ellipsoidal damping is only supported by solvers with type "
-                f"[{_EDAMP_SOLVERS_STR}]."
-            )
+        if ellipsoidal_damping:
+            if not self._allows_ellipsoidal:
+                raise NotImplementedError(
+                    f"Ellipsoidal damping is only supported by solvers with type "
+                    f"[{_EDAMP_SOLVERS_STR}]."
+                )
+            if isinstance(self._damping, torch.Tensor):
```
fantaosha, line 125: Do we need to disable `epplipsodial_damping` for `adaptive_damping=False`? Looks like the computation of `den` is incorrect if diagonal scaling is used.

fantaosha (approved):
LGTM. Just one minor comment.

mhmukadam (commented):
LGTM!

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -107,6 +107,8 @@ def resolve(key: Union[str, "BackwardMode"]) -> "BackwardMode":
 #                ignoring indices given by `reject_indices`
 #    3. Check convergence
 class NonlinearOptimizer(Optimizer, abc.ABC):
+    _MAX_ALL_REJECT_ATTEMPTS = 3
```
mhmukadam, line 110: Any tradeoff or overhead to consider here? Curious about the choice of `3` attempts. Are these all kept in the compute graph?

Also might be helpful to add in a comment that these max attempts are per iteration and not overall.

PR #389, opened 2022-11-30
Remove softmax from end to end test and do some clean up

luisenp:


tests/test_theseus_layer.py
```diff
@@ -383,14 +386,14 @@ def test_backward(
     cost_weight_model,
     learning_method,
 ):
-    optim_kwargs = {} if nonlinear_optim_cls == th.GaussNewton else {"damping": 0.01}
+    optim_kwargs = {
+        th.GaussNewton: {},
+        th.LevenbergMarquardt: {"damping": 0.01},
+    }[nonlinear_optim_cls]
```
mhmukadam, line 392: Ah in-place dict is a neat way to avoid if block. Any particular reason besides the elegance :D ?

mhmukadam (approved):

tests/test_theseus_layer.py
luisenp: Not that I know. I saw Olivier using this in another project and I also liked how clean it looks, so I appropriated it :P

luisenp: In an upcoming PR I'll have 3 cases here, so the change would be more appropriate then.

PR #392, opened 2022-12-01
Add a differentiable sparse matrix vector product on top of our ops

luisenp:
Backward pass can be made more efficient in GPU if we write a custom CUDA kernel for it, but this should be reasonable enough for now.


luisenp:
> Looks great! happy to know we now have this, it can be used for instance to use iterative solvers (not sure if that was already the plan).

I want to add at least Conjugate Gradient at some point, but there is only so many hours in the day :) Will probably do eventually.

mhmukadam (commented):
LGTM!

PR #393, opened 2022-12-02
Add diagonal scaling method and re-enable LM adaptive + ellipsoidal

luisenp:
This adds a linearization diagonal scaling method and implementations for dense and sparse linearization. The sparse implementation could be faster with a custom kernel, to avoid python loop over rows, but this should be good enough to get an initial implementation.

### DIFF
--- a/theseus/optimizer/dense_linearization.py
+++ b/theseus/optimizer/dense_linearization.py
@@ -72,3 +72,6 @@ class DenseLinearization(Linearization):
 
     def Av(self, v: torch.Tensor) -> torch.Tensor:
         return self.A.bmm(v.unsqueeze(2)).squeeze(2)
+
+    def diagonal_scaling(self, v: torch.Tensor) -> torch.Tensor:
+        return v * self._AtA.diagonal(dim1=1, dim2=2)
--- a/theseus/optimizer/nonlinear/levenberg_marquardt.py
+++ b/theseus/optimizer/nonlinear/levenberg_marquardt.py
@@ -123,11 +123,6 @@ class LevenbergMarquardt(NonlinearLeastSquares):
                     f"Ellipsoidal damping is only supported by solvers with type "
                     f"[{_EDAMP_SOLVERS_STR}]."
                 )
-            if isinstance(self._damping, torch.Tensor):
-                raise NotImplementedError(
-                    "Adaptive damping is not currently supported together with "
-                    "ellipsoidal_damping."
-                )
         if damping_eps is not None and not self._allows_ellipsoidal:
             raise NotImplementedError(
                 f"damping eps is only supported by solvers with type "
@@ -150,6 +145,7 @@ class LevenbergMarquardt(NonlinearLeastSquares):
         down_damping_ratio: float = 9.0,
         up_damping_ratio: float = 11.0,
         damping_accept: float = 0.1,
+        ellipsoidal_damping: bool = False,
         **kwargs,
     ) -> Optional[torch.Tensor]:
         if adaptive_damping:
@@ -160,6 +156,7 @@ class LevenbergMarquardt(NonlinearLeastSquares):
                 damping_accept,
                 down_damping_ratio,
                 up_damping_ratio,
+                ellipsoidal_damping,
             )
         else:
             return None
@@ -179,13 +176,13 @@ class LevenbergMarquardt(NonlinearLeastSquares):
         damping_accept: float,
         down_damping_ratio: float,
         up_damping_ratio: float,
+        ellipsoidal_damping: bool,
     ) -> torch.Tensor:
         linearization = self.linear_solver.linearization
-        damping = (
-            self._damping.view(-1, 1)
-            if isinstance(self._damping, torch.Tensor)
-            else self._damping
-        )
+        assert isinstance(self._damping, torch.Tensor)
+        damping = self._damping.view(-1, 1)
+        if ellipsoidal_damping:
+            damping = linearization.diagonal_scaling(damping)
         # Deliberately using Atb before updating the variables, according to
         # the LM reference above
         den = (delta * (damping * delta + linearization.Atb.squeeze(2))).sum(dim=1) / 2
--- a/theseus/optimizer/sparse_linearization.py
+++ b/theseus/optimizer/sparse_linearization.py
@@ -169,3 +169,15 @@ class SparseLinearization(Linearization):
         return sparse_mv(
             self.num_cols, A_row_ptr, A_col_ind, self.A_val.double(), v.double()
         ).to(v.dtype)
+
+    def diagonal_scaling(self, v: torch.Tensor) -> torch.Tensor:
+        assert v.ndim == 2
+        assert v.shape[1] == self.num_cols
+        A_val = self.A_val
+        diag = torch.zeros(A_val.shape[0], self.num_cols)
+        for row in range(self.num_rows):
+            start = self.A_row_ptr[row]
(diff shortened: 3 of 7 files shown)


theseus/optimizer/sparse_linearization.py
```diff
@@ -174,3 +174,15 @@ def Av(self, v: torch.Tensor) -> torch.Tensor:
...
+    def diagonal_scaling(self, v: torch.Tensor) -> torch.Tensor:
+        assert v.ndim == 2
+        assert v.shape[1] == self.num_cols
+        A_val = self.A_val
+        diag = torch.zeros(A_val.shape[0], self.num_cols)
+        for row in range(self.num_rows):
+            start = self.A_row_ptr[row]
+            end = self.A_row_ptr[row + 1]
+            columns = self.A_col_ind[start:end]
+            diag[:, columns] += A_val[:, start:end] ** 2
```
fantaosha, line 187: Should we cache `diag` since it might be used multiple times if the current step is rejected?

theseus/optimizer/nonlinear/levenberg_marquardt.py
```diff
@@ -178,13 +175,13 @@ def _check_accept(
...
-            if isinstance(self._damping, torch.Tensor)
-            else self._damping
-        )
+        assert isinstance(self._damping, torch.Tensor)
+        damping = self._damping.view(-1, 1)
+        if ellipsoidal_damping:
+            damping = linearization.diagonal_scaling(damping)
         # Deliberately using Atb before updating the variables, according to
         # the LM reference above
         den = (delta * (damping * delta + linearization.Atb.squeeze(2))).sum(dim=1) / 2
```
fantaosha, line 187: In trust region method, `den` is also called `predicted reduction`. In addition, a test can be added for `den` as we discussed before.

theseus/optimizer/dense_linearization.py
```diff
@@ -72,3 +72,6 @@ def _atb_impl(self) -> torch.Tensor:
 
     def Av(self, v: torch.Tensor) -> torch.Tensor:
         return self.A.bmm(v.unsqueeze(2)).squeeze(2)
+
+    def diagonal_scaling(self, v: torch.Tensor) -> torch.Tensor:
+        return v * self._AtA.diagonal(dim1=1, dim2=2)
```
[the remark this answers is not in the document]
fantaosha: How do we add `damping` to `self._AtA`? Is there another variable to cache the hessian matrix?

theseus/optimizer/sparse_linearization.py
fantaosha: I checked the code of `sparse_linearization.py`. It seems that there is no `self._AtA`?

fantaosha (approved):
LGTM. Only some minor comments.

theseus/optimizer/sparse_linearization.py
luisenp: Good idea, I'll add this since the diag will be valid until linearize is run again.

luisenp: Added in 4bbea0f

theseus/optimizer/dense_linearization.py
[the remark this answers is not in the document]
luisenp: The order shouldn't matter here because `self._AtA.diagonal` is also a vector, not a matrix (see attached example), so this is just elementwise product. 

image

We add the damping in the dense solver before solving, [here](https://github.com/facebookresearch/theseus/blob/main/theseus/optimizer/linear/dense_solver.py#L38). We don't currently cache (AtA + \lambda * diag(AtA)), although I don't think we are calling this more than once per iteration.

theseus/optimizer/nonlinear/levenberg_marquardt.py
luisenp: I'll add this to `TrustRegion` class in the Dogleg PR (remind me if I forget). We'll unify these two in a later PR.

luisenp:
@fantaosha That's correct. Out of our 3 sparse solvers, 2 don't need to explicitly construct AtA. The only exception is `LUCudaSparseSolver`, which does it [here](https://github.com/facebookresearch/theseus/blob/main/theseus/optimizer/autograd/lu_cuda_sparse_autograd.py#L49). But note that this class has no access to our linearization structures. Also note that if you try to call `lin.AtA` on a sparse linearization it actually raises `NotImplementedError`. I'd prefer to leave like this for now until we have an actual use for such method. 

On the other hand, your question made me realize that caching the solution of linear solvers could also be useful now that we are allowing retries. I'll add this in a later PR.

theseus/optimizer/sparse_linearization.py
```diff
@@ -86,12 +86,18 @@ def __init__(
...
         # self._linearize_jacobian_impl()
         self._Atb: torch.Tensor = None
 
+        # computed lazily by self.diagonal_scaling() and reset to None by
+        # self._linearize_jacobian_impl()
+        self._AtA_diag: torch.Tensor = None
+
     def _linearize_jacobian_impl(self):
         self._Atb = None
+        self._AtA = None
```
fantaosha, line 100: `self._AtA_diag = None` here? If I remember correctly, `SparseLinearization` will be reused for the same optimization problem and we need to reinitialize `self._AtA_diag` every time.

theseus/optimizer/dense_linearization.py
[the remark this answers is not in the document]
fantaosha: Thanks!

fantaosha (approved):
LGTM. Only one minor comment.

theseus/optimizer/sparse_linearization.py
```diff
@@ -174,3 +180,16 @@ def Av(self, v: torch.Tensor) -> torch.Tensor:
         return sparse_mv(
             self.num_cols, A_row_ptr, A_col_ind, self.A_val.double(), v.double()
         ).to(v.dtype)
+
+    def diagonal_scaling(self, v: torch.Tensor) -> torch.Tensor:
+        assert v.ndim == 2
+        assert v.shape[1] == self.num_cols
+        if self._AtA_diag is None:
```
fantaosha, line 187: I'm wondering if this is a bug. In a batch, some instances might be updated while some remain the same. Therefore, do we need a mask for this?

theseus/optimizer/sparse_linearization.py
luisenp: Oh sorry, that's a typo. Good catch!

theseus/optimizer/sparse_linearization.py
luisenp: I don't think this should cause incorrect results/behavior, but for sure some duplicated computation is possible. And this is true at the linearization level, not just for this method; meaning linearization is run for the full batch every time linearization needs to be called. And, while the `Vectorize` caches jacobians/errors results, as soon as a single batch index is updated for *any* variable, vectorization is recomputed again, which is the largest part of the cost.  

But at least in terms of the backward pass running without issues, this should be OK, since we have been using this masking logic for a long time without spotting any issues. You can think of this as just recomputing the scaling with the tensors at the tip of the compute graph. 

That being said, I think it's interesting to add an issue for considering a more selective linearization logic, it might be worth adding at some point.

theseus/optimizer/sparse_linearization.py
luisenp: Fixed.

mhmukadam (commented):
LGTM!

Issue #395, opened 2022-12-05
Add masked linearization


luisenp:
To account for indices with rejected steps

See
_Originally posted by @luisenp in https://github.com/facebookresearch/theseus/pull/393#discussion_r1038810025_

PR #398, opened 2022-12-08
Add support for masking jacobians of zero weights in the batch

luisenp:
With this feature, the vectorizer is able to compute errors and jacobians only over batch indices that are known to be greater than zero. Preliminary tests on a simple problem indicate savings of ~18% compute time both in CPU and GPU. 

For now I made the feature only applicable inside vectorization, although it might be safe to turn off in general.

### DIFF
--- a/theseus/core/cost_function.py
+++ b/theseus/core/cost_function.py
@@ -46,7 +46,7 @@ def masked_jacobians(
     err = aux_tensor.new_zeros(batch_size, cost_fn.dim())
     with masked_variables(cost_fn_vars, mask):
         masked_jacobians_, err[mask] = cost_fn.jacobians()
-        for (masked_jac, jac) in zip(masked_jacobians_, jacobians):
+        for masked_jac, jac in zip(masked_jacobians_, jacobians):
             jac[mask] = masked_jac
     return jacobians, err
 
--- a/theseus/core/cost_function.py
+++ b/theseus/core/cost_function.py
@@ -104,7 +104,8 @@ class CostFunction(TheseusFunction, abc.ABC):
         done = False
         if self.__supports_masking__:
             try:
-                mask = ~self.weight.is_zero()
+                with torch.no_grad():
+                    mask = ~self.weight.is_zero()
                 if mask.numel() > 1:  # no broadcasting for masks
                     jacobian, err = masked_jacobians(self, mask)
                     done = True
--- a/theseus/core/cost_function.py
+++ b/theseus/core/cost_function.py
@@ -19,7 +19,36 @@ from theseus.geometry.lie_group_check import no_lie_group_check
 
 from .cost_weight import CostWeight, ScaleCostWeight
 from .theseus_function import TheseusFunction
-from .variable import Variable
+from .variable import Variable, masked_variables
+
+
+# This function returns the jacobians and error of the cost function
+# evaluated as if all cost function variables (optim and aux) are
+# masked with the given mask (applied over batch dimension).
+#
+# The mask is applied temporarily, so that the variables will retain
+# their original tensors after the function returns. The jacobians and error
+# will have their normal unmasked shapes, and elements masked out will be
+# set to zero.
+def masked_jacobians(
+    cost_fn: "CostFunction", mask: torch.Tensor
+) -> Tuple[List[torch.Tensor], torch.Tensor]:
+    cost_fn_vars: List[Variable] = cast(
+        List[Variable], list(cost_fn.optim_vars)
+    ) + list(cost_fn.aux_vars)
+    batch_size = max(v.shape[0] for v in cost_fn_vars)
+    aux_tensor = cost_fn_vars[0].tensor
+    # use cost_fn_vars[0] to get dtype and device
+    jacobians = [
+        aux_tensor.new_zeros(batch_size, cost_fn.dim(), v.dof())
+        for v in cost_fn.optim_vars
+    ]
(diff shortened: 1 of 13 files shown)


theseus/core/cost_weight.py
```diff
@@ -67,6 +73,9 @@ def __init__(
         self.scale.tensor = self.scale.tensor.view(-1, 1)
         self.register_aux_vars(["scale"])
 
+    def is_zero(self) -> torch.Tensor:
+        return (self.scale.tensor.abs() < EPS).view(-1)
```
fantaosha, line 77: Do we allow negative `self.scale`?

theseus/core/cost_weight.py
```diff
@@ -107,6 +116,9 @@ def __init__(
             self.diagonal.tensor = self.diagonal.tensor.view(1, -1)
         self.register_aux_vars(["diagonal"])
 
+    def is_zero(self) -> torch.Tensor:
+        return self.diagonal.tensor.abs().max(dim=1)[0] < EPS
```
fantaosha, line 120: Same here.

theseus/core/cost_function.py
```diff
@@ -37,6 +66,13 @@ def __init__(
         super().__init__(name=name)
         self._weight = cost_weight
 
+        # For now this is only supported inside vectorized cost functions
+        # When set to true, `weighted_jacobians_error()` first queries the
+        # weight to check indices of zero elements, then uses
+        # `masked_jacobians()` to evaluate jacobians only over non-zero
+        # elements.
+        self.__supports_masking__ = False
```
fantaosha, line 74: If I understand correctly, we can use this feature to mask some of the optimization problems that have already converged?

theseus/core/cost_weight.py
fantaosha: I agree this might speedup optimization. However, I'm afraid we might also lose some information for small `scale <EPS` but `scale != 0`. Not sure if this will raise some problems for differentiable optimization when we try to learn cost weights.

fantaosha (commented):
LGTM. But I have some concerns about whether it is reasonable to ignore small non-zero cost weights.

theseus/core/cost_function.py
luisenp: Yeah, I think we can find a way to use this to address #395.

theseus/core/cost_weight.py
mhmukadam: I guess `_ZERO_EPS = 1.0e-15` is a reasonably small value to catch this? Is there a case where something in this range would indeed need to pass through as non zero?

theseus/core/cost_weight.py
```diff
@@ -13,6 +13,9 @@
 from .variable import Variable, as_variable
 
 
+_ZERO_EPS = 1.0e-15
```
mhmukadam, line 16: Could move to constants.py if this is outside of the class?

theseus/embodied/motionmodel/double_integrator.py
```diff
@@ -118,6 +118,9 @@ def __init__(
         )
         self.register_aux_vars(["Qc_inv", "dt"])
 
+    def is_zero(self) -> torch.Tensor:
+        raise NotImplementedError
```
mhmukadam, line 122: Should this return a more reasonable value? We know GPCostWeight is non zero. Or we could make the `is_zero` function not abstract so every derived class doesn't need to define it?

theseus/core/variable.py
```diff
@@ -121,3 +121,27 @@ def as_variable(
...
     return Variable(tensor, name=name)
+
+
+# This context can be used to temporarily use a masked version of a list of
+# variables. For example,
+#
+# with masked_variables([v1, v2], mask):
+#   some_computation(v1, v2)
+#
+# Performs `some_computation()` using v1.tensor[mask] and v2.tensor[mask].
```
mhmukadam, line 132: Maybe add that typically (or always?) in practice this would mean masking some of the indices in a batch.

mhmukadam, line 132: In practice, we would use this to mask any dimension (not necessarily just batch indices), correct?

mhmukadam, line 132: Maybe add that typically (or always?) in practice this would mean masking some of the indices in a batch.

mhmukadam (approved):
LGTM!

theseus/core/cost_weight.py
luisenp: On the other hand, I think we can just use `tensor == 0`. Seems that torch's cutoff is about `7e-46` for floats and `1e-323` for doubles.

theseus/core/variable.py
luisenp: I think that's implied by `v1.tensor[mask]`, because the first dimension is the batch dimension.

theseus/core/cost_weight.py
luisenp: Changed accordingly.

theseus/core/cost_weight.py
luisenp: This is not used anymore.

theseus/embodied/motionmodel/double_integrator.py
luisenp: Changed as discussed offline. I'd prefer to make abstract, because this is always used under the hood, and there is no obvious default.

mhmukadam (approved):

fantaosha (approved):
LGTM.

PR #399, opened 2022-12-08
Add a timer util that adapts to a torch.device.

luisenp:
Removes boilerplate. Can be used as

```python
from thesus.utils import Timer

with Timer("cuda:0") as timer:
   some_stuff
print(timer.elapsed_time)
```

### DIFF
--- a/theseus/utils/__init__.py
+++ b/theseus/utils/__init__.py
@@ -10,4 +10,4 @@ from .sparse_matrix_utils import (
     split_into_param_sizes,
     tmat_vec,
 )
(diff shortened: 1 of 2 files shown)


fantaosha (approved):
LGTM!

mhmukadam:
Nice, might want to add the usage as comments above the class code.

luisenp:
> ```python
> from thesus.utils import Timer
> 
> with Timer("cuda:0") as timer:
>    some_stuff
> print(timer.elapsed_time)
> ```

Added in 556d6ba6da52540e7b1b76595cdeec96485acd41

PR #400, opened 2022-12-14
Fix bug in Vectorization of autodiff costs.

luisenp:
Previous version would group autodiff costs with different error functions, if the optim vars and aux vars were all of the same type.


mhmukadam (approved):
Good catch on this, funnily we did not see any uses cases run into this so far. Looks good!

fantaosha (approved):
LGTM. Good catch!

Issue #402, opened 2022-12-29
Refactor tests for differentiable Lie groups.


fantaosha:
## 🚀 Feature
See #378 and #380.

PR #404, opened 2023-01-01
Add so3.log()

fantaosha:
## Motivation and Context
Add `so3.log` and `so3.jlog`.

## Types of changes

## Checklist

### DIFF
--- a/tests/geometry/functional/common.py
+++ b/tests/geometry/functional/common.py
@@ -17,8 +17,17 @@ def check_lie_group_function(module, op_name: str, atol: float, args, funcs=None
         for jac_impl, jac in zip(jacs_impl, jacs):
             assert torch.allclose(jac_impl, jac, atol=atol)
     else:
-        for jac_impl, jac, func in zip(jac_impl, jacs, funcs):
+        for jac_impl, jac, func in zip(jacs_impl, jacs, funcs):
             if func is None:
                 assert torch.allclose(jac_impl, jac, atol=atol)
             else:
                 assert torch.allclose(func(jac_impl), func(jac), atol=atol)
+
+
+def left_project_func(module, group):
+    sels = range(group.shape[0])
+
+    def func(matrix: torch.Tensor):
+        return module.left_project(group, matrix[sels, ..., sels, :, :])
+
+    return func
--- a/tests/geometry/functional/common.py
+++ b/tests/geometry/functional/common.py
@@ -6,12 +6,19 @@
 import torch
 
 
-def check_lie_group_function(module, op_name: str, atol: float, *args):
+def check_lie_group_function(module, op_name: str, atol: float, args, funcs=None):
     op_impl = getattr(module, "_" + op_name + "_impl")
     op = getattr(module, op_name)
 
     jacs_impl = torch.autograd.functional.jacobian(op_impl, args)
     jacs = torch.autograd.functional.jacobian(op, args)
 
-    for jac_impl, jac in zip(jacs_impl, jacs):
-        assert torch.allclose(jac_impl, jac, atol=atol)
+    if funcs is None:
+        for jac_impl, jac in zip(jacs_impl, jacs):
+            assert torch.allclose(jac_impl, jac, atol=atol)
+    else:
+        for jac_impl, jac, func in zip(jac_impl, jacs, funcs):
+            if func is None:
+                assert torch.allclose(jac_impl, jac, atol=atol)
+            else:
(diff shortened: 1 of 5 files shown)


tests/geometry/functional/test_so3.py
```diff
@@ -19,7 +19,20 @@ def test_exp(batch_size: int, dtype: torch.dtype):
     tangent_vector = torch.rand(batch_size, 3, dtype=dtype, generator=rng)
 
     # check analytic backward for the operator
-    check_lie_group_function(so3, "exp", TEST_EPS, tangent_vector)
+    check_lie_group_function(so3, "exp", 1e-6, (tangent_vector,))
```
mhmukadam, line 22: Hardcoded `1e-6`.

tests/geometry/functional/test_so3.py
```diff
@@ -30,7 +43,7 @@ def test_adjoint(batch_size: int, dtype: torch.dtype):
     group = so3.rand(batch_size, generator=rng, dtype=dtype)
 
     # check analytic backward for the operator
-    check_lie_group_function(so3, "adjoint", TEST_EPS, group)
+    check_lie_group_function(so3, "adjoint", TEST_EPS, (group,))
```
mhmukadam, line 46: Minor point, but is there a workaround so you don't need to add `(x,)` for a single `x`?

mhmukadam (approved):
LGTM!

tests/geometry/functional/common.py
```diff
@@ -6,12 +6,28 @@
 import torch
 
 
-def check_lie_group_function(module, op_name: str, atol: float, *args):
+def check_lie_group_function(module, op_name: str, atol: float, args, funcs=None):
```
luisenp, line 9: What is `args` now? It seems that these are the input tensors? If so, better rename so it's easier to figure out what they are. Also missing type information for this and a few other inputs to this function. 

Can you also add some description of why we need to pass `funcs`?

tests/geometry/functional/test_so3.py
luisenp: I think this is fine, since this is only used for tests.

tests/geometry/functional/test_so3.py
luisenp: Probably fine? I think we have a bunch of tolerances hardcoded in many unit tests.

tests/geometry/functional/test_so3.py
```diff
@@ -41,7 +54,7 @@ def test_inverse(batch_size: int, dtype: torch.dtype):
     group = so3.rand(batch_size, generator=rng, dtype=dtype)
 
     # check analytic backward for the operator
-    check_lie_group_function(so3, "inverse", TEST_EPS, group)
+    check_lie_group_function(so3, "inverse", TEST_EPS, (group,))
```
luisenp, line 57: You can make all this code much more concise and readable, and all in a single unit tests, by doing something similar to:


```python
FN_CFGS = [
  ("inverse", ("group0",)),
  ("hat", ("tangent",)),
  ("compose", ("group0", "group1")),
]
@pytest.mark.parametrize(fn_cfg, FN_CFGS)
def test_so3_op(fn_cfg):
  # create the generator, manual seed, etc. (also include batch size, etc.)
  input_vals = {
    "group0": so3.rand(batch_size, generator=rng, dtype=dtype), 
    "tangent": torch.rand(batch_size, 3, dtype=dtype, generator=rng), 
    "group1": torch.rand(batch_size, 3, dtype=dtype, generator=rng),
    "matrix": torch.rand(
        batch_size,
        int(torch.randint(1, 20, (1,), generator=rng)),
        3,
        dtype=dtype,
        generator=rng,
    ),
  }  # and anything that's useful like near_zero stuff or w/e

  fn_inputs = tuple(input_vals[k] for k in fn_cfg[1])   # convert cfg to tensors
  check_lie_group_function(so3, fn_cfg[0], TEST_EPS, fn_inputs)  
```

luisenp (changes requested):
Logic LGTM, but I proposed a way to clean up the tests in this file.

luisenp:
On the other hand, for the purposes of this PR you can ignore the test refactor, since it's also related to #402. I was planning to work on that issue myself, but I think it would be good if you can take a shot at it, after we merge all the open PRs. The comment above should be a good place to start (applies to SE3 as well).

tests/geometry/functional/common.py
mhmukadam: +1

PR #415, opened 2023-01-02
Add forward kinematics package to `labs`

fantaosha:
## Types of changes

## Checklist


theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(name, parent, child, origin, dtype)
+
+    def dof(self) -> int:
+        return 1
+
+    @abc.abstractmethod
+    def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
```
luisenp, line 102: `_rotation_impl` implies that there is another `rotation()` method that's public and that works something like the following:

```python
def rotation(self, angle: torch.Tensor) -> torch.Tensor:
    # common initialization such as checking angle shape
    pre_out = self._rotation_impl(angle)
    # post process pre_out
    return processed_out
```

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        parent: int = -1,
+        child: int = -1,
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(name, parent, child, origin, dtype)
+        self._axis[4] = 1
+
+    def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
+        if angle.ndim == 2 and angle.shape[1] == 1:
```
luisenp, line 183: If you implement `rotation()` in the super class as mentioned above, all this angle checking code can go there and you don't need to reimplement this check in all subclasses.

luisenp (changes requested):
Looks good! Requesting some small changes that should be easy to address.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import abc
+import warnings
+from typing import Optional
+import torch
+
+from theseus.geometry.functional import so3, se3
```
mhmukadam, line 11: Based on offline discussion we'll move the files to `/theseus/labs/kinematics/...`

Also the import needs updating.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+class Joint(abc.ABC):
```
mhmukadam, line 14: Should we have any unit tests accompanying these or will that come in later PRs that add other data structures.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+    def __init__(
+        self,
+        name: str,
+        parent: int = -1,
+        child: int = -1,
```
mhmukadam, line 19: These could be `Joint` types instead of `int`. Are there any tradeoffs worth considering?

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
```
mhmukadam, line 21: Also add `device`?

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+                warnings.warn(
+                    f"tensor.dtype {origin.dtype} does not match given dtype {dtype}, "
+                    "tensor.dtype will take precendence."
+                )
+            dtype = origin.dtype
+        else:
+            origin = torch.zeros(1, 3, 4, dtype=dtype)
+            origin[:, 0, 0] = 1
+            origin[:, 1, 1] = 1
+            origin[:, 2, 2] = 1
```
mhmukadam, line 39: Could we add a identity SE3 helper function and then call that here?

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def set_child(self, child: int):
+        self._child = child
+
+    @abc.abstractmethod
+    def dof(self) -> int:
+        pass
+
+    @abc.abstractmethod
+    def relative_pose(self, angle: torch.Tensor) -> torch.Tensor:
```
mhmukadam, line 83: Add a comment about this, for instance relative to what?

theseus/embodied/robot/joint.py
luisenp: I think these are made `Link` types in a later PR.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        child: int = -1,
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(name, parent, child, origin, dtype)
+
+    def dof(self) -> int:
+        return 1
+
+    @abc.abstractmethod
```
mhmukadam, line 101: If you have abstract methods then you want to derive from abc as well
`class _RevoluteJointImpl(Joint, abc.ABC):`

theseus/embodied/robot/joint.py
mhmukadam: If just a private/internal implementation is needed then `def _rotation(...)` is sufficient.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        rot = angle.new_zeros(angle.shape[0], 3, 3)
+        rot[:, 1, 1] = 1
+        rot[:, 0, 0] = angle.cos()
+        rot[:, 0, 2] = angle.sin()
+        rot[:, 2, 0] = -rot[:, 0, 2]
+        rot[:, 2, 2] = rot[:, 0, 0]
+        return rot
+
+
+class RevoluteJointZ(_RevoluteJointImpl):
```
mhmukadam, line 196: Wondering if the following alternative is better/worse in any way: only one `RevoluteJoint` class that has a full, x, y, or z type as a property (this would get rid of the Impl intermediate as well).

mhmukadam (commented):
Looks good so far. A few comments for us to hash out.

theseus/embodied/robot/joint.py
luisenp: Yeah, this crossed my mind as well. I think I'm OK with either.

mhmukadam: I am okay with either as well. We should consider tradeoffs like ease of use and potential overheads.

theseus/embodied/robot/joint.py
mhmukadam: Looks like this is in #422.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        return self._dtype
+
+    @property
+    def axis(self) -> torch.Tensor:
+        return self._axis
+
+    def set_parent(self, parent: int):
+        self._parent = parent
+
+    def set_child(self, child: int):
```
luisenp, line 75: Actually, I think a lot of these properties (maybe all) can be removed and replaced with public attributes. Using a property is more appropriate if we want to hide some extra computation under the hood (e.g., checking for correct inputs).

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+    def origin(self) -> torch.Tensor:
+        return self._origin
+
+    @property
+    def dtype(self) -> torch.dtype:
+        return self._dtype
+
+    @property
+    def axis(self) -> torch.Tensor:
+        return self._axis
```
exhaustin, line 70: Only revolute joints have axes, so I don't think a general `Joint` abstract class should have this as an attribute.

theseus/embodied/robot/joint.py
exhaustin: I am guessing that as joints are basically pose transforms between links, `relative_pose` refers to the pose of the child link relative to the parent link.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+    def set_child(self, child: int):
+        self._child = child
+
+    @abc.abstractmethod
+    def dof(self) -> int:
+        pass
+
+    @abc.abstractmethod
+    def relative_pose(self, angle: torch.Tensor) -> torch.Tensor:
+        pass
```
exhaustin, line 84: - Why output an se3 vector instead of an SE3 object?
- Should each joint also have its own Jacobian method?
- Different types of joints have different parametrizations. A revolute joint requires a single `angle` value, but that is not the case for other types of joints. Since this is a general abstract `Joint` class, maybe this function should have a more general input, such as a parameter vector of length `self.dof`?

theseus/embodied/robot/joint.py
exhaustin: Maybe a function name like `compute_joint_transform` or `parent_to_child_transform` would be more descriptive?

theseus/embodied/robot/joint.py
luisenp: For the first question, it might be because this is based on the initial version of the new Lie groups library, which was all regular tensor to regular tensors. This will probably change with the upcoming class-based API (see #450 and #452, for a quick peek). 

Agree on the third point, I think this would be preferable. 

@fantaosha any thoughts on the second question? (or anything I missed above)

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+    ):
+        if origin is None and dtype is None:
+            dtype = torch.get_default_dtype()
+        if origin is not None:
+            if origin.shape[0] != 1 or not se3.check_group_tensor(origin):
+                raise ValueError("Origin must be an element of SE(3).")
+
+            if dtype is not None and origin.dtype != dtype:
+                warnings.warn(
+                    f"tensor.dtype {origin.dtype} does not match given dtype {dtype}, "
```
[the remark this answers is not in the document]
fantaosha: Refactored.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        parent: int = -1,
+        child: int = -1,
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        if angle_axis.ndim == 1:
+            angle_axis = angle_axis.view(-1, 1)
+
+        if angle_axis.ndim != 2 or angle_axis.shape != (3, 1):
+            raise ValueError("The angle axis must be a 3-D vector.")
```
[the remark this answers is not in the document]
fantaosha: `numel()` is used now and the error message is `"The revolute axis must have 3 elements."`.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(name, parent, child, origin, dtype)
+        self._axis[3] = 1
+
+    def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
+        if angle.ndim == 2 and angle.shape[1] == 1:
+            angle = angle.view(-1)
+        if angle.ndim != 1:
```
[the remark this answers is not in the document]
fantaosha: Fixed.

theseus/embodied/robot/joint.py
```diff
@@ -0,0 +1,219 @@
...
+
+    @property
+    def dtype(self) -> torch.dtype:
+        return self._dtype
+
+    @property
+    def axis(self) -> torch.Tensor:
+        return self._axis
+
+    def set_parent(self, parent: int):
```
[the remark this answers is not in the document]
fantaosha: All the setters have been removed for encapsulation.

theseus/embodied/robot/joint.py
fantaosha: `rotation()` and `translation()` are defined.

theseus/embodied/robot/joint.py
fantaosha: Refactored.

theseus/embodied/robot/joint.py
fantaosha: Will go back to this once `identity()` is added to the Lie group library.

theseus/embodied/robot/joint.py
fantaosha: Now there are two classes for `RevoluteJoint`, i.e., `RevoluteJointFull` and `RevoluteJointXYZ`

theseus/embodied/robot/joint.py
fantaosha: All the setters have been removed.

theseus/embodied/robot/joint.py
fantaosha: For any types of joints, the input `tensor` shape should be `batch_size x dof`.

PR #416, opened 2023-01-02
Add prismatic joint

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/theseus/embodied/robot/joint.py
+++ b/theseus/embodied/robot/joint.py
@@ -23,6 +23,7 @@ class Link(abc.ABC):
     ):
         self._name = name
         self._id = id
+        # self._parent_joint is None means that the link is rigidly connected to the world
         self._parent_joint = parent_joint
         self._child_joints = child_joints if child_joints else []
         self._ancestor_links: List[Link] = []
@@ -226,11 +227,8 @@ class RevoluteJoint(_RevoluteJointImpl):
         dtype: Optional[torch.dtype] = None,
         device: DeviceType = None,
     ):
-        if revolute_axis.ndim == 1:
-            revolute_axis = revolute_axis.view(-1, 1)
-
-        if revolute_axis.ndim != 2 or revolute_axis.shape != (3, 1):
-            raise ValueError("The revolute axis must be a 3-D vector.")
+        if revolute_axis.numel() != 3:
+            raise ValueError("The revolute axis must have 3 elements.")
 
         super().__init__(name, id, parent_link, child_link, origin, dtype, device)
 
@@ -240,7 +238,7 @@ class RevoluteJoint(_RevoluteJointImpl):
         if revolute_axis.device != self.device:
             raise ValueError(f"The device of revolute_axis should be {self.device}.")
 
-        self._axis[3:] = revolute_axis
+        self._axis[3:] = revolute_axis.view(-1, 1)
 
     def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
         if angle.ndim == 1:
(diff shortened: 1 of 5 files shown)


theseus/embodied/robot/joint.py
```diff
@@ -217,3 +217,158 @@ def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
...
+        name: str,
+        parent: int = -1,
+        child: int = -1,
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(name, parent, child, origin, dtype)
+        self._axis[3] = 1
+
+    def _translation_impl(self, angle: torch.Tensor) -> torch.Tensor:
```
luisenp, line 293: The code in `_translation_impl` and `relative_pose` of these 3 classes is exactly the same except for one number, so you can just have a base class `_PrismaticJointAxis` that defines a member variable `self._axis_idx: Optional[int] = None`, which gets overridden by all base classes. Then we have 
```python
def _translation_impl(self, angle: torch.Tensor) -> torch.Tensor:
    self._check_angle_shape(angle)
    trans = angle.new_zeros(angle.shape[0], 3)
    trans[:, self._axis_idx] = angle
    return trans

def relative_pose(self, angle: torch.Tensor) -> torch.Tensor:
    self._check_angle_shape(angle)
    ret = angle.new_empty(angle.shape[0], 3, 4)
    ret[:, :, :3] = self.origin[:, :, :3]
    ret[:, :, 3] = angle * self.origin[:, :, self._axis_idx] + self.origin[:, :, 3]
    return ret
```

theseus/embodied/robot/joint.py
```diff
@@ -217,3 +217,158 @@ def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
...
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(name, parent, child, origin, dtype)
+
+    def dof(self) -> int:
+        return 1
+
+    @abc.abstractmethod
+    def _translation_impl(self, angle: torch.Tensor) -> torch.Tensor:
```
luisenp, line 237: Same comment as in `RevoluteJoint` regarding a `translation()` method, with similar structure.

luisenp, line 237: Maybe `translation()` and `rotation()` (in the other class) should be called `translate()` and `rotate()`?

luisenp (changes requested):
Left some comments to address.

theseus/embodied/robot/joint.py
```diff
@@ -113,25 +113,25 @@ def relative_pose(self, angle: torch.Tensor) -> torch.Tensor:
 class RevoluteJoint(_RevoluteJointImpl):
     def __init__(
         self,
-        angle_axis: torch.Tensor,
         name: str,
+        revolute_axis: torch.Tensor,
```
mhmukadam, line 117: Maybe just `axis`, I am guessing revolute is implied from the class name.

theseus/embodied/robot/joint.py
```diff
@@ -217,3 +217,158 @@ def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
...
+        parent: int = -1,
+        child: int = -1,
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        if prismatic_axis.ndim == 1:
+            prismatic_axis = prismatic_axis.view(-1, 1)
+
+        if prismatic_axis.ndim != 2 or prismatic_axis.shape != (3, 1):
+            raise ValueError("The prismatic axis must be a 3-D vector.")
```
exhaustin, line 264: I imagine this logic of "checking if the input is a valid vector of correct length, and unsqueezing it into a vertical vector if it's 1-D" would be very commonly used, not just within the robot model module. Should this be a piece of logic that can be standardized and deduplicated within an `utils` of some sort?

luisenp: If we anticipate this to be used outside robot model, it would be useful to make an utility function. Otherwise, we should at least delegate this repeated logic to the super class.

theseus/embodied/robot/joint.py
fantaosha: I think noun should be better than verb since the outputs are `3x3` or `3x1` tensors

theseus/embodied/robot/joint.py
```diff
@@ -217,3 +217,158 @@ def _rotation_impl(self, angle: torch.Tensor) -> torch.Tensor:
...
+        prismatic_axis: torch.Tensor,
+        parent: int = -1,
+        child: int = -1,
+        origin: Optional[torch.Tensor] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        if prismatic_axis.ndim == 1:
+            prismatic_axis = prismatic_axis.view(-1, 1)
+
+        if prismatic_axis.ndim != 2 or prismatic_axis.shape != (3, 1):
```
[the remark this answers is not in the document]
fantaosha: Refactored. And now the error message is "The prismatic axis must have 3 elements."

theseus/embodied/robot/joint.py
fantaosha: Refactored. Thanks!

theseus/embodied/robot/joint.py
fantaosha: Refactored.

PR #417, opened 2023-01-02
Add robot model

fantaosha:
## Types of changes

## Checklist


theseus/embodied/robot/joint.py
```diff
@@ -69,31 +75,61 @@ def dtype(self) -> torch.dtype:
     def axis(self) -> torch.Tensor:
         return self._axis
 
-    def set_parent(self, parent: int):
+    def set_id(self, id: int):
+        self._id = id
+
+    def set_parent(self, parent: Optional[Link]):
```
mhmukadam, line 81: Is `Optional` intended here, if so should it be followed by a `= None`?

theseus/embodied/robot/link.py
```diff
@@ -0,0 +1,75 @@
...
+import torch
+
+
+class Link(abc.ABC):
+    def __init__(
+        self,
+        name: str,
+        id: int = -1,
+        parent: Any = None,
+        children: Optional[List[Any]] = None,
```
mhmukadam, line 17: Is it `Any` or `Union[Link, Joint]`?

theseus/embodied/robot/link.py
```diff
@@ -0,0 +1,75 @@
...
+        dtype: Optional[torch.dtype] = None,
+    ):
+        self._name = name
+        self._id = id
+        self._parent = parent
+        self._children = children if children else []
+        self._ancestors: List[Link] = []
```
mhmukadam, line 24: SInce this is not passed in `__init__` I am guessing this is compute internally. If so should `self.update_ancestors()` be called here?

theseus/embodied/robot/link.py
```diff
@@ -0,0 +1,75 @@
...
+        while joint is not None:
+            link = joint.parent
+            self._ancestors.insert(0, link)
+            joint = link.parent
+
+    def add_child(self, child: Any):
+        self._children.append(child)
+
+    def remove_child(self, child: Any):
+        self._children.remove(child)
```
mhmukadam, line 75: Would these mess with the `_ancestors` list? Would need to trigger `update_ancestors` but across the entire tree not just for a single class instance which would be more of a global update call that I guess recursively calls all the individual `update_ancestors`.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+from .joint import Joint, FixedJoint, RevoluteJoint, PrismaticJoint
+from .link import Link
+
+
+class Robot(abc.ABC):
+    def __init__(self, name: str, dtype: torch.dtype = None):
+        if dtype is None:
+            dtype = torch.get_default_dtype()
+        self._name: str = name
+        self._dtype: torch.dtype = dtype
```
mhmukadam, line 21: Maybe can consolidate the if block and assignment line for _dtype.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+        self._dof: int = 0
+        self._num_joints: int = 0
+        self._num_links: int = 0
+        self._joints: List[Joint] = []
+        self._links: List[Link] = []
+        self._joint_map: Dict[str, Joint] = {}
+        self._link_map: Dict[str, Link] = {}
+
+    @classmethod
+    def from_urdf_file(cls, urdf_file: str, dtype: torch.dtype = None) -> "Robot":
```
mhmukadam, line 31: Trailing `.._file` could be dropped from the name.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+            num_joints = num_joints + 1
+            joint.child.set_id(num_joints)
+            robot._links.append(joint.child)
+
+        robot._num_links = len(robot.links)
+        robot._num_joints = len(robot.joints)
+
+        for link in robot.links:
+            if link.parent is not None:
+                link.set_ancestors(link.parent.parent.ancestors + [link.parent.parent])
```
mhmukadam, line 137: `update_ancestors` doesn't seem to be called anywhere.

theseus/embodied/robot/FK.py
```diff
@@ -0,0 +1,58 @@
...
+
+from theseus.geometry.functional import se3
+from .robot import Robot
+from .link import Link
+from .joint import Joint
+
+# TODO: Add support for joints with DOF>1
+
+
+def ForwardKinematicFactory(robot: Robot, link_names: Optional[List[str]] = None):
```
mhmukadam, line 18: The full diff FK test in #420 is the overall test to verify everything put together. Are there any unit tests we can add for the smaller pieces?

theseus/embodied/robot/joint.py
```diff
@@ -8,23 +8,24 @@
...
 from theseus.labs.lie_functional import so3, se3
 
 
 class Joint(abc.ABC):
     def __init__(
         self,
         name: str,
-        parent: int = -1,
-        child: int = -1,
+        id: int = -1,
```
luisenp, line 19: I think making `id` optional with default to `None` is cleaner that using `-1`. Rather than checking `id >= 0`, we can check `id is not None`. It is also easier to understand when others read the code.

mhmukadam (commented):
Reviewed so far, left some comments.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+
+        joints_to_visit: List[Joint] = []
+        root = robot.link_map[urdf_model.get_root()]
+        num_joints = 0
+        root.set_id(0)
+        robot._links.append(root)
+        robot._dof = 0
+        joints_to_visit = joints_to_visit + root.children
+
+        while joints_to_visit:
```
luisenp, line 109: Add some comments here explaining what this does. Here is what I understood from our conversation:

* This part creates IDs for all joints and links, and appends the nodes to the `_joints` and `_links` lists.
* The robot's dof only depends on non-fixed joints.
* Fixed joints are processed after non-fixed ones to make their IDs larger values.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+                robot._dof += joint.dof()
+                robot._joints.append(joint)
+                num_joints = num_joints + 1
+                joint.child.set_id(num_joints)
+                robot._links.append(joint.child)
+
+                joints_to_visit = joints_to_visit + joint.child.children
+
+        for _, joint in robot.joint_map.items():
+            if joint.id >= 0:
```
luisenp, line 122: Following a comment above, better to do `if joint.id is not None:`

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+                continue
+            if not isinstance(joint, FixedJoint):
+                raise ValueError(f"{joint.name} is expected to a fixed joint.")
+            joint.set_id(num_joints)
```
luisenp, line 126: This code is the same used above for non-fixed joints. You can replace this by some shared function, and with an extra line for the above loop to increase the robot's dof.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+            robot._joints.append(joint)
+            num_joints = num_joints + 1
+            joint.child.set_id(num_joints)
+            robot._links.append(joint.child)
+
+        robot._num_links = len(robot.links)
+        robot._num_joints = len(robot.joints)
+
+        for link in robot.links:
```
luisenp, line 135: Add a comment that says something like "caching a link's ancestor for ease of FK computation".

luisenp (changes requested):
Looking good! Left a few comments.

theseus/embodied/robot/robot.py
luisenp: Maybe `def from_urdf(fname: str)`?

theseus/embodied/robot/link.py
luisenp: Is this a typical use case? If this just results in some annoying bookkeeping for what turns out to be an uncommon scenario, I'd rather just "lock" links once the ancestors are set for the first time, and prevent users from adding or removing any children.

fantaosha: These functions are only used to construct the kinematic tree from urdf. In fact, `_ancestors` are updated at the end so these functions should not induce any bugs for current implementation. But I'm considering making these functions private so that users have no access to them.

theseus/embodied/robot/FK.py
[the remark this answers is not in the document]
fantaosha: The `link_names` are links that we want to get their poses through FK. In some cases, we don't want the poses for all the links. This will simplify the computation.

theseus/embodied/robot/FK.py
```diff
@@ -0,0 +1,58 @@
...
+        links = robot.links
+    else:
+        links = [robot.link_map[name] for name in link_names]
+
+    ancestors = []
+
+    for link in links:
+        ancestors += [anc.id for anc in link.ancestors]
+
+    link_ids = sorted(list(set(ancestors)))
```
exhaustin, line 31: This seems rather inefficient, as the same chain of ancestors would be traversed multiple times if FK for multiple links on the same chain are desired to be computed.
Can we either
1. Do a BFS/DFS from the root, and only keep links that are in the path to links that are in `links`?
2. From each link in `links`, trace back to the root and stop when reaching a link that is previously traced.

Both of these methods can directly output a sorted list of relevant links without repeated traversals.
By doing so we can also completely eliminate the need to cache ancestors.

Also, if for some reason the link ids are not topologically sorted (some child having a smaller id than one of its parents), the methods mentioned above would also cause no problems for computing the FK, in contrary to the current implementation that relies on correct order of link ids.

theseus/embodied/robot/FK.py
```diff
@@ -0,0 +1,58 @@
...
+        poses[0][:, 2, 2] = 1
+
+        for id in link_ids[1:]:
+            curr: Link = robot.links[id]
+            joint: Joint = robot.links[id].parent
+            prev: Link = joint.parent
+            relative_pose = (
+                joint.relative_pose(angles[:, joint.id])
+                if joint.id < robot.dof
+                else joint.relative_pose()
```
exhaustin, line 52: I'm worried that if joint ID assignment is not closely adhering to some standard this might break. Is there a more systematic way to map the vector joint states to individual joints? 
One possibility I'm thinking is to decouple the topological relationship of links and joints with their IDs. The IDs are used to order state and action vectors, while FK and other kinematic / dynamic computations directly operate on the tree data structure.

Thus, there also won't be the need to "process fixed joints after non-fixed ones to make their IDs larger value"

theseus/embodied/robot/link.py
exhaustin: Might just be my subjective opinion, but I don't think it's safe to have the validity of a state variable depend on correct usage of other parts of the code. Even having a boolean flag like `_is_ancestor_up_to_date`, setting it to `True` in `update_ancestors()`, setting it to `False` in `add_child()`, `remove_child()`, ..etc, and printing out a warning in `ancestors()` if `_is_ancestor_up_to_date is False` would be a good safeguard against incorrect usage of the module.

theseus/embodied/robot/link.py
exhaustin, line 17: Each link is connected to exactly one joint that connects to one parent link (except the root), and 0~N joints that connect to children links.
I find the Pybullet convention to be rather clean: Associate each link with the joint that connects to its parent.
So I suggest that we can have `Link._parent` and `Link._children` contain adjacent Links, which makes traversal for kinematics/dynamics easier instead of having to go through link-joint-link every time. If we want to access the joint that connects to some link, it can be done by something like `Link._joint` or `Link._parent_joint`.

exhaustin (commented):
It would be great if we can find time to discuss the data structure used to contain kinematic trees. I feel like I would very likely use this a lot if this gets shipped, so I definitely want it to be as intuitive and easy to use as possible!

theseus/embodied/robot/FK.py
fantaosha: - BFS/DFS is definitely useful and I will refactor the code later with this.
- We already guaranteed that the links are topologically sorted when creating the robot model from URDF, thus, this should not be an issue for our code.

theseus/embodied/robot/FK.py
fantaosha: Yeah, I do notice that some libraries decouple link and joint ids. The most immediate benefit to associate link and joint ids improves the computational efficiency. For example, we could get the parent joint id of a link. As far as I know, lots of fast kinematics and dynamics libraries construct the kinematic tree in this way.

theseus/embodied/robot/link.py
fantaosha: That's a good point. Thanks!

theseus/embodied/robot/FK.py
exhaustin: I'm not sure I fully understand, do you mean that getting the some ancestor joint id is now constant time (joint.id - X) to traverse X steps, instead of needing to backtrack X times? But that only applies to kinematic chains and would break with trees?
Let's discuss in more detail during the meeting today.

theseus/embodied/robot/robot.py
fantaosha: I remember in ROS, there are methods called `from_urdf()` and `from_urdf_file()` while the input from the former is a `urdf_model`.
Screenshot 2023-02-28 at 7 03 31 PM

theseus/embodied/robot/robot.py
fantaosha: I have made this function `private` for encapsulation.

theseus/embodied/robot/joint.py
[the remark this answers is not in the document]
fantaosha: I removed all property setters since these properties are assumed to be fixed once the robot is created.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import abc
+from typing import List, Dict, Optional
+import urdf_parser_py.urdf as urdf
```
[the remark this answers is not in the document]
fantaosha: I don't think so. Austin's implementation also used this dependence.

theseus/embodied/robot/link.py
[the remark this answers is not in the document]
fantaosha: There seem some circular dependence issues since `Link` and `Joint` depend on each other

theseus/embodied/robot/joint.py
fantaosha: The parent can be `None` or `Link`, where `None` means the `parent_link` is corrected to a fixed world frame.

fantaosha: I will add a comment there.

theseus/embodied/robot/link.py
fantaosha: It should `Joint`. But due to circular dependence, it seems impossible to `import Joint` in `Link`.

theseus/embodied/robot/link.py
fantaosha: These `setters` are removed for encapsulation.

theseus/embodied/robot/robot.py
fantaosha: refactored.

theseus/embodied/robot/FK.py
fantaosha: Not sure yet. Maybe inverse kinematics.

theseus/embodied/robot/link.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @property
+    def ancestors(self) -> List["Link"]:
+        return self._ancestors
+
+    @property
+    def dtype(self) -> torch.dtype:
+        return self._dtype
+
+    def set_id(self, id: int):
```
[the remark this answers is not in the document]
fantaosha: I removed all property setters since these properties are assumed to be fixed once the robot is created.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+        self._dtype: torch.dtype = dtype
+        self._dof: int = 0
+        self._num_joints: int = 0
+        self._num_links: int = 0
+        self._joints: List[Joint] = []
+        self._links: List[Link] = []
+        self._joint_map: Dict[str, Joint] = {}
+        self._link_map: Dict[str, Link] = {}
+
+    @classmethod
```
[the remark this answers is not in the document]
fantaosha: Refactored this function as a static method.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+                joint = joint_type(
+                    urdf_joint.name, parent=parent, child=child, origin=origin
+                )
+            child.set_parent(joint)
+            parent.add_child(joint)
+            robot._joint_map[urdf_joint.name] = joint
+
+        for _, link in robot.link_map.items():
+            for joint in link._children:
+                if isinstance(joint, FixedJoint):
```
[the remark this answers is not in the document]
fantaosha: That's correct. Comments added.

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+            return origin
+
+        urdf_model = urdf.URDF.from_xml_file(urdf_file)
+        robot = Robot(urdf_model.name, dtype)
+
+        for urdf_link in urdf_model.links:
+            link = Link(urdf_link.name)
+            robot._link_map[urdf_link.name] = link
+
+        for urdf_joint in urdf_model.joints:
```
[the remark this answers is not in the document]
fantaosha: Refactored

theseus/embodied/robot/robot.py
```diff
@@ -0,0 +1,175 @@
...
+                )
+            child.set_parent(joint)
+            parent.add_child(joint)
+            robot._joint_map[urdf_joint.name] = joint
+
+        for _, link in robot.link_map.items():
+            for joint in link._children:
+                if isinstance(joint, FixedJoint):
+                    subjoints: List[Joint] = joint.child.children
+                    joint.child.set_children([])
```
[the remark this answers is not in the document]
fantaosha: Added.

theseus/embodied/robot/joint.py
fantaosha: Good idea.

theseus/embodied/robot/robot.py
fantaosha: Refactored

theseus/embodied/robot/robot.py
fantaosha: Added

theseus/embodied/robot/robot.py
fantaosha: Refactored.

theseus/embodied/robot/robot.py
fantaosha: Added.

theseus/embodied/robot/link.py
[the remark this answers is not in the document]
fantaosha: Move `Link` to `joint.py` to avoid circular dependence.

fantaosha: Moved `Link` to `joint.py` and now `Any->Joint`

PR #421, opened 2023-01-04
Fix black and wheel script errors

luisenp:


build_scripts/build_wheel.sh
```diff
@@ -69,7 +69,7 @@ else
     BASPACHO_CUDA_ARGS="-DCMAKE_CUDA_COMPILER=/usr/local/cuda-${CUDA_VERSION}/bin/nvcc -DBASPACHO_CUDA_ARCHS='${BASPACHO_CUDA_ARCHS}'"
 fi
 
-for PYTHON_VERSION in 3.9; do
+for PYTHON_VERSION in 3.10; do
```
mhmukadam, line 72: Are we making 3.10 as the default for pypi install?

mhmukadam (commented):
LGTM!

build_scripts/build_wheel.sh
luisenp: Yes. Looks like we are not able to compile for CPython 3.9 anymore, as far as I can tell because of sparsesuite (same reason I was only able to do 3.9 before).

mhmukadam: Sounds good. We should update the readme (and tags on top of readme) accordingly.

PR #424, opened 2023-01-11
Add a `labs` package for experimental stuff

luisenp:
* Compiled wheels don't include this package by default (which will be the case for pypi).
* It's available via `pip install -e .` or by compiling wheels with `THESEUS_INCLUDE_LABS=1`.
* Lie group functional library moved to labs.


build_scripts/build_wheel.sh
```diff
@@ -69,7 +76,7 @@ else
     BASPACHO_CUDA_ARGS="-DCMAKE_CUDA_COMPILER=/usr/local/cuda-${CUDA_VERSION}/bin/nvcc -DBASPACHO_CUDA_ARCHS='${BASPACHO_CUDA_ARCHS}'"
 fi
 
-for PYTHON_VERSION in 3.10; do
+for PYTHON_VERSION in 3.10  ; do
```
mhmukadam, line 79: Was the added space after 3.10 intentional?

tests/labs/lie_functional/common.py
```diff
@@ -6,6 +6,9 @@
 import torch
 
 
+TEST_EPS = 5e-7
```
mhmukadam, line 9: Where was this before getting added here? I didn't see another file in this PR using it.

mhmukadam (approved):
LGTM! Also nice, you consolidated the decorators.
1. Maybe update the readme with the labs specific instruction you have in the PR message above?
2. `labs/__init__.py` is missing license header :D

tests/labs/lie_functional/common.py
luisenp: Before it was imported from here, meaning that labs code was peaking into Theseus main code. 

image

build_scripts/build_wheel.sh
luisenp: Not intentional.

tests/labs/lie_functional/common.py
mhmukadam: Ah!

mhmukadam (approved):

PR #431, opened 2023-01-17
Use IMPLICIT for test_theseus_layer and fix related bugs

luisenp:


theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -566,12 +567,19 @@ def _step(
...
         converged_indices: torch.Tensor,
         force_update: bool,
+        truncated_grad_loop: bool,
         **kwargs,
     ) -> Tuple[torch.Tensor, bool]:
         tensor_dict, err = self._compute_retracted_tensors_and_error(
             delta, converged_indices, force_update
         )
-        reject_indices = self._complete_step(delta, err, previous_err, **kwargs)
+        if truncated_grad_loop:
```
mhmukadam, line 576: Can't remember, is the implicit grad part of the truncated_grad_loop flag?

mhmukadam (approved):
Nice catch, looks good!

bamos (approved):

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Yes, `truncated_grad_loop` is the final attached loop of "TRUNCATED" and "IMPLICIT". I was thinking that this is kind of a confusing name tbh.

mhmukadam: Yeah we could rename to `implicit_and_trunc_grad_block` or maybe `..._loop` is okay. We can also add a comment to clarify.

PR #436, opened 2023-01-18
Remove nox dependency

luisenp:
This standardizes local and CI test suite.


tests/geometry/test_point_types.py
```diff
@@ -63,6 +64,18 @@ def test_operations_mypy_cast():
         import mypy.api
     except ModuleNotFoundError:
         return
+    import sys
+
+    python_version = sys.version_info
+    if python_version[0] == 3 and (
+        python_version[1] > 10 or python_version[1] == 10 and python_version[2] > 6
```
mhmukadam, line 71: Nit
```
if python_version[0] >= 3 and python_version[1] >= 10 and python_version[2] >= 7
```

mhmukadam (approved):

tests/geometry/test_point_types.py
luisenp: I think this doesn't work because it won't catch something like 3.11.1 (fails the third check).

mhmukadam: Oh yeah, haha

.circleci/config.yml
```diff
@@ -220,55 +247,135 @@ run_tests_from_wheel: &run_tests_from_wheel
...
 jobs:
-  py37_linux:
+  # Pre-commit tests
+  py38_precommit:
+    docker:
+      - image: cimg/python:3.8
+    steps:
+      - checkout
+      - <<: *install_suitesparse
+      - <<: *install_functorch
```
mhmukadam, line 261: do we still need install_functorch since it is now in pytorch?

.circleci/config.yml
```diff
@@ -220,55 +247,135 @@ run_tests_from_wheel: &run_tests_from_wheel
...
+      - <<: *run_lie_group_tests
+
+  py310_lie_tests:
+    docker:
+      - image: cimg/python:3.10
+    steps:
+      - checkout
+      - <<: *install_suitesparse
+      - <<: *install_functorch
+      - <<: *build_baspacho_cpu
```
mhmukadam, line 344: You probably don't need baspacho to run lie group tests?

mhmukadam (approved):

.circleci/config.yml
luisenp: Not sure actually, let me check.

.circleci/config.yml
luisenp: Looks like we never made suitesparse optional, and pip install -e . fails.

.circleci/config.yml
luisenp: Yes, this works. Curiously, we were installing torch via the command to install functorch.

.circleci/config.yml
mhmukadam: Ah okay yeah it would need extra things for the selective tests. We probably don't save that much time anyways by avoiding the install.

mhmukadam (approved):

PR #440, opened 2023-01-19
New NonlinearOptimizer hierarchy

luisenp:
This PR moves the logic specific to nonlinear least squares problems from `NonlinearOptimizer` to the `NonlinearLeastSquares` subclass. The base class now only provides some default info tracking methods, and defines an abstract method called `_error_metric()`, which can be used instead of the previously hardcoded sum of squares.


bamos (approved):

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -4,18 +4,83 @@
...
+    NonlinearOptimizerStatus,
+)
 
-from .nonlinear_optimizer import NonlinearOptimizer
 
+class BackwardMode(Enum):
+    UNROLL = 0
+    IMPLICIT = 1
+    TRUNCATED = 2
+    DLM = 3
```
mhmukadam, line 32: Should BackwardModes go to the base Optimizer class level? Any derived optimizers can then specify which modes they currently support and also have their custom implementation to handle that specific class of optimization problems.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -32,21 +97,404 @@ def __init__(
...
+            objective,
+            linearization_cls=linearization_cls,
+            linearization_kwargs=linearization_kwargs,
+            **linear_solver_kwargs,
+        )
+        self.ordering = self.linear_solver.linearization.ordering
+        self._tmp_optim_vars = tuple(v.copy(new_name=v.name) for v in self.ordering)
+
+    # Modifies the (no grad) info in place to add data of grad loop info
+    def _merge_infos(
```
joeaortiz, line 122: Could this be a fn in the NonLinearOptimizer class instead? I need this fn in GBP so the alternative is that I copy this code into the GBP class.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -32,21 +97,404 @@ def __init__(
...
+            iters_done += 1
+            it_ += 1
+
+        info.status[
+            info.status == NonlinearOptimizerStatus.START
+        ] = NonlinearOptimizerStatus.MAX_ITERATIONS
+        return iters_done
+
+    # Returns how many iterations to do with and without autograd
+    def _split_backward_iters(self, **kwargs) -> Tuple[int, int]:
```
joeaortiz, line 299: Same goes here, could this be a fn of the NonLinearOptimizer class instead?

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -51,96 +49,41 @@ class NonlinearOptimizerInfo(OptimizerInfo):
...
-#           2.2. Retract all tmp_optim_vars given delta
-#           2.3. Evaluate new error
-#           2.4. reject_indices = self._complete_step(delta, new_err, prev_err)
-#           2.5. Update objective's optim var containers with retracted values,
-#                ignoring indices given by `reject_indices`
-#    3. Check convergence
+# Base class for all nonlinear optimizers.
+# Provides methods useful for bookkeeping during optimization.
+# Subclasses need to provide `error_metric()`, which computes the error to
+# to optimized (e.g., sum of squared costs for NLLS), in addition to
```
mhmukadam, line 55: typo: to to -> to be

mhmukadam (approved):
LGTM!

theseus/optimizer/nonlinear/nonlinear_least_squares.py
mhmukadam: @dishank-b, @bamos does this function get similarly used for CEM? If so then we could move it one level up. If not then we can figure out the common intersection for this function between NLS, GBP and CEM.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
luisenp: Moved back up.

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Fixed.

joeaortiz (approved):
LGTM

theseus/optimizer/nonlinear/nonlinear_optimizer.py
```diff
@@ -16,6 +17,30 @@
 from theseus.optimizer import Optimizer, OptimizerInfo
 
 
+class BackwardMode(Enum):
```
mhmukadam, line 20: I was thinking all the way up to `Optimizer` but maybe here is okay for now?

mhmukadam (approved):

theseus/optimizer/nonlinear/nonlinear_optimizer.py
luisenp: Let's leave here for now, as this is not currently relevant to any other optimizers. It'll be easy to move around if we ever need it a level up.

fantaosha (approved):

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -32,21 +72,329 @@ def __init__(
...
-            linearization_cls=linearization_cls,
-            linearization_kwargs=linearization_kwargs,
-            linear_solver_kwargs=linear_solver_kwargs,
             abs_err_tolerance=abs_err_tolerance,
             rel_err_tolerance=rel_err_tolerance,
             max_iterations=max_iterations,
             step_size=step_size,
             **kwargs,
         )
+        linear_solver_cls = linear_solver_cls or CholeskyDenseSolver
```
dishank-b, line 85: `NonlinearLeastSquares` still seems to be making assumption that solver/optimizer is using linearization to solve. There might be `Nonlinearleastsquare` optimizers not using linearization. I see that it's being done so that subclasses such as `gauss-newton` and `LM` can be derived from this. But I feel there should one more level of abstraction between NonlinearLeastSquares and GN/LM optimizer class.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -4,18 +4,58 @@
...
+#    1. compute delta
+#    2. step(delta, prev_err)
+#           2.1. Store current optim var tensors in tmp_optim_vars containers
+#           2.2. Retract all tmp_optim_vars given delta
+#           2.3. Evaluate new error
+#           2.4. reject_indices = self._complete_step(delta, new_err, prev_err)
+#           2.5. Update objective's optim var containers with retracted values,
+#                ignoring indices given by `reject_indices`
+#    3. Check convergence
 class NonlinearLeastSquares(NonlinearOptimizer, abc.ABC):
```
dishank-b, line 59: Shouldn't this class just define `_error_metric()`? rest such as `_optimize_impl()` should still be left abstracted for a new subclass?

dishank-b (commented):
Left some comments.

dishank-b (approved):

theseus/optimizer/nonlinear/nonlinear_least_squares.py
luisenp: This is a reasonable point, but I wonder if it's worth it to preemptively introduce a more complicated hierarchy for an optimizer that we don't currently have. At least for DCEM, what I'm picturing is for it to be more general than NLLS via the `error_metric` function, so it should inherit directly from `NonlinearOptimizer`. But maybe GBP? 

@joeaortiz  and @mhmukadam, is GBP meant to be used only for NLLS? If that's the case then it makes sense to do 

`NonlinearLeastSquares` --> `GBP` (NLLS only defines the error metric)
`NonlinearLeastSquares --> SomeCoolName --> GN/LM` (most of the logic would go to the intermediate class)

But otherwise, seems cleaner o leave as is until we have some optimizer for NLLS that doesn't use the type of linearization that GN and LM uses.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
luisenp: Maybe? I guess depends on how we resolve [this](https://github.com/facebookresearch/theseus/pull/440#discussion_r1088356231).

theseus/optimizer/nonlinear/nonlinear_least_squares.py
mhmukadam: Since these are optimizers, in my mind `NonlinearLeastSquares` are methods that linearize the objective and call a linear solver iteratively. The confusion might be thinking about the name as solving nonlinear least squares objectives.

Luis is correct that CEM should derive from `NonlinearOptimizer` even if it only supports nonlinear least squares objectives right now, since the objective type it could support is more general.

GBP seems tricky since it does involve local linearization and local linear solves. Since this PR is blocking CEM progress let's proceed assuming `NonlinearOptimizer` --> `GBP` and then we can make any updates in the GBP PR once @joeaortiz has more bandwidth.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -4,18 +4,58 @@
...
+# do optimization loop:
+#    1. compute delta
+#    2. step(delta, prev_err)
+#           2.1. Store current optim var tensors in tmp_optim_vars containers
+#           2.2. Retract all tmp_optim_vars given delta
+#           2.3. Evaluate new error
+#           2.4. reject_indices = self._complete_step(delta, new_err, prev_err)
+#           2.5. Update objective's optim var containers with retracted values,
+#                ignoring indices given by `reject_indices`
+#    3. Check convergence
```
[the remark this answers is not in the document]
mhmukadam: I guess we could add these to `Optimizer` more generally and then move the set of any appropriate functions in `NonlinearLeastSquares` to the appropriate pre/post bucket?

theseus/optimizer/nonlinear/nonlinear_least_squares.py
```diff
@@ -4,18 +4,58 @@
...
+EndIterCallbackType = Callable[
+    ["NonlinearOptimizer", NonlinearOptimizerInfo, torch.Tensor, int], NoReturn
+]
 
+
+# Base class for all optimizers for NLLS problems,
+# providing the skeleton of the
+# optimization loop. Subclasses need to implement the following method:
+#
+#   - `compute_delta`: returns a descent direction given the current values
```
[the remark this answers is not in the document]
mhmukadam: This (or a similar fucntion) should be in trust region?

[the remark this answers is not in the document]
luisenp: In response to Mustafa, I think line search is a more general procedure, and I'd argue it would even belong in `NonlinearOptimizer`. Also, a method is probably not the right abstraction, but something like a member with type specified at construction time, to specify which line search method to use, each implemented in a different class. 

But in any case, I'd prefer not to include anything for `line_search()` until we have a concrete implementation of a differentiable line search method with batching support, and we know what an implementation would look like.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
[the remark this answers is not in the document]
luisenp: This seems safe enough to add. Can you elaborate a bit on what the usual API for these are? (what are typical inputs/outputs?)

theseus/optimizer/nonlinear/nonlinear_least_squares.py
luisenp: Agree on doing `NonlinearOptimizer --> GBP` for now.

Also, IIRC, CEM should already very easily support other objectives, by replacing all calls current calls to error squared norm with `error_metric()`. 

That being said, and also relevant to the discussion above, it seems to me that the `error_metric()` method more appropriately belongs to `Objective` class. Otherwise, we will need to have some nonlinear optimizer subclass just so that users can pass the error metric to use, which seems strange.

mhmukadam: Yeah makes sense that error metric is a function of the objective. But do we need to make some class hierarchy for objective with the common stuff in a base class `Objective` and the NLS related stuff in a `NLSObjective`?

luisenp: Maybe, but I'd rather not go too crazy right now. We can add a subclass that just hardcodes the squared error metric, but leave all of the other stuff inside the base class. And revisit this if needed.

mhmukadam: Agreed not needed yet or in this PR.

So have we settled on merging after moving error_metric to objective or do we need anything else? cc @dishank-b

theseus/optimizer/nonlinear/nonlinear_least_squares.py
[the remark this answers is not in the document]
mhmukadam: Maybe we can discuss more with @fantaosha and circle back in a new PR? Probably flag an issue for now.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
dishank-b: Yeah, LGTM! We can merge.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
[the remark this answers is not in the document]
luisenp: Created #454

theseus/optimizer/nonlinear/nonlinear_least_squares.py
luisenp: Made #455 for this feature. If that looks good, I'll merge both PRs.

mhmukadam (approved):

theseus/optimizer/nonlinear/nonlinear_least_squares.py
dishank-b: One more thing, I will still be using the very similar `_optimize_impl` and `_optimize_loop` core structure for DCem, the different is instead of using linearization based solving i would use DCem solver. From that perspective, would it make sense to have more general `_optimize_impl` and `_optimize_loop` in `NonlinearOptimizer`? 
I can also just copy from here, but that would make it repetitive.

luisenp: I'm less clear on what a general version of these two methods would look like. Let's start by merging this PR, updating your PR with these changes, and possibly have a second round of refactoring after some review. I'd like to see what exactly is the diff between your version of `_optimize_loop` and `_optimize_impl` after we merge these changes.

theseus/optimizer/nonlinear/nonlinear_least_squares.py
dishank-b: Sounds good.

PR #450, opened 2023-01-26
Restructures Lie groups labs package and adds LieTensor

luisenp:


tests/labs/lie/functional/test_se3.py
```diff
@@ -30,7 +30,7 @@
 @pytest.mark.parametrize("batch_size", BATCH_SIZES_TO_TEST)
 @pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
 def test_op(op_name, batch_size, dtype):
-    import theseus.labs.lie_functional.se3 as se3
+    import theseus.labs.lie.functional.se3_impl as se3
```
mhmukadam, line 33: Any reason we renamed se3.py to se3_impl.py? Does `import theseus.labs.lie.functional.se3 as se3` cause issues?

luisenp: I think it's better if that file is hidden from users as much as possible, since it has a bunch of backend stuff that's not meant to be accessed. My idea is that the import that people will be using is `theseus.labs.lie.functional`, which provides the namespaces `se3` and `so3`. This `as se3` is just for the test, that's not meant to be the recommended usage.

theseus/labs/lie/functional/lie_group.py
```diff
@@ -150,3 +153,45 @@ def jop(
         return jop_autograd_fn(input0, input1)
 
     return op, jop
+
+
+_CheckFnType = Callable[[torch.Tensor], None]
+_RadnFnType = Callable[
```
mhmukadam, line 159: typo Rand?

tests/labs/lie/functional/test_se3.py
luisenp: In fact, maybe we can should rename these packages as `theseus.labs.lie.functional._impl.so3`.

theseus/labs/lie/functional/lie_group.py
luisenp: Oh weird, must have been some rebase failure.

luisenp: Fixed.

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+SE3 = LieType.SE3
+SO3 = LieType.SO3
+
+_JFnReturnType = Tuple[List[torch.Tensor], TensorType]
+
+
+class LieTensor:
+    def __init__(self, data_tensor: torch.Tensor, ltype: LieType):
```
mhmukadam, line 30: In terms of usage what is `ltype` passed as and printed as?
Can I do `x = LieTensor(data, 'SE3')` or is the correct way to use `x = LieTensor(data, lie_tensor.SE3)`
And what prints when I do `print(x.ltype)`, is it 0 from the Enum?

Maybe we add some convenience stuff if needed in `lie/__init__.py`.

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __repr__(self) -> str:
+        return f"LieTensor({self._t}, ltype=lie.{self.ltype})"
+
+    def _check_ltype(self, other: "LieTensor", op_name: str):
+        if other.ltype != self.ltype:
+            raise ValueError("f{op_name} requires both tensors to have same ltype.")
+
+    # Returns a new LieTensor with the given data and the same ltype as self
+    def new(self, t: TensorType) -> "LieTensor":
```
mhmukadam, line 47: Maybe also a `as_lietensor`?

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,112 @@
...
+_JFnReturnType = Tuple[List[torch.Tensor], TensorType]
+
+
+class LieTensor:
+    def __init__(self, data_tensor: torch.Tensor, ltype: LieType):
+        self._t = data_tensor
+        self._fn_lib = {
+            SE3: _se3_base,
+            SO3: _so3_base,
+        }[ltype]
```
mhmukadam, line 35: This could live outside the `LieTensor.__init__` function too. Not sure if there is advantage to doing it though.

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Operators
+    def exp(self, tangent_vector: torch.Tensor) -> "LieTensor":
+        return self.new(self._fn_lib.exp(tangent_vector))
+
+    def log(self, group: "LieTensor") -> torch.Tensor:
+        return self._fn_lib.log(group._t)
+
+    def adj(self, group: "LieTensor") -> "LieTensor":
+        return self.new(group)
```
mhmukadam, line 59: I know adj returns the same group but should we still call `self._fn_lib.adj` for consistency and just in case anything in the functional code changes but doesn't propagate up here.

theseus/labs/lie/lie_tensor.py
luisenp: It should be printing the string, but I forgot to add `self.ltype.name` to the format. 

To be consistent with torch, I'd prefer if people do `x = LieTensor(data, ltype=lie.SE3)` and disallow strings.

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,112 @@
...
+    def jexp(self, tangent_vector: torch.Tensor) -> _JFnReturnType:
+        return self._unary_jop_base(tangent_vector, self._fn_lib.exp)
+
+    def jlog(self, group: "LieTensor") -> _JFnReturnType:
+        return self._unary_jop_base(group._t, self._fn_lib.exp, out_is_group=False)
+
+    def jinv(self, group: "LieTensor") -> _JFnReturnType:
+        return self._unary_jop_base(group._t, self._fn_lib.exp)
+
+    def jcompose(self, other: "LieTensor") -> _JFnReturnType:
```
mhmukadam, line 108: Should we have some not implemented / not available error for the remaining jops similar to the functional code?

mhmukadam (approved):
Looks great! Added some comments.

theseus/labs/lie/lie_tensor.py
luisenp: I was thinking about something like `torch.Tensor.new_tensor(x)`, which uses the data in `x` but makes a tensor with the same `dtype` and `device`.

theseus/labs/lie/lie_tensor.py
mhmukadam: Could `lie.SE3` for type use cause confusion with any functional use via `lie.se3.` or `lie.SE3.` I don't think we have any functions in the latter namespace but do we expect any later?

tests/labs/lie/functional/test_se3.py
mhmukadam: Makes sense. I like the suggestion to add underscore to indicate 'internal'. Maybe `_se3_impl` is sufficient?

theseus/labs/lie/lie_tensor.py
luisenp: I think that defeats the purpose of having a separate `lie.functional` package for users to get lie group ops w/o using `LieTensor`. We could consider removing the "functional" subpackage, and moving that stuff one level up, but I like having the clear separation. 

Also, I have this idea that `lie.se3` and `lie.so3`, etc. would actually refer to Lie Algebra types, and `lie.SE3` and `lie.SO3` the corresponding group types (I think @maurimo was the first one that floated this). Although this is more of a loose thought right now, not sure if we'll need this.

theseus/labs/lie/lie_tensor.py
luisenp: Oh, yeah, we shouldn't be doing this, good catch!

tests/labs/lie/functional/test_se3.py
luisenp: Thinking here of how `torch` has a bunch of private [packages](https://github.com/pytorch/pytorch/tree/master/torch) in their repo.

theseus/labs/lie/lie_tensor.py
luisenp: See [this](https://github.com/facebookresearch/theseus/pull/450#discussion_r1088341427) comment.

theseus/labs/lie/lie_tensor.py
luisenp: I wasn't sure. I guess maybe better to add this for consistency.

tests/labs/lie/functional/test_se3.py
mhmukadam: Ah. `_impl` as package sounds to generic. How about `_groups`?

theseus/labs/lie/lie_tensor.py
luisenp: `self` is used to pass the `ltype` along (line 49).

theseus/labs/lie/lie_tensor.py
mhmukadam: Oh good point. I missed the `[ltype]` and was thinking this dict is a constant.

Yeah I agree we should reserve both the `lie.SE3` and `lie.se3` versions in case we do decide to add the algebra support later. We need to switch the current namespace to uppercase since lower case is algebra, right?

mhmukadam: For consistency we could rename `_se3_base` -> `_se3_impl`?

theseus/labs/lie/lie_tensor.py
luisenp: Added.

theseus/labs/lie/lie_tensor.py
luisenp: Fixed. This was also flat out wrong for SE3.

theseus/labs/lie/lie_tensor.py
mhmukadam:
```python
x = lie.LieTensor(xdata, lie.LieType.SE3)
t = torch.Tensor(tdata)

y = x.new(x) # Is this usage a bit strange?
z = x.new(t) # this seems okay

# on the other hand both below are fine
y = lie.new(x)
z = lie.new(t) # for this maybe ltype can be passed to new as an optional argument?
```

mhmukadam: Actually `lie.new(t, lie.LieType.SE3)` is basically same as `lie.LieTensor()`. I guess we can keep as is (or make new only support LieTensor not tensor)?

theseus/labs/lie/lie_tensor.py
mhmukadam: Did the tests still pass? Oh I guess we don't have tests on top of the LieTensor interface, right?

theseus/labs/lie/lie_tensor.py
luisenp: `z = x.new(y)` is meant to receive new data from y, but copy dtype, ltype, device (and other tensor properties) from x. In `torch` one could do `y = x.new_tensor(x)`, but this would be a pointless operation (`torch` throws a warning to use `clone()` instead).

luisenp: Also, not a fan of `lie.LieType`. Torch simply refers to `torch.double`, `torch.float`, etc. Why not follow this convention?

mhmukadam: Okay cool, I like the idea to follow torch here with the clone warning (maybe we can add a clone op? sorry for dragging this out too much).

Ah ignore my bad use of LieType, haha. Yes we should have it setup like you suggest with lie.SE3, etc.

luisenp: Actually, after playing around with this, I think this should have just being 'clone()' all along :) 

Following torch, `x.new()` would be more appropriate if one were to pass an arbitrary shape. For example in `torch` one can do `x.new_tensor([1, 2, 3]))`, which makes a new tensor of shape (3,) with the contents in the list.

luisenp: In the end decided to keep `new()` as well, since we are using to create a LieTensor from a torch tensor. I also added clone and a warning. Let me know if it looks OK.

theseus/labs/lie/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from .types import ltype, SE3, SO3
+from .lie_tensor import LieTensor, new
```
mhmukadam, line 6: Should we also pull functional API here so they are accessible at the `lie.` level too.

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,146 @@
...
+    def __repr__(self) -> str:
+        return f"LieTensor({self._t}, ltype=lie.{self.ltype})"
+
+    def _check_ltype(self, other: "LieTensor", op_name: str):
+        if other.ltype != self.ltype:
+            raise ValueError("f{op_name} requires both tensors to have same ltype.")
+
+    # Returns a new LieTensor that clones the given data and
+    # has the same ltype as self
+    def clone(self) -> "LieTensor":
```
mhmukadam, line 39: Should we also add a to() that internally calls the torch to()?

theseus/labs/lie/lie_tensor.py
mhmukadam: Looks good!

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,146 @@
...
+    jproject = _no_unary_op
+    jleft_act = _no_binary_op
+    jleft_project = _no_binary_op
+
+
+# ----------------------------
+# Tensor creation functions
+# ----------------------------
+def new(
+    data: Any,
```
mhmukadam, line 135: Any -> TensorType?

theseus/labs/lie/lie_tensor.py
luisenp: Yes, we can add that in a separate PR, this is getting too big.

theseus/labs/lie/lie_tensor.py
luisenp: This is what torch uses because `torch.as_tensor` can receive float, numpy array, sequences of many different types, torch tensors, and who knows what else. Handling types for this is probably a big mess.

mhmukadam: Ah okay, makes sense.

tests/labs/lie/functional/common.py
```diff
@@ -48,6 +48,11 @@ def get_test_cfg(op_name, dtype, dim, data_shape, module=None):
 
 
 # Sample inputs with the desired types.
+# Input type is one of:
+#   ("tangent", dim)
+#   ("group", module)
+#   ("quat", dim)  # like tangent but normalized
```
mhmukadam, line 54: It might be confusing to say quaternion are like tangent. Do you just mean the values are in a tensor of (..., dim)? .

theseus/labs/lie/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    LieTensor,
+    adj,
+    compose,
+    exp,
+    hat,
+    inv,
+    jcompose,
+    jexp,
+    jinv,
+    jlog,
```
mhmukadam, line 16: do we want to add the other jacobinas (jadjoint, etc) even if they all point to _no_jop at the moment?

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,272 @@
...
+    ltype: Optional[_ltype] = None,
+    dtype: torch.dtype = torch.float,
+    device: Union[torch.device, str] = None,
```
mhmukadam, line 154: Can dtype and device be optional since they have sensible defaults in the function signature? These could also possibly conflict with what a passed torch tensor might have.

theseus/labs/lie/lie_tensor.py
```diff
@@ -0,0 +1,272 @@
...
+
+def adj(group: LieTensor) -> torch.Tensor:
+    return group.adj()
+
+
+def inv(group: LieTensor) -> LieTensor:
+    return group.inv()
+
+
+def exp(ltype: _ltype, tangent_vector: torch.Tensor) -> "LieTensor":
```
mhmukadam, line 225: Curious if there is any standard convention on param first vs data first in a function call.

mhmukadam (approved):

theseus/labs/lie/__init__.py
luisenp: Now that I think about it, the question is if we plan to have support for this in the future. If some of these operations don't make sense, I think it's better to completely remove them. We can then add the remaining ones both to the class and global scopes.

I think @fantaosha can make the choice of which ones.

mhmukadam: Makes sense to differentiate ones that can/will be added vs ones that don't work theoretically. It might be a good idea to still keep around all of them and raise a different error depending on not implemented or not possible.

theseus/labs/lie/lie_tensor.py
luisenp: Not sure I understand the question. If the question is about who has priority between the passed torch tensor and the given dtype/device, I think the priority should be the explicit values passed by the user (will need to add this, not currently the case).

theseus/labs/lie/lie_tensor.py
luisenp: In torch, the dtype is usually an optional keyword for any torch creation ops; however, for us the `ltype` is not optional, so it's a bit different. I don't have a strong preference if should go first or last.

theseus/labs/lie/lie_tensor.py
luisenp: Changed.

theseus/labs/lie/lie_tensor.py
luisenp: I have added some tests now.

theseus/labs/lie/__init__.py
mhmukadam: Yeah on second thought, you are right, it doesn't seem useful.

theseus/labs/lie/lie_tensor.py
mhmukadam: Similar to the other comment, we can follow norms from torch for what a user would expect regarding which goes first.

theseus/labs/lie/__init__.py
fantaosha: > Now that I think about it, the question is if we plan to have support for this in the future. If some of these operations don't make sense, I think it's better to completely remove them. We can then add the remaining ones both to the class and global scopes.
> 
> I think @fantaosha can make the choice of which ones.

It depends. Some of these operators do not make sense at all. But some like `jadjoint` does have geometric meaning. We need to go back to this later.

theseus/labs/lie/lie_tensor.py
fantaosha: Not sure if the same `LieTensor` is good. In my view, `LieTensor` usually refers to something like the tangent vector on a Lie group. I personally prefer `LieGroup` to avoid ambiguity. Maybe we can keep `LieTensor` for the tangent vectors on Lie groups, which are usually a pair of `(group, v)`.

theseus/labs/lie/lie_tensor.py
fantaosha: Not sure if this affects the compatibility of functorch and `autograd.jacobian` with `vectorize=True`, in both cases the `tensor` is not a regular pytorch tensor.

theseus/labs/lie/lie_tensor.py
fantaosha: > In torch, the dtype is usually an optional keyword for any torch creation ops; however, for us the `ltype` is not optional, so it's a bit different. I don't have a strong preference if should go first or last.

For generality, we can treat pytorch tensor as a Lie group as well. Maybe we can create another Lie group class called `Matrix` that are essentially torch tensors.

fantaosha (commented):
Only minor comments. I'm considering to add a new Lie group type called `Matrix` which is essentially pytorch tensors.

theseus/labs/lie/__init__.py
luisenp: For now, I'll just keep the ones we have.

theseus/labs/lie/lie_tensor.py
luisenp: Addressed the `ltype` arg order in 5a72342ccb6035db2054d038e9283650e3ac54c6.

PR #452, opened 2023-01-28
Make LieTensor a subclass of torch.Tensor and override function/operators

luisenp:
This PR is mostly to illustrate some ideas on how to safely overload operators in LieTensor class. It'd better to start looking at the example script, and if that looks OK, you can take a quick look at what's happening under the hood. This is not meant to be a final implementation.


examples/lie_labs_api.py
```diff
@@ -0,0 +1,20 @@
...
+x = torch.rand(5, 6)
+
+z = g.retract(x)
+
+try:
+    y = g + x
+except RuntimeError as e:
+    print(e)
+
+y = g + lie.TangentTensor(x)
```
luisenp, line 15: Note that `TangentTensor` should be able to support arbitrary torch operations. It's a direct subclass of torch.Tensor, only with added ltype.

theseus/labs/lie/lie_tensor.py
```diff
@@ -152,18 +189,22 @@ def new(
...
+    if ltype == _ltype.tgt:
+        return TangentTensor(data)
+    return LieTensor(data, ltype=ltype)
 
 
 as_lietensor = new
 
+# With this new version of the code I like @fantaosha's proposal of using the name cast()
+# because it implies type conversion rather than a wrapper being created.
+cast = new
```
mhmukadam, line 206: Agreed!

examples/lie_labs_api.py
```diff
@@ -0,0 +1,20 @@
...
+
+torch.testing.assert_close(z._t, y._t)
```
mhmukadam, line 17: We could probably override assert close too?

examples/lie_labs_api.py
```diff
@@ -0,0 +1,20 @@
...
+
+w = g + lie.cast(x, ltype=lie.tgt)
```
mhmukadam, line 19: Related to previous comment. rand has ltype first and cast has data first.

mhmukadam (commented):
The example script looks good to me!

examples/lie_labs_api.py
luisenp: I did a quick test and with some slight modification this works if we subclass torch.Tensor (so it works for `TangentTensor`). 

I was playing around with subclassing LieTensor over the weekend, but had some issues with the constructor not being compatible with the superclass. I'll keep looking to see if there is a way to make this work.

examples/lie_labs_api.py
luisenp: The funny things is that what we probably want is exactly the opposite. `cast` receive type first, so that it's consistent with `typing.cast`, and `rand` has ltype after size, so that it's more consistent with `torch` (not 100% the same, because `dtype` is optional in torch, but not for us). 

The reason I put `ltype` first in `rand` was not API, but that I was having some errors due to the use of `*size`. But I think I can fix this.

theseus/labs/lie/lie_tensor.py
```diff
@@ -143,6 +166,20 @@ def _no_jop(self, input0: TensorType) -> _JFnReturnType:
...
 
+    def retract(self, delta: TensorType) -> "LieTensor":
+        assert isinstance(delta, torch.Tensor)
+        return self.compose(self.exp(delta))
+
+    def __add__(self, other: TensorType) -> "LieTensor":
+        if not isinstance(other, TangentTensor):
+            raise RuntimeError(
+                "Operator + is only supported for tensors of ltype=tgt. "
+                "If you intend to add the raw tensor data, please use group._t. "
```
luisenp, line 177: Rather than using `torch._t`, we should provide some method that return the group's data as a raw torch tensor (e.g., `g.mat()` or maybe `g.as_torch(fmt="mat")`.

examples/lie_labs_api.py
luisenp: Fixed `rand` in the other PR.

theseus/labs/lie/lie_tensor.py
mhmukadam: `as_tensor` with an optional format passed, sounds good. Could you use cast here?

luisenp:
One other detail to consider for discussion. I did some quick tests with torch optimizers, and could only get the subclass based version to work by doing the following, which requires an extra step. We also need to do some tests to make sure backprop is doing the right thing. Curious to hear thoughts from @albanD

```python
import theseus.labs.lie as lie
import theseus.labs.lie.functional as lieF

# Leaf tensor needs to be a regular tensor, so we need to explicitly pass the tensor data
g1_data = lieF.se3.rand(1, requires_grad=True)
g1 = lie.cast(g1_data, ltype=lie.SE3)
g2 = lie.rand(1, lie.SE3)
opt = torch.optim.Adam([g1_data])

opt.zero_grad()
d = g1.inv().compose(g2).log()
loss = torch.sum(d**2)
loss.backward()
opt.step()

```

luisenp:
cc @rmurai0610

theseus/labs/lie/lie_tensor.py
```diff
@@ -32,35 +57,115 @@ def _get_fn_lib(ltype: _ltype):
...
 
 
-class LieTensor:
-    def __init__(self, data_tensor: torch.Tensor, ltype: _ltype):
-        self._t = data_tensor
-        self._fn_lib = _get_fn_lib(ltype)
-        self._fn_lib.check_group_tensor(data_tensor)
+class _LieTensorBase(torch.Tensor):
+    def __new__(cls, data: Any, ltype: _ltype, requires_grad=None):
+        if requires_grad is None:
```
albanD, line 62: I would not have this case and just the one with `_make_subclass` below. It will make your autograd story simpler:
`lt = _LieTensorBase(t)` is never a differentiable op (always creates a leaf). And so lt never has an autograd graph. You can use `requires_grad` flag just like `torch.Tensor()` though to make it require grad if you want.

If you want a way to create your class in a differentiable way, use a custom method implemented with a custom autograd Function such as `lt = _LieTensorBase.from_tensor(t)` which will always be differentiable.

theseus/labs/lie/lie_tensor.py
```diff
@@ -32,35 +57,115 @@ def _get_fn_lib(ltype: _ltype):
...
-    def __repr__(self) -> str:
+    def __repr__(self) -> str:  # type: ignore
         return f"LieTensor({self._t}, ltype=lie.{self.ltype})"
 
+    @classmethod
+    def to_torch(cls, args: Any):
+        def _resolve(x):
+            return x._t if isinstance(x, cls) else x
+
+        return tree_map(_resolve, args)
```
albanD, line 82: Equivalently
```suggestion
    def to_torch(cls, args: Any):
        return tree_map_only(cls, lambda x: x._t, args)
```

theseus/labs/lie/lie_tensor.py
```diff
@@ -32,35 +57,115 @@ def _get_fn_lib(ltype: _ltype):
...
+        ltypes = [cls.resolve_ltype(a) for a in tree_flatten(args)[0]]
+        ltypes = set(x for x in ltypes if x is not None)
+        if len(ltypes) > 1:
+            raise ValueError(
+                f"All LieTensors must be of the same ltype. " f"But found {ltypes}"
+            )
+        ltype = next(iter(ltypes)) if len(ltypes) > 0 else None
+        ret = func(*torch_args, **torch_kwargs)
+        return (
+            cls(ret, ltype) if isinstance(ret, torch.Tensor) and not raw_tensor else ret
```
albanD, line 102: nit: you most likely want a tree_map_only(torch.Tensor, lambda t: cls(t, ltype), ret)` here to cover functions returning multiple things.

examples/lie_labs_api.py
```diff
@@ -0,0 +1,76 @@
...
+g = lie.rand(5, lie.SE3)
+print("LieTensor is tensor", isinstance(g, torch.Tensor), "\n")
+
+# Can retract standard torch tensors
+x = torch.rand(5, 6)
+z = g.retract(x)
+
+# However, for safety, one cannot use overriden + with torch tensors
+try:
+    y = g + x
```
fantaosha, line 16: Honestly, I'm a little nervous about overriding `__add()` since they are very dangerous. Also, one can not use `*`, `-`, `/`, either in that case.

fantaosha: Do we have some operators called `to_euclidean()` that makes a pytorch tensor copy (not reference) of current data.

fantaosha: In summary, I would like exactly function names rather than overrides to avoid confusion. I encountered lots of problems to generalize the first-order optimizers on Lie groups.

fantaosha: To be consistent, if `g + x = g.compose(x)`, then `g-x` should be `g.compose(x.inverse())`.

examples/lie_labs_api.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+# Unsupported operations throw a NotImplementedError
+try:
+    assert torch.allclose(w, z)
+except NotImplementedError as e:
+    print(e)
+
+# But you can add lie.as_euclidean() and then everything is valid
+with lie.as_euclidean():
```
fantaosha, line 68: I'm concerned with this. If the user wants to use Lie group as a Euclidean tensor. Maybe it is safer to do something like `g.to_euclidean().matmul(torch.rand(4,7)`.  Maybe `to_euclidean()` should be a property such that users can not modify the data.

theseus/labs/lie/lie_tensor.py
fantaosha: `mat` seems to have different meanings for Lie groups, which depends on the representation. For example, if we use quaternions to represent SO3, `g.mat()` should be a `3x3` matrix rather than a `4x1` vector.

theseus/labs/lie/types.py
```diff
@@ -8,7 +8,9 @@
 class ltype(Enum):
     SE3 = 0
     SO3 = 1
+    tgt = 2
 
 
 SE3 = ltype.SE3
 SO3 = ltype.SO3
+tgt = ltype.tgt
```
fantaosha, line 16: Different Lie groups should have different types of tangent vectors.

fantaosha (commented):
Some comments. Not sure if we should override `__add()` and use `LieAsEuclidean`.

examples/lie_labs_api.py
luisenp: Is the concern that users would override the internal data? (e.g., something like `g.add_(something)`? This seems a reasonable concern, because I believe subclassing allows this. I'm OK with replacing this with `to_euclidean()`.

theseus/labs/lie/lie_tensor.py
luisenp: `mat` means **always** return a matrix independently of what the internal representation is. My vision is the actual representation is hidden from the user, and they can get access to tensor data with something like `mat()`, `quat()`, etc. BTW, I just realized that this also means that we don't need `to_euclidean()`, methods such as `mat()` and `quat()` would cover this functionality.

theseus/labs/lie/types.py
luisenp: Can you elaborate? Do they differ in anything other than their dimension?

theseus/labs/lie/lie_tensor.py
luisenp: Awesome, changed!

theseus/labs/lie/lie_tensor.py
```diff
@@ -143,6 +314,67 @@ def _no_jop(self, input0: TensorType) -> _JFnReturnType:
...
+            raise RuntimeError(
+                "LieTensor.set_ is only supported for LieTensor arguments."
+            )
+        if not tensor.ltype == self.ltype:
+            raise RuntimeError(
+                f"Tried to set a tensor of type {self.ltype} with a "
+                f"tensor of type {tensor.ltype}"
+            )
+        super().set_(tensor)  # type: ignore
+
```
luisenp, line 345: A less esoteric but perhaps less safe alternative for `add_` could be to check the tensor shape and if it looks like (...,3, 4) (for SE3) this is interpreted as an euclidean gradient, something with shape `(..., 6)`interpret this as a Riemannian graadient, and everything else throws an error.

theseus/labs/lie/types.py
mhmukadam: Is this resolved? It would be useful to get feedback during alpha release on the tangent stuff and it's usage.

theseus/labs/lie/lie_tensor.py
mhmukadam: Sounds good that direct access to internal storage is hidden and mat/quat return appropriate sized tensors. But for us internally are we storing things as mat or quat, and is this a setting exposed to user in case they want to switch?

theseus/labs/lie/lie_tensor.py
```diff
@@ -143,26 +337,130 @@ def _no_jop(self, input0: TensorType) -> _JFnReturnType:
...
+            raise ValueError(
+                f"Incorrect ltype for local. Expected {self.ltype}, "
+                f"received {other.ltype}."
+            )
+        return self.inv().compose(other).log()
+
+    # ------------------------------------------------------
+    # Overloaded python and torch operators
+    # ------------------------------------------------------
+    def __add__(self, other: TensorType) -> "LieTensor":
```
mhmukadam, line 363: There was some concern on overloading `+` etc since it could mean different things. Is that resolved? Taosha pointed out that Sophus for instance only overrides `*` and no other operators.

Since we have this stuff added already we can try to get explicit feedback from users on this in the alpha release and then decide if we want to adapt or scale back.

theseus/labs/lie/lie_tensor.py
```diff
@@ -143,26 +337,130 @@ def _no_jop(self, input0: TensorType) -> _JFnReturnType:
...
     if isinstance(data, LieTensor):
-        return data.new(data)
+        return data
     if ltype is None:
         raise ValueError("ltype must be provided.")
-    if not isinstance(data, torch.Tensor):
-        data = torch.as_tensor(data, dtype=dtype, device=device)
-    return LieTensor(data, ltype=data.ltype)
+    if ltype == _ltype.tgt:
+        return _TangentTensor(data, None)
```
mhmukadam, line 443: If explicit dtype and device are passed to as_lietensor then these would be missed when making a tangent tensor from data alone?

theseus/labs/lie/lie_tensor.py
```diff
@@ -32,44 +59,209 @@ def _get_fn_lib(ltype: _ltype):
...
 
-    # Returns a new LieTensor that clones the given data and
-    # has the same ltype as self
-    def clone(self) -> "LieTensor":
-        return LieTensor(self._t.clone(), ltype=self.ltype)
-
     # Returns a new LieTensor with the given data and the same ltype as self
-    def new(self, t: TensorType) -> "LieTensor":
-        if isinstance(t, LieTensor):
+    def new(self, args: Any, device: Device = None) -> "LieTensor":  # type: ignore
```
mhmukadam, line 242: Does dtype need to be an argument?

mhmukadam (commented):
Looks great! Let's merge once the open comments are resolved.

theseus/labs/lie/lie_tensor.py
mhmukadam: Offline discussed that tangent tensor might get removed.

theseus/labs/lie/lie_tensor.py
mhmukadam: Will resolve in example/alpha PR.

theseus/labs/lie/types.py
mhmukadam: Converged that we will likely remove tangent tensor stuff. Discussing in example/alpha PR.

mhmukadam (approved):
LGTM!

PR #456, opened 2023-02-03
Add a unit test to check the PGO loss values automatically

luisenp:


theseus/utils/examples/pose_graph/dataset.py
```diff
@@ -135,7 +135,7 @@ def read_2D_g2o_file(
 
                 sel = [0, 3, 5]
                 weight = th.Variable(
-                    torch.from_numpy(np.array(tokens[6:], dtype=np.float64)[sel])
+                    torch.from_numpy(np.array(1, tokens[6:], dtype=np.float64)[sel])
```
fantaosha, line 138: Why do we need this change?

examples/configs/pose_graph/pose_graph_synthetic.yaml
```diff
@@ -10,16 +11,15 @@ loop_closure_outlier_ratio: 0.25
 dataset_size: 256 
 batch_size: 128
 
-
 inner_optim:
   optimizer_cls: LevenbergMarquardt
+  linear_solver_cls: LUCudaSparseSolver
```
fantaosha, line 16: Wonder if we should change the default solver to be `BaSpaCho` since it is much faster.

tests/test_pgo_benchmark.py
```diff
@@ -0,0 +1,53 @@
...
+    cfg.num_poses = 64
+    cfg.profile = False
+    cfg.savemat = False
+    cfg.inner_optim.optimizer_kwargs.verbose = False
+    return cfg
+
+
+@pytest.mark.parametrize(
+    "linear_solver_cls",
+    ["CholeskyDenseSolver", "LUCudaSparseSolver", "CholmodSparseSolver"],
```
fantaosha, line 23: Do we need to add `BaSpaCho`?

fantaosha (approved):
LGTM. Maybe set `BaSpaCho` as the default solver.

theseus/utils/examples/pose_graph/dataset.py
luisenp: Explicitly adding a batch dimension removes a warning from `DiagonalCostWeight` that the shape is ambiguous.

examples/configs/pose_graph/pose_graph_synthetic.yaml
luisenp: This is set to this one because Baspacho is an optional install, so this should work for a larger set of users by default.

tests/test_pgo_benchmark.py
luisenp: I think I was going to add another test just for baspacho (to mark `run_if_baspacho`), but forgot to do it. I'll add, thanks.

fantaosha: Got it. Feel free to merge.

mhmukadam (approved):
Looks great, this is super useful! Feel free to merge once you add the baspacho block.

PR #461, opened 2023-02-09
add auto check jacobian

xphsean12:
## Motivation and Context
Add an example on auto-checking Jacobian by using the AutoDiffCostFunction, which requires only few changes at the original customized Cost function

## Types of changes

## Checklist

### DIFF
--- a/tutorials/03_custom_cost_functions.ipynb
+++ b/tutorials/03_custom_cost_functions.ipynb
@@ -29,7 +29,12 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2023-02-09T04:54:05.474667Z",
+     "start_time": "2023-02-09T04:54:04.903036Z"
+    }
+   },
    "outputs": [],
    "source": [
     "from typing import List, Optional, Tuple\n",
@@ -81,7 +86,12 @@
   {
    "cell_type": "code",
    "execution_count": 2,
-   "metadata": {},
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2023-02-09T04:54:05.482105Z",
+     "start_time": "2023-02-09T04:54:05.476902Z"
+    }
+   },
    "outputs": [],
    "source": [
     "import torch \n",
@@ -139,7 +149,12 @@
   {
    "cell_type": "code",
    "execution_count": 3,
-   "metadata": {},
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2023-02-09T04:54:05.506687Z",
+     "start_time": "2023-02-09T04:54:05.483181Z"
+    }
+   },
    "outputs": [
     {
      "name": "stdout",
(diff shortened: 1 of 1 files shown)


luisenp:
Hi @xphsean12, thanks a lot for the contribution! It would be nice if the auto check example would use the same cost function we have been using in the notebook, to avoid introducing extra complexity beyond the goals of this tutorial. 

Also, looking at the logic you are following, I think we should be able to add a generic utility function that would do the check against autograd jacobians automatically, and then we could just call this in the notebook, rather than explaining to users how to do this check manually on their own. 

If this is something you'd be interested in adding, happy to let you drive it, otherwise I can take a look at it in the next few days. We can then update this PR once that part is ready. Let me know your preference.

luisenp:
@xphsean12 following up on the comment above. Let me know if adding the generic function for autochecking is something you are interested in doing. Thanks!

xphsean12:
> Hi @xphsean12, thanks a lot for the contribution! It would be nice if the auto check example would use the same cost function we have been using in the notebook, to avoid introducing extra complexity beyond the goals of this tutorial.
> 
> Also, looking at the logic you are following, I think we should be able to add a generic utility function that would do the check against autograd jacobians automatically, and then we could just call this in the notebook, rather than explaining to users how to do this check manually on their own.
> 
> If this is something you'd be interested in adding, happy to let you drive it, otherwise I can take a look at it in the next few days. We can then update this PR once that part is ready. Let me know your preference.

Thanks for your comments and great efforts on this project. I have submitte the CLA, but I don't know how to check it. For your advices on adding a generic utility, I am confused on the implementation in several aspects: 
1. When should the auto-checking be conducted. For example, the moment when the user creates the CostFunction or Objective? 
2. How many times should the auto-checking be conducted? Because checking jacobian requires extra calculation. I suppose most of the users will only check it for the very first time when the CostFunction is created.
Regarding the above question, I believe it would be much more proper to make it generic on your own. Since it requires adding more components into the CostFunction or Objective layer. By the way, I check it in the process of adding CostFunction into Objective, because I do not need to create any artifact data by myself. I can just use the data that define the optimization problem.

luisenp:
> > Hi @xphsean12, thanks a lot for the contribution! It would be nice if the auto check example would use the same cost function we have been using in the notebook, to avoid introducing extra complexity beyond the goals of this tutorial.
> > Also, looking at the logic you are following, I think we should be able to add a generic utility function that would do the check against autograd jacobians automatically, and then we could just call this in the notebook, rather than explaining to users how to do this check manually on their own.
> > If this is something you'd be interested in adding, happy to let you drive it, otherwise I can take a look at it in the next few days. We can then update this PR once that part is ready. Let me know your preference.
> 
> Thanks for your comments and great efforts on this project. I have submitte the CLA, but I don't know how to check it. For your advices on adding a generic utility, I am confused on the implementation in several aspects:
> 
> 1. When should the auto-checking be conducted. For example, the moment when the user creates the CostFunction or Objective?
> 2. How many times should the auto-checking be conducted? Because checking jacobian requires extra calculation. I suppose most of the users will only check it for the very first time when the CostFunction is created.
>    Regarding the above question, I believe it would be much more proper to make it generic on your own. Since it requires adding more components into the CostFunction or Objective layer. By the way, I check it in the process of adding CostFunction into Objective, because I do not need to create any artifact data by myself. I can just use the data that define the optimization problem.

For now I was just thinking of creating the utility that receives a `CostFunction` object, runs the check and returns a boolean whether the check passes or not, and if it doesn't, prints out some information about the error. For the notebook, we would just need to call this utility manually.

Following on your last comment, I'll write a draft of this on my own. Is it OK if I tag you as a reviewer in the resulting PR?

Thanks!

xphsean12:
> 

Sure. It is my pleasure.

mhmukadam:
Let's rebase this on main after #465 is merged and use the new utility function in the notebook to merge this PR.

PR #465, opened 2023-02-20
Added utility to automatically check jacobians of a given cost function.

### LINKED ISSUE #321: Automated check for Cost's jacobian method
## 🚀 Feature
Automated check for Cost's jacobian method

## Motivation
Hand-writing jacobians is complex an error prone. We might still do it because it's faster than using numeric_jacobian or automated differentiation. But it should be easy to check that our implementation is correct.

## Pitch
We can add a `verify_jacobians` method in the optimizer, which will evaluate 100 costs per type, computing both the numeric_jacobian and the hand-written jacobian method, and verify if they are correct. It might be silent if they match very well eg < 1e-6, and compute and print maximum error otherwise (the user might decide to ignore the problem as the numeric jacobian might occasionally be unstable).

## Alternatives
None I'm aware of.

## Additional context
We should not have to compute manually derivatives anymore. If we do, it should be easy to check they are correct ;)

luisenp:
Closes #321. 

Not integrated with `TheseusLayer` yet, but I think we can set this up so that the check is (optionally) performed once for each cost function right after the objective's version is fixed. 

cc @xphsean12 as related to #461

### DIFF
--- a/tests/test_theseus_layer.py
+++ b/tests/test_theseus_layer.py
@@ -196,13 +196,15 @@ def _run_optimizer_test(
     force_vectorization=False,
     max_iterations=10,
     lr=0.075,
+    loss_ratio_target=0.01,
 ):
     device = "cuda:0" if torch.cuda.is_available() else "cpu"
     print(f"_run_test_for: {device}")
     print(
         f"testing for optimizer {nonlinear_optimizer_cls.__name__}, "
         f"cost weight modeled as {cost_weight_model}, "
-        f"linear solver {linear_solver_cls.__name__}"
+        f"linear solver {linear_solver_cls.__name__} "
+        f"learning method {learning_method}"
     )
 
     rng = torch.Generator(device=device)
@@ -286,6 +288,7 @@ def _run_optimizer_test(
         max_iterations=max_iterations,
     )
     layer_to_learn.to(device)
+    layer_to_learn.verify_jacobians()
 
     # Check the initial solution quality to check how much has loss improved later
 
@@ -372,10 +375,11 @@ def _run_optimizer_test(
             loss = mse_loss
 
         loss.backward()
-        print("Loss: ", loss.item())
         optimizer.step()
 
-        if mse_loss.item() / loss0 < 1e-2:
+        loss_ratio = mse_loss.item() / loss0
+        print("Loss: ", mse_loss.item(), ". Loss ratio: ", loss_ratio)
+        if loss_ratio < loss_ratio_target:
             solved = True
             break
     assert solved
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -9,6 +9,7 @@ import scipy.sparse
 import torch
 import torch.nn as nn
 
+import theseus as th
 import theseus.utils as thutils
 
 
@@ -154,3 +155,16 @@ def test_sparse_mv_cpu(batch_size, num_rows, num_cols, fill):
 @pytest.mark.parametrize("fill", [0.1, 0.9])
 def test_sparse_mv_cuda(batch_size, num_rows, num_cols, fill):
     _check_sparse_mv_and_mtv(batch_size, num_rows, num_cols, fill, "cuda:0")
+
+
+def test_jacobians_check():
+    se3s = [th.SE3() for _ in range(3)]
+    w = th.ScaleCostWeight(0.5)
+    cf = th.Difference(se3s[0], se3s[1], w)
+    thutils.check_jacobians(cf, 1)
+
+    cf = th.Between(se3s[0], se3s[1], se3s[2], w)
+    thutils.check_jacobians(cf, 1)
+
+    cf = th.eb.DoubleIntegrator(se3s[0], th.Vector(6), se3s[1], th.Vector(6), 1.0, w)
+    thutils.check_jacobians(cf, 1)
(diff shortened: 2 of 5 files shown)


xphsean12 (commented):
I copy this function and run it in my local Jupyter, with 0.1.4 version of theseus. An error occurs in Line 153, showing "property' object is not iterable"

Besides, should Line 154 be `aux_var = list(cf.aux_vars)`?

luisenp:
> I copy this function and run it in my local Jupyter, with 0.1.4 version of theseus. An error occurs in Line 153, showing "property' object is not iterable"
> 
> Besides, should Line 154 be `aux_var = list(cf.aux_vars)`?

Good catch, fixed this in 1a0047bbb67692eab7036bbbf5e1a85602488806. Can you post a screen shot of the error you are getting? I wrote a unit tests for this code and it seems to be working will with our existing cost functions.

xphsean12:
> > I copy this function and run it in my local Jupyter, with 0.1.4 version of theseus. An error occurs in Line 153, showing "property' object is not iterable"
> > Besides, should Line 154 be `aux_var = list(cf.aux_vars)`?
> 
> Good catch, fixed this in [1a0047b](https://github.com/facebookresearch/theseus/commit/1a0047bbb67692eab7036bbbf5e1a85602488806). Can you post a screen shot of the error you are getting? I wrote a unit test for this code and it seems to be working will with our existing cost functions.

image
```
class toyCost(CostFunction):
    def __init__(
        self,
        Pos_x: Vector,
        Pos_y: Vector,

        Range: Vector, 
        BasePos_x: Vector,
        BasePos_y: Vector,


        cost_weight: CostWeight,
        
        name: Optional[str] = None,
    ):
        super().__init__(cost_weight, name=name)
     

        self.Pos_x = Pos_x
        self.Pos_y = Pos_y
        
        self.Range = as_variable(Range)
        self.BasePos_x = as_variable(BasePos_x)
        self.BasePos_y = as_variable(BasePos_y)

        self.register_optim_vars(["Pos_x", "Pos_y"])
       
        self.register_aux_vars(["Range", "BasePos_x", "BasePos_y"])
        self.weight = cost_weight


    # Add this function for autochecking Jocobian. What you need to do is to warp the inputs to two types: optim_vars and aux_vars    
    def err_func_numeric(self, optim_vars, aux_vars):
        Pos_x,Pos_y = optim_vars
        Pos_x = Pos_x.tensor
        Pos_y = Pos_y.tensor
        
        Range, BasePos_x, BasePos_y = aux_vars
        Range = Range.tensor
        BasePos_x = BasePos_x.tensor
        BasePos_y = BasePos_y.tensor
        
        # Copy it from the self-defined cost function _err_func
        distance_0 = ((Pos_x - BasePos_x).square()+ (Pos_y - BasePos_y).square()).sqrt()


        err = distance_0 - Range

        return err
    
    
    def _err_func(self):
        
        Pos_x = self.Pos_x.tensor
        Pos_y = self.Pos_y.tensor
        Range = self.Range.tensor
        BasePos_x = self.BasePos_x.tensor
        BasePos_y = self.BasePos_y.tensor
      
        distance_0 = ((Pos_x - BasePos_x).square()+ (Pos_y - BasePos_y).square()).sqrt()


        err = distance_0 - Range
        print(err.shape)
       
        #print(err_0.shape)
        return err
    
    
    def dim(self):
        return BasePos_x.shape[1]

    def error(self) -> torch.Tensor:
        return self._err_func()

        batch_size = self.Range.shape[0]
        
        
        dtype = self.Range.dtype
        device = self.Range.device
        

        
        Pos_x = self.Pos_x.tensor
        Pos_y = self.Pos_y.tensor
        

        Range = self.Range.tensor
        BasePos_x = self.BasePos_x.tensor
        BasePos_y = self.BasePos_y.tensor
        dof = BasePos_x.shape[1]
        
        interTerm = (Pos_x - BasePos_x).square()+ (Pos_y - BasePos_y).square()
        
        
        Jacob_Pos_x = torch.zeros(batch_size, dof, 1, dtype=dtype, device=device)
        Jacob_Pos_y = torch.zeros(batch_size, dof, 1, dtype=dtype, device=device)
        
        dErr_x = torch.pow(interTerm, -0.5) * (Pos_x - BasePos_x)

        dErr_y = torch.pow(interTerm, -0.5) * (Pos_y - BasePos_y)
       
        Jacob_Pos_x[:,:,0] = dErr_x
        Jacob_Pos_y[:,:,0] = dErr_y
        
        error = self.error()
        
        return [Jacob_Pos_x, Jacob_Pos_y], error

    def _copy_impl(self, new_name: Optional[str] = None) -> "toyCost":
        return toyCost(
            self.Pos_x.copy(),
            self.Pos_y.copy(),
            self.Range.copy(),
            self.BasePos_x.copy(),
            self.BasePos_y.copy(),
            self.weight.copy(),
            name=new_name,
        )
```

mhmukadam (approved):
LGTM! If we want to add improvements like you mention then we can flag an issue to switch to passing class name instead of an instance.

I am not sure we should enable some check for the full objective since it would slow down the construction with unnecessary checks especially if there are many cost functions of the same type in the objective or check cost functions in the library that are already verified in unit tests. Could we assume a user would just use this function adhoc when they need it for their custom cost functions?

luisenp:
> LGTM! If we want to add improvements like you mention then we can flag an issue to switch to passing class name instead of an instance.
> 
> I am not sure we should enable some check for the full objective since it would slow down the construction with unnecessary checks especially if there are many cost functions of the same type in the objective or check cost functions in the library that are already verified in unit tests. Could we assume a user would just use this function adhoc when they need it for their custom cost functions?

I'll at least add a method one can run on the full objective on-demand, even if it's slow. I think this would be the most convenient API for this feature.

mhmukadam (approved):
LGTM! On demand checks is a good compromise than always silently running the checks under the hood.

PR #467, opened 2023-02-21
Modified build scripts to support nightly packages.

luisenp:


mhmukadam (commented):
LGTM! Should we include any update to readme instructions about nightly usage?

luisenp:
> LGTM! Should we include any update to readme instructions about nightly usage?

Yeah, but I would prefer to do it once we have a CI job that consistently pushes once EOD (maybe next week). I wanted to get this in first to unblock the other things we are working on.

PR #469, opened 2023-02-24
Minor fixes to motion planner code.

luisenp:


theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -115,7 +115,8 @@ def __init__(
         # ------------------------------- Cost weights ------------------------------ #
         # --------------------------------------------------------------------------- #
         # For the GP cost functions, we create a single GPCost weight
-        gp_cost_weight = th.eb.GPCostWeight(torch.tensor(Qc_inv, dtype=dtype), dt)
+        qc_inv_tensor = torch.as_tensor(Qc_inv, dtype=dtype)
+        gp_cost_weight = th.eb.GPCostWeight(qc_inv_tensor.to(dtype=dtype), dt)
```
mhmukadam, line 119: You probably don't need the `.to` since the previous line already sets the dtype, right?

luisenp: Ah yes, good catch.

mhmukadam (commented):
LGTM!

PR #474, opened 2023-03-01
Added planning example with SE2 and nonholonomic cost

luisenp:


theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -53,6 +53,57 @@ def _copy_impl(self, new_name: Optional[str] = None) -> "_XYDifference":
         )
 
 
+class Nonholonomic(th.CostFunction):
```
mhmukadam, line 56: Maybe we should move this to embodied/motionmodel/

theseus/utils/examples/motion_planning/motion_planner.py
```diff
@@ -53,6 +53,57 @@ def _copy_impl(self, new_name: Optional[str] = None) -> "_XYDifference":
...
+        self.weight = cost_weight
+
+    def dim(self):
+        return 1
+
+    def _compute_error(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        vel = self.vel.tensor.unsqueeze(2)
+        cos_v_theta = vel[:, 2].cos()
+        sin_v_theta = vel[:, 2].sin()
+        error = vel[:, 1] * cos_v_theta - vel[:, 0] * sin_v_theta
```
mhmukadam, line 79: I think this might not be entirely correct. Given
pos = [x, y, theta] (or SE2) and vel = [vx, vy, vtheta]
We want
error = vy * cos(theta) - vx * sin(theta)

Ref [Slide 6, [here](https://www.cs.cmu.edu/~16311/current/schedule/ppp/Lec18-Non_Holonomic_Constraints.pdf)], and [gpmp2 code](https://github.com/gtrll/gpmp2/blob/main/gpmp2/dynamics/VehicleDynamics.h#L39).

The current version is using vtheta instead of theta; this cost function will also need pose as an input aside from velocity.

mhmukadam (commented):
We might need to modify the example to verify the cost function works, but we should also add a unit test for the cost function itself.

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: IIRC, I had problems with the version that uses `theta` instead of `vtheta`, but it's been a while. Also, maybe this change was related to the formulation where we only had poses and not velocities. I'll revisit this with this example.

mhmukadam: Sounds good on revisiting. Is there a reference for the version that uses vtheta? I am not sure mathematically we can replace theta for vtheta.

tutorials/se2_planning.py
```diff
@@ -0,0 +1,104 @@
+import random
```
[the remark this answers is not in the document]
mhmukadam: Yeah might as well since we have it now :)

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: I don't remember what was my rationale back then. I remember looking at [this](http://robotics.caltech.edu/wiki/images/4/4b/IntroNoholonomicMotionPlanning.pdf) among other things, but I'm looking at it right now and the current code doesn't really make sense lol. 

Back then we were changing a lot both this and the `Dynamics` cost, and adding/removing velocities, so maybe this made sense for some version of the code that no longer exists. Not really sure.

mhmukadam: Oh lol no worries. Good we can get it fixed now.

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: Moved in 1918a240e39eed6b2186fd3c3bc17c860a445e75.

theseus/utils/examples/motion_planning/motion_planner.py
luisenp: Fixed now. Implemented in 3010936f70e0e1e2f58899c44703344222356366. Unit tests in 7413a708b16acee8abd9e96b453e25564897bb68.

tutorials/se2_planning.py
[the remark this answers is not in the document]
luisenp: Moved in a69893168e040488e1823c93cd2ec6f2c4c21176.

theseus/embodied/motionmodel/misc.py
```diff
@@ -0,0 +1,160 @@
...
+
+from typing import List, Optional, Sequence, Tuple, Union
+
+import torch
+
+from theseus.core import CostFunction, CostWeight, Variable, as_variable
+from theseus.geometry import SE2, Point3, Vector
+
+
+class HingeLoss(CostFunction):
```
luisenp, line 14: This should be move to a more general folder (along with `th.Local` implementation).

examples/README.md
```diff
@@ -6,6 +6,8 @@ noisy GPS and odometry sensors. In this example the noise is state dependent, an
 learn the cost weight as a function of pose. 
 - **motion_planning_2d.py:** Is an example of how to set up a differentiable motion planning
 problem, inspired by [Bhardwaj et al. 2020](https://arxiv.org/pdf/1907.09591.pdf).
+- **se2_planning.py:** Brief example of how to use SE2 variables with the motion planner
+utility, to handle poses with angles in 2D maps.
```
mhmukadam, line 10: Also add that it handles nonholonomic cost.

theseus/embodied/__init__.py
```diff
@@ -11,5 +11,7 @@
     DoubleIntegrator,
     GPCostWeight,
     GPMotionModel,
+    HingeLoss,
```
mhmukadam, line 14: Since HingeLoss is derived from CostFunction should we name it HingeCost or just Hinge? The loss part might indicate this is just the error function or similar to robust loss vs robust cost function.

theseus/embodied/motionmodel/misc.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def dim(self):
+        if self.dims is not None:
+            return len(self.dims)
+        return self.vector.dof()
+
+    def _compute_error(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        vector = self.vector.tensor
+        limit = self.limit.tensor
+        threshold = self.threshold.tensor
```
mhmukadam, line 49: Each dim in vector has it's own limit and threshold that could be same or different than other dims, right?

I am guessing if you wanted to not apply any hinge to a specific dimension you could just pass a inf/max or something for limit and threshold for that dim.

theseus/embodied/motionmodel/misc.py
```diff
@@ -0,0 +1,160 @@
...
+        self,
+        pose: Union[SE2, Point3, Vector],
+        vel: Union[Point3, Vector],
+        cost_weight: CostWeight,
+        name: Optional[str] = None,
+    ):
+        super().__init__(cost_weight, name=name)
+        if vel.dof() != 3 or pose.dof() != 3:
+            raise ValueError(
+                "Nonholonomic only accepts 3D velocity or poses (x, y, z dims). "
```
mhmukadam, line 110: x, y, theta dims

theseus/embodied/motionmodel/misc.py
```diff
@@ -0,0 +1,160 @@
...
+        # Pre-allocate jacobian tensors
+        batch_size = self.vel.shape[0]
+        Jvel = self.vel.tensor.new_zeros(batch_size, 1, 3)
+        error, cos, sin = self._compute_error()
+        if isinstance(self.pose, SE2):
+            Jpose = error.new_zeros(error.shape[0], 1, 3)
+            Jvel = error.new_zeros(error.shape[0], 1, 3)
+            Jvel[:, 0, 1] = 1
+        else:
+            Jpose = self.vel.tensor.new_zeros(self.vel.shape[0], 1, 3)
```
mhmukadam, line 147: Does this if/else branching make it incompatible with our vectorization?

tests/embodied/motionmodel/test_misc.py
```diff
@@ -0,0 +1,61 @@
+import theseus as th
+import torch
+
+from theseus.utils import check_jacobians
+
+
+def test_nonholonomic():
```
mhmukadam, line 7: When we add any new custom costs like nonholonomic or hinge do we need to also test our vectorization for them or does the current unit test cover any derived classes on CostFunction?

mhmukadam (commented):
Looks good to me, thanks for reviving and getting this working! A few comments to discuss and then we can merge.

examples/se2_planning.py
```diff
@@ -0,0 +1,105 @@
+import random
```
mhmukadam, line 1: license header missing

theseus/embodied/__init__.py
luisenp: Good point about the word loss, but not totally sure about just `Hinge` either. Maybe `HingeCost` is fine for this one?

theseus/embodied/motionmodel/misc.py
luisenp: In the current implementation we force `limit` and `threshold` to be scalars.

theseus/embodied/motionmodel/misc.py
luisenp: Not sure to be honest, but I could rewrite this in a way that this branching happens at construction time, maybe safer.

tests/embodied/motionmodel/test_misc.py
luisenp: The planner example uses vectorization, so this should be working fine. I think adding a unit test for every such cost function might be overkill (as long as we always have some examples to test it).

examples/README.md
luisenp: Fixed in 9a7cb885bfffa9bdb2fd8d41c77bf6563f61e666.

theseus/embodied/motionmodel/misc.py
luisenp: Fixed in 9a7cb885bfffa9bdb2fd8d41c77bf6563f61e666.

theseus/embodied/motionmodel/misc.py
luisenp: Fixed in 9a7cb885bfffa9bdb2fd8d41c77bf6563f61e666.

theseus/embodied/motionmodel/misc.py
mhmukadam: Ah so it's the same value for all dims. We will likely need to extend this when we start working on full body planning and joint limits get involved. But good for now to make progress.

mhmukadam (commented):
LGTM!

PR #476, opened 2023-03-01
Lie Labs API example

luisenp:
## Seeking feedback 

This PR has a small example illustrating the current features of our new Lie group tensor library for PyTorch. We are looking for feedback on: 

 1. **API**: Is it easy to use? Any confusing points? Any source of friction?
 2. **Features**: Any features you'd like to see that are not in the planned features below?
 3. **Bugs**: Any bugs you noticed?

## Current features in `alpha`

* A `LieTensor` subclass of `torch.Tensor` serving as entry point.
* SE3 and SO3 groups.
* Implementation of several Lie groups operators, each with custom backward pass: 
    * log and exp maps
    * adjoint, vee, hat operators
    * inverse, compose
    * transform_from (convert point from pose to world coordinates)
* Differentiable jacobians to use in optimization libraries (the jacobians are computed in closed form using `torch`, but w/o a custom backward pass; we rely on torch autograd for this).
* Seamless support for optimizing `LieTensor` parameters using `torch` optimizers. 
* Overloaded operators for `compose` (`*`) and `transform_from` (`@`). 

## Features planned for `beta`

* Support for arbitrary number of batch dimensions and broadcasting (current version assumes a single batch dimension for everything).
* Automatic normalization of operations result (to ensure group properties).
* SO2, SE2, and Sim group support.
* `vmap` support.
* Storage as matrix or quaternion and operators for conversions between these representations (including axis angle).
* cpp and CUDA backends if they lead to significant gains in efficiency.

### DIFF
--- a/examples/lie_labs.py
+++ b/examples/lie_labs.py
@@ -70,9 +70,9 @@ g6 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor()
 assert is_shared(g3_data, g6)
 
 # -- LieTensor.new()
-g7 = g3.new(g3_data)
-assert is_shared(g3_data, g7)  # shares storage
-assert g7.grad_fn is not None  # differentiable
+g7 = g3.new_lietensor(g3_data)
+assert not is_shared(g3_data, g7)  # doesn't share storage
+assert g7.grad_fn is None  # creates a leaf
 
 # ### Lie operations
 v = torch.randn(batch_size, 6)
--- a/examples/lie_labs.py
+++ b/examples/lie_labs.py
@@ -47,7 +47,7 @@ g3_leaf = lie.SO3(g3_data)  # creates a leaf tensor and copies data
 assert g3_leaf.grad_fn is None
 assert not is_shared(g3_leaf, g3_data)
 
-# -- lie.LieTensor() constructor is equilvalent to lie.SO3()
+# -- lie.LieTensor() constructor is equivalent to lie.SO3()
 g3_leaf_2 = lie.LieTensor(g3_data, lie.SO3)
 assert g3_leaf_2.grad_fn is None
 assert not is_shared(g3_leaf_2, g3_data)
@@ -56,7 +56,7 @@ assert not is_shared(g3_leaf_2, g3_data)
 # -- as_lietensor()
 g4 = lie.as_lietensor(g3_data, lie.SO3)
 assert is_shared(g3_data, g4)  # shares storage if possible
-assert g4.grad_fn is not None  # result is not a leaf
+assert g4.grad_fn is not None  # result is not a leaf tensor
 # Calling with a LieTensor returns the same tensor...
 g5 = lie.as_lietensor(g3, lie.SO3)
 assert g5 is g3
@@ -66,7 +66,7 @@ assert g5_double is not g3
 assert not is_shared(g5_double, g3)
 
 # -- cast()
-g6 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor
+g6 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor()
 assert is_shared(g3_data, g6)
 
 # -- LieTensor.new()
--- a/examples/lie_labs.py
+++ b/examples/lie_labs.py
@@ -10,17 +10,69 @@ g1 = lie.SE3.rand(batch_size, requires_grad=True)
 print(f"Created SE3 tensor with shape {g1.shape}")
 g2 = g1.clone()
 
-# Can create from a tensor as long as it's consistent with the desired ltype
-g3_data = lieF.so3.rand(5)
-g3 = lie.from_tensor(g3_data, lie.SO3)
+# Identity element
+i1 = lie.SO3.identity(2)
+i2 = lie.SE3.identity(2)
+print("SO3 identity", i1, i1.shape)
+print("SE3 identity", i2, i2.shape)
 
+# Indexing
+g1_slice = g1[2:4]
+assert g1_slice.shape == (2, 3, 4)
+torch.testing.assert_close(g1_slice._t, g1._t[2:4])  # type: ignore
+try:
+    bad = g1[3, 2]
+except NotImplementedError:
+    print("INDEXING ERROR: Can only slice the first dimension for now.")
+
+# ## Different constructors
+g3_data = lieF.SO3.rand(5, requires_grad=True)  # this is a regular tensor with SO3 data
+
+# Can create from a tensor as long as it's consistent with the desired ltype
+g3 = lie.from_tensor(g3_data, lie.SO3)  # keeps grad history
+assert g3.grad_fn is not None
 try:
     x = lie.from_tensor(torch.zeros(1, 3, 3), lie.SO3)
 except ValueError as e:
     print(f"ERROR: {e}")
 
+
+def is_shared(t1, t2):  # utility to check if memory is shared
+    return t1.storage().data_ptr() == t2.storage().data_ptr()
+
+
+# # Let's check different copy vs no-copy options
+# -- lie.SO3() lie.SE3()
+g3_leaf = lie.SO3(g3_data)  # creates a leaf tensor and copies data
+assert g3_leaf.grad_fn is None
+assert not is_shared(g3_leaf, g3_data)
+
+# -- lie.LieTensor() constructor is equilvalent to lie.SO3()
+g3_leaf_2 = lie.LieTensor(g3_data, lie.SO3)
+assert g3_leaf_2.grad_fn is None
+assert not is_shared(g3_leaf_2, g3_data)
+
+
+# -- as_lietensor()
 g4 = lie.as_lietensor(g3_data, lie.SO3)
-g5 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor
+assert is_shared(g3_data, g4)  # shares storage if possible
+assert g4.grad_fn is not None  # result is not a leaf
+# Calling with a LieTensor returns the same tensor...
+g5 = lie.as_lietensor(g3, lie.SO3)
+assert g5 is g3
+# ... unless dtype or device is different
+g5_double = lie.as_lietensor(g3, lie.SO3, dtype=torch.double)
+assert g5_double is not g3
+assert not is_shared(g5_double, g3)
+
+# -- cast()
+g6 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor
+assert is_shared(g3_data, g6)
+
+# -- LieTensor.new()
+g7 = g3.new(g3_data)
+assert is_shared(g3_data, g7)  # shares storage
+assert g7.grad_fn is not None  # differentiable
 
 # ### Lie operations
 v = torch.randn(batch_size, 6)
--- a/examples/lie_labs.py
+++ b/examples/lie_labs.py
@@ -1,0 +1,87 @@
+import torch
+
+import theseus.labs.lie as lie
+import theseus.labs.lie.functional as lieF
+
+batch_size = 5
+
+# ### Lie Tensor creation functions
+g1 = lie.SE3.rand(batch_size, requires_grad=True)
+print(f"Created SE3 tensor with shape {g1.shape}")
+g2 = g1.clone()
+
+# Can create from a tensor as long as it's consistent with the desired ltype
+g3_data = lieF.so3.rand(5)
+g3 = lie.from_tensor(g3_data, lie.SO3)
+
+try:
+    x = lie.from_tensor(torch.zeros(1, 3, 3), lie.SO3)
+except ValueError as e:
+    print(f"ERROR: {e}")
+
+g4 = lie.as_lietensor(g3_data, lie.SO3)
+g5 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor
+
+# ### Lie operations
+v = torch.randn(batch_size, 6)
+
+# Exponential and logarithmic map
+out1 = lie.SE3.exp(v)  # also lie.exp(v, g1.ltype)
+print(f"Exp map returns a {type(out1)}.")
+out2 = g1.log()  # also lie.log(g1)
+print(f"Log map returns a {type(out2)}.")
+
+# Inverse
+out1 = g1.inv()  # also lie.inv(g1)
+
+# Compose
+# also lie.compose(g1, g2)
+out1 = g1.compose(g2)  # type: ignore
+
+# Differentiable jacobians
+jacs, out = g1.jcompose(g2)  # type: ignore
+print("Jacobians output is a 2-tuple.")
+print("    First element is a list of jacobians, one per group argument.")
+print(f"    For compose this means length {len(jacs)}.")
+print("    The second element of the tuple is the result of the operation itself.")
+print(f"    Which for compose is a {type(out).__name__}.")
+
+# Other options:
+#   * adj(), hat(), vee(), retract(), local(),
+#   * Jacobians: jlog(), jinv(), jexp()
+
+# ### Overriden operators
+# Compose
+out2 = g1 * g2
+torch.testing.assert_close(out1, out2, check_dtype=True)
+
+# Transfrom from (from local to world coordinate frame)
+p = torch.randn(batch_size, 3)
+pt1 = g1.transform_from(p)
+pt2 = g1 @ p
+torch.testing.assert_close(pt1, pt2)
+
+# For convenience, we provide a context to drop all ltype checks, and operate
+# on raw tensor data. However, keep in mind that this is prone to error.
+# Here is one example of how this works.
+with lie.as_euclidean():
+    gg1 = torch.sin(g1)
+# The above is the same as this next call, but the context might be more convenient
+# if one is doing similar hacky stuff on several group objects.
+gg2 = torch.sin(g1._t)
+torch.testing.assert_close(gg1, gg2)
+print("Success: We just did some ops that make no sense for SE3 tensors.")
+
+# ### Lie tensors can also be used as leaf tensors for torch optimizers
+g1 = lie.SE3.rand(1, requires_grad=True)
+g2 = lie.SE3.rand(1)
+
+opt = torch.optim.Adam([g1], lr=0.1)
+
+for i in range(10):
+    opt.zero_grad()
+    d = g1.local(g2)
+    loss = torch.sum(d**2)
+    loss.backward()
+    opt.step()
+    print(f"Iter {i}. Loss: {loss.item(): .3f}")


examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+    print(f"ERROR: {e}")
+
+g4 = lie.as_lietensor(g3_data, lie.SO3)
+g5 = lie.cast(g3_data, lie.SO3)  # alias for as_lietensor
+
+# ### Lie operations
+v = torch.randn(batch_size, 6)
+
+# Exponential and logarithmic map
+out1 = g1.exp(v)  # also lie.exp(v, g1.ltype)
```
fantaosha, line 29: In differential geometry, `g.exp(v)` usually means `compose(g, exp(v))`. This is particularly the case on general Riemannian geometry where the exponential map is something like `exp_g(v)`.

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+
+# Other options:
+#   * adj(), hat(), vee(), retract(), local(),
+#   * Jacobians: jlog(), jinv(), jexp()
+
+# ### Overriden operators
+# Local map
+z = g2 - g1  # equivalent to g1.local(g2)
+print(f"Output of local is a {type(z)}.")
+g4 = -g1  # equivalent to g1.inv()
```
fantaosha, line 57: Not sure if this could be confusing for `SO(2)` where `-g` is also an element of `SO(2)`.

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+torch.testing.assert_close(g1.inv(), g4)
+
+# Some operators have strict typecheck requirements
+try:
+    g2 - z
+except TypeError as e:
+    print(f"ERROR: {e}")
+
+try:
+    g3 - g1
```
fantaosha, line 67: In pose graph optimization, the Euclidean distance is also a popular cost function. Is it still possible for us to use the Euclidean distance with this overriding?

fantaosha (commented):
Some comments about the behaviors of overriding.

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
+import torch
+
+import theseus.labs.lie as lie
+import theseus.labs.lie.functional as lieF
+
+batch_size = 5
+
+# ### Lie Tensor creation functions
+g1 = lie.rand(batch_size, lie.SE3, requires_grad=True)
```
exhaustin, line 9: Can `batch_size` be an optional argument that defaults to `1`?
I imagine there would be a lot of non-training usecases where we just want a random lie variable.

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+print(f"Created SE3 tensor with shape {g1.shape}")
+g2 = g1.clone()
+
+# Can create from a tensor as long as it's consistent with the desired ltype
+g3_data = lieF.so3.rand(5)
```
exhaustin, line 14: Is `g3_data` a tensor of shape (5, 3, 3)? I assume that data for the lie group is the transformation matrix, but just wanted to be sure.

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+print(f"Exp map returns a {type(out1)}.")
+out2 = g1.log()  # also lie.log(g1)
+print(f"Log map returns a {type(out2)}.")
+
+# Inverse
+out1 = g1.inv()  # also lie.inv(g1)
+
+# Compose
+# also lie.compose(g1, g2)
+out1 = g1.compose(g2)  # type: ignore
```
exhaustin, line 39: Overload `*` or `@` operator for `compose`?

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+print("    The second element of the tuple is the result of the operation itself.")
+print(f"    Which for compose is a {type(out).__name__}.")
+
+# Other options:
+#   * adj(), hat(), vee(), retract(), local(),
+#   * Jacobians: jlog(), jinv(), jexp()
+
+# ### Overriden operators
+# Local map
+z = g2 - g1  # equivalent to g1.local(g2)
```
exhaustin, line 55: Another clarification question: `g1.local(g2)` is equivalent to `g2.inv().compose(g1)`, correct?
Is there a way to make these operations clear in terms of the order of operands?

exhaustin (commented):
I like the clean interface, can't wait to use it!

Some features I would love to see:
- Conversions to/from quaternions & matrices, even though they might be specific to SE3 and maybe out of scope, but that's a really really common usecase in robotics.
- I find the [Sophus](https://pypi.org/project/sophuspy/) API to be more intuitive in terms of accessing the matrix data, and accessing the individual rotation and translation components. Maybe we can take some inspiration from that?

examples/lie_labs.py
luisenp: In fact, our plan for the library is to not have a required batch dimension, and support everything from no batch dimension to arbitrary number of batch dimensions. @fantaosha is working on this feature at the moment. In this case, the API for `rand` would be closer to `torch.rand` (e.g., you can do `lie.rand(lie.SE3)`, `lie.rand(4, 5, 6, lie.SE3)`, and so on.

examples/lie_labs.py
luisenp: In this case, yes. Currently we are only using matrix representation, but ideally, we'll get to the point where we can offer alternative storage options (as you are suggesting above). 

My current vision is that something like `print(g)` would return stuff like:

`LieTensor(..., shape(1, 2, 3), ltype=se3, storage="mat")` # raw tensor is shape (1, 2, 3, 3, 4)
or
`LieTensor(..., shape(1, 2, 3), ltype=se3, storage="q")`  # raw tensor is shape (1, 2, 3, 7)

with a bunch of backend function to interact across representations.

examples/lie_labs.py
exhaustin: Makes sense! However as torch has a default dtype and lie does not (because there's no sensible default), maybe the required parameter should come first to be less confusing? e.g. `lie.rand(lie.SE3, 4, 5, 6)`

Would actually be even cleaner to be able to do
```
lie.SE3.rand()
lie.SE3.rand(4, 5, 6)
```
similar to some of the Sophus initialization methods.

examples/lie_labs.py
luisenp: AFAIU, `g1.local(g2)` is `g1.inv().compose(g2)`. At least that's how we implemented in core Theseus. Any comments from @fantaosha.

examples/lie_labs.py
luisenp: > I like the clean interface, can't wait to use it!
> 
> Some features I would love to see:
> 
> * Conversions to/from quaternions & matrices, even though they might be specific to SE3 and maybe out of scope, but that's a really really common usecase in robotics.
> * I find the [Sophus](https://pypi.org/project/sophuspy/) API to be more intuitive in terms of accessing the matrix data, and accessing the individual rotation and translation components. Maybe we can take some inspiration from that?

Left some comments related to your first suggestion below. Also, @fantaosha already implemented some quaternion conversions in the "functional" package, although they are not yet exposed to the `LieTensor` class.

Regarding the Sophus suggestion. Can you post some examples of what you have in mind?

examples/lie_labs.py
luisenp: In this case you could do something like `g3._t - g3._t` (need to add some better accessors here) or 

```python
with lie.as_euclidean():
  d = g3 - g1
```

examples/lie_labs.py
luisenp: Meaning that `-g` is not equivalent to `g.inv()`?

examples/lie_labs.py
luisenp: The funny thing is that our first version had the `ltype`-first convention, and changed it after some discussion to be more similar to `torch`. But I actually like `lie.SE3.rand` suggestion, this seems cleaner to me.

examples/lie_labs.py
exhaustin: That sounds great, would be even nicer if these backend functions are available from the frontend too. (Maybe through lieF?)

examples/lie_labs.py
exhaustin: Hmm, so `g1.local(g2)` means "g2 in the local frame of g1". This is the exact type of operand order confusion I was mentioning. Or maybe it's just me that's confused.

examples/lie_labs.py
exhaustin: Pasted some examples from the Sophus docs:
```py
## Creation

# 5. R, t constructor of SE3
sp.SE3(np.eye(3), np.ones(3)) # R, t
'''
SE3([[1, 0, 0, 1],
     [0, 1, 0, 1],
     [0, 0, 1, 1],
     [0, 0, 0, 1]])
'''

R = sp.SO3([[0, 1, 0],
            [0, 0, 1],
            [1, 0, 0]])
T = sp.SE3(R.matrix(), np.ones(3))

## Interfaces

# 1. 
R.matrix()
'''
array([[0., 1., 0.],
       [0., 0., 1.],
       [1., 0., 0.]])
'''

# 5.
T.matrix()
'''
array([[0., 1., 0., 1.],
       [0., 0., 1., 1.],
       [1., 0., 0., 1.],
       [0., 0., 0., 1.]])
'''

# 6.
T.matrix3x4()
'''
array([[0., 1., 0., 1.],
       [0., 0., 1., 1.],
       [1., 0., 0., 1.]])
'''

# 7.
T.so3()
'''
SO3([[0, 1, 0],
     [0, 0, 1],
     [1, 0, 0]])
'''

# 11.
T.translation() # array([1., 1., 1.])

# 12.
T.rotationMatrix()
'''
array([[0., 1., 0.],
       [0., 0., 1.],
       [1., 0., 0.]])
'''
```

examples/lie_labs.py
exhaustin: Actually `g2 - g1` seems like a perfectly good way to remove the confusion.

examples/lie_labs.py
```diff
@@ -0,0 +1,121 @@
...
+# is a LieTensor of ltype=lie.tgt, which we require as an extra layer of safety.
+# 6 is the DOF of SE3
+x = torch.randn(batch_size, 6)  # type: ignore
+try:
+    g1 + x
+except TypeError as e:
+    print(f"ERROR: {e}")
+
+y = lie.cast(x, lie.tgt)
+g2 = g1 + y
```
rmurai0610, line 81: I find it slightly inconsistant that the operations such as SE3.exp(v) allows v to be a tensor, where as the overloaded manifold plus which is g1.compose(SE3.exp(v)) doesn't allow v to be a tensor and you need the cast it to lie.tgt

To be consistent we should either allow it or not allow it for both? It's probably going to be weird writing code where in some part this cast is required and in others it's not

examples/lie_labs.py
luisenp: There has been some discussion about whether to make `lieF` available or not. I don't remember what the cons we discussed were, maybe we had some concerns that people would be confused about what the canonical entry point for the library would be? @mhmukadam do you remember?

examples/lie_labs.py
luisenp: Thanks!

examples/lie_labs.py
luisenp: I think @fantaosha had also mentioned some people use `+` for compose? I don't really have a strong preference here, except we should use what our typical user would be more accustomed to and/or find more natural.

exhaustin: Here's what I, a single datapoint for a typical user, are accustomed to using:
`g1 * g2` for composing `g1` and `g2`
Explicitly doing `g1.inv() * g2` for locals, to remove any confusion as to what exactly is happening.

examples/lie_labs.py
luisenp: One concern we have had since the beginning is people trying to do confusing stuff with overloaded operators and have an operation going through with a calculation that makes absolutely no sense. For example, some naive user wanting to do standard addition on the data matrix, and just passing `g + x` without realizing this is actually a retraction map. That's the kind of silent error we are trying to avoid with this. It's different with `SE3.retract()` because there is no ambiguity about what they are trying to do. 

Do you think this concern is overstated? Any thoughts from @exhaustin, @dongjing3309, @ddetone?

examples/lie_labs.py
luisenp: I also like this suggestion. One thing that was bothering is that this overrides `torch.Tensor.exp()`, which may have been confusing for some users, which this would also take care of.

examples/lie_labs.py
exhaustin: Disclaimer: This is my own personal preference and might not apply to some or the majority of users.
I would be fine without operator overloading, as it makes the code look more transparent and easier to debug.
The one exception is `compose` as it is extremely common and relatively straightforward. Overloading the `@` operator with that functionality is nice as 1) it is already a common operator for matrix multiplications, and performing matrix operations on the data matrix is AFAIU equivalent to compose? 2) It won't be mistaken as an elementwise multiplication. (Personally I'm not against `*` either)

examples/lie_labs.py
mhmukadam: I think it was mostly that now we have worked out subclassing torch tensor cleanly so exposing two types of API seemed bloat.

If these backend functions are also exposed in the frontend through the class etc then do you think there is still utility in having access to the functional API through lieF directly?

examples/lie_labs.py
luisenp: Thanks for sharing! This is valuable input. Would love to hear from others, but FWIW, I think this aligns well with @fantaosha's opinion (he is concerned about the transparency/ambiguity of these operators).

examples/lie_labs.py
fantaosha: Ah, sorry. Just saw it. Yes, `-g` is still an element of `SO(2)` but it is not `g.inv()`. For example, `g=[c, s]`, `-g=[-c, -s]`, `g.inv()=[c, -s]`.

examples/lie_labs.py
luisenp: Added in 530172f5005a92b0eb52446092cd3f8620c673d9.

examples/lie_labs.py
luisenp: Removed in 530172f5005a92b0eb52446092cd3f8620c673d9.

examples/lie_labs.py
luisenp: Added in 530172f5005a92b0eb52446092cd3f8620c673d9.

examples/lie_labs.py
luisenp: Added `*` for compose in 530172f5005a92b0eb52446092cd3f8620c673d9. Still need to add other methods to access the data.

examples/lie_labs.py
luisenp: Removed `__sub__` in 530172f5005a92b0eb52446092cd3f8620c673d9. We decided to only overload `*` for now, as seems to be the case in Sophus.

examples/lie_labs.py
luisenp: Removed `+` in 530172f5005a92b0eb52446092cd3f8620c673d9.

luisenp:
Made a new version of the API based on the discussion on operator overload, and also the suggestions for things like `SE3.rand` and `SE3.exp`. Feel free to ignore the implementation details, and just look at the example script. If this looks good, I'll probably move everything to a separate PR merge, and keep this PR with only the script.

exhaustin (approved):
Again, just my subjective opinion, but the new version looks great to me!

rmurai0610:
The api looks great! 
I've tried using the api, and I've noticed few minor details:

```
import theseus.labs.lie as lie
import theseus as th

g_old = th.SE3.rand(5)
print(g_old[0])

g_new = lie.SE3.rand(5)
print(g_new[0])
```
```
tensor([[-0.5758,  0.2524, -0.7777,  0.5286],
        [ 0.1701, -0.8934, -0.4159,  0.8484],
        [-0.7997, -0.3718,  0.4715,  0.3887]])
Traceback (most recent call last):
  File "/Users/riku/phd/manim/test.py", line 8, in <module>
    print(g_new[0])
  File "/Users/riku/phd/manim/theseus/theseus/labs/lie/lie_tensor.py", line 202, in __torch_function__
    return cls._torch_function_impl_lie(func, types, args, kwargs)
  File "/Users/riku/phd/manim/theseus/theseus/labs/lie/lie_tensor.py", line 189, in _torch_function_impl_lie
    raise NotImplementedError(
NotImplementedError: Tried to call a torch function not supported by LieTensor. If trying to operate on the raw tensor data, please use group._t, or run inside the context lie.as_euclidean().
```
I think it would be useful if indexing was implemented.
The old implementation in theseus returns tensor, but is there a reason why we don't return a LieGroup?


-----------------------
```
import theseus.labs.lie as lie
import theseus as th

g1 = lie.SE3.rand(5)
print(type(g1.shape))
print(g1.shape)
print(type(g1._t.shape))
print(g1._t.shape)

g2 = th.SE3.rand(5)
print(type(g2.shape))
print(g2.shape)

```
```
<class 'tuple'>
(5, 3, 4)
<class 'torch.Size'>
torch.Size([5, 3, 4])
<class 'torch.Size'>
torch.Size([5, 3, 4])
```
Maybe it would be nice to keep the type consistent for typehint etc.

examples/lie_labs.py
```diff
@@ -0,0 +1,87 @@
...
+
+batch_size = 5
+
+# ### Lie Tensor creation functions
+g1 = lie.SE3.rand(batch_size, requires_grad=True)
+print(f"Created SE3 tensor with shape {g1.shape}")
+g2 = g1.clone()
+
+# Can create from a tensor as long as it's consistent with the desired ltype
+g3_data = lieF.so3.rand(5)
```
akashsharma02, line 14: I just wanted to make sure that I understand this API correctly: 

typically `so3` implies the lie algebra elements which lie in the tangent space, and `SO3` implies LieGroup elements on the manifold, and therefore I was expecting this API to simply return random 3-vectors. 

However, this API returns 3x3 matrices of tensor type. If this is expected `lieF.so3.rand` should simply be a function like `lieF.SO3rand` or something similar to follow functional style as well.

luisenp: Hi @akashsharma02. Thanks for pointing this out, I can see why this is confusing. We were aware of this convention, but, tbh, I forgot we have `so3` name under `LieF`; I've been mostly focusing on the `LieTensor` class lately. 

Your suggestion makes sense, although it would also need to be applied to all functions under `lie.so3` (see [here](https://github.com/facebookresearch/theseus/blob/main/theseus/labs/lie/functional/lie_group.py#L173) and [here](https://github.com/facebookresearch/theseus/blob/main/theseus/labs/lie/functional/so3_impl.py#L920)). Any thoughts on something like `lieF.SO3.op_name()` instead? 

cc @fantaosha @mhmukadam

exhaustin: > Any thoughts on something like `lieF.SO3.op_name()` instead?
> 

I like `lieF.SO3.op_name()` as it makes it consistent with `lie.SO3`.
Sophus also uses functional form operations such as `sp.SO3.exp()`, so this API wouldn't be new to a lot of people.

mhmukadam: Thanks @akashsharma02. Agreed with @luisenp we can switch to `lieF.SO3.op_name()`.

I am wondering do you see the utility in keeping the functional api exposed via `lieF` or do you find it redundant?

examples/lie_labs.py
```diff
@@ -0,0 +1,87 @@
+import torch
+
+import theseus.labs.lie as lie
```
mhmukadam, line 3: @rmurai0610 thanks for the feedback!

Agreed we can include support for indexing. I wasn't clear on the second point about type consistency with the older Lie groups. Any thoughts @luisenp @fantaosha?

examples/lie_labs.py
akashsharma02: @luisenp  I think the `lieF.SO3.op_name()` would work well. 

@mhmukadam I do think there is utility in exposing both APIs, just from the standpoint of allowing users to choose either functional style or OOP style. In either case, some functions like `rand` *will* need to be exposed from the functional side.

examples/lie_labs.py
```diff
@@ -0,0 +1,87 @@
...
+#   * Jacobians: jlog(), jinv(), jexp()
+
+# ### Overriden operators
+# Compose
+out2 = g1 * g2
+torch.testing.assert_close(out1, out2, check_dtype=True)
+
+# Transfrom from (from local to world coordinate frame)
+p = torch.randn(batch_size, 3)
+pt1 = g1.transform_from(p)
```
akashsharma02, line 60: Would it make more sense to call this API `act` instead of `transform_from` to imply the "Action" of the group element on a vector element?

mhmukadam: Hmm there is [left_act and left_project](https://github.com/facebookresearch/theseus/blob/lep.labs_api_example/theseus/labs/lie/lie_tensor.py#L479). We went with `transform_from` since it is commonly used like in GTSAM but we can always alias it with `act` if this is standard anywhere else (like Sophus)?

examples/lie_labs.py
mhmukadam: Ah okay, makes sense to have consistent torch.Size, we can update that. I don't think it was intentional, @luisenp?

luisenp: Thanks for the feedback, @rmurai0610. This was not intentional, neither was the indexing. Will definitely fix this in the next batch of changes!

rmurai0610: Also, is there an easy way to create an identity group element with the new api?
Currently, to create an identity element of SO3, I can do 
`lie.cast(torch.eye(3)[None], lie.SO3)`
but it's quite verbose. Maybe we can have lie.SO3.identity() etc? (unless there is already a way to do this)

Also, this is personal preference, but in the older api, it was possible to write something like
```
g1 = th.SO3(tensor=...)
```
to make a new lie group element. 
With the new API, it's
```
g2 = lie.cast(..., lie.SO3) # or lie.from_tensor()
```
I feel these are not as intuitive, and isn't consitent with how other libraries like Sophus constructs a new lie group element.

mhmukadam: Yes we can identity function.

The cast/from_tensor are more torch like functions since this is meant to be a torch first library. We could also add sophus like functions and try to balance not making the api too messy. Any thoughts?

examples/lie_labs.py
fantaosha: I agree that `SO3` and `SE3` are better. BTW, I wonder if we need to change `adjoint->Adjoint` since `Adjoint` and `adjoint` have different meaning for Lie group operators.

examples/lie_labs.py
fantaosha: Just a clarification, we keep `act` for the action of Lie group elements on Lie algebra such as twists and angular velocities, where `transform`  for the action of Lie group elements on positions.

examples/lie_labs.py
rmurai0610: Thank you, identity function would be quite useful (at least for some of my applications).

For the from_tensor, in pytorch, we can do
`x = torch.tensor([[1.0, 0.0, 0.0]])`
To me it feels more intuitive to follow torch's oop style, rather than having a functional style?

Also, the behavior of `torch.tensor` and `torch.from_numpy` is different (first one copies memory and latter does not)
```
import theseus.labs.lie as lie
import torch
import numpy as np

g1 = lie.SE3.rand(5)
g2 = lie.from_tensor(g1._t, lie.SE3)


print("g1 data_ptr", g1._t.data_ptr())
print("g2 data_ptr", g2._t.data_ptr())

t = np.array([0, 0, 0])
print("torch tensor 1", torch.tensor(t).data_ptr())
print("torch tensor 2", torch.tensor(t).data_ptr())
print("torch from_numpy 1", torch.from_numpy(t).data_ptr())
print("torch from_numpy 2", torch.from_numpy(t).data_ptr())
```

```
g1 data_ptr 5189291520
g2 data_ptr 5189291520 # points to same data as g2
torch tensor 1 5199293184
torch tensor 2 5199127232 # torch.tensor copied data elsewhere
torch from_numpy 1 105553175841472 
torch from_numpy 2 105553175841472 # torch.from_numpy points to the same data
```

Maybe the same convention should be followed for the library? (the lie.SE3 copies, whereas lie.cast/from_tensor does not)

mhmukadam: We are exposing both oop and functional style api. In that spirit, we could move such functional like methods to the `lieF` namespace and add missing methods on the oop `lie` namespace side.

rmurai0610: It’s my personal preference but I think something like
`lie.SE3(tensor=…)` and `lie.SE3.from_tensor(…)` in the oop version and `lieF.from_tensor` in the functional would be nice.
Being able to choose between copy and no-copy initialisation I think would be handy.

examples/lie_labs.py
mhmukadam: Confirming that we leave this one as is for now and add `act` later when we include Lie algebra? @fantaosha

examples/lie_labs.py
luisenp: I just realized that I didn't include in the example script something like `lie.LieTensor(t, lie.SO3)`, which was also allowed. The difference with `lie.from_tensor()` is that directly calling `LieTensor` returns a leaf tensor, but `from_tensor()` keeps the compute graph's history (see the image). 

image

I like the suggestion of having `lie.SO3(tensor)` as well. I'll add this. 

Anything else I should consider?

luisenp: Regarding the copy vs. no-copy, at the moment it seems we don't have a way to create a lie tensor without making a copy. My understanding was that the implementation of `LieTensor` was such that the same underlying storage was used, but I'm running some tests and it doesn't seem to be the case (but I might be doing something wrong). I need to look more into this.

mhmukadam (approved):
LGTM with all the feedback now addressed in the other PR.

Thanks @luisenp and @fantaosha for developing everything on Lie groups alpha and thanks to everyone else (on the PR and offline) for the valuable feedback!

PR #477, opened 2023-03-07
Change in data model in test_theseus_layer

dishank-b:
## Motivation and Context

The data model was wrong as in it had to two optima, leading to some of the optimizer not able to pass the test. The comments in the file also suggests that the intentional model and actually implemented model were different.
## How Has This Been Tested

All the current optimizer passes the test with the changed model. 

## Types of changes

## Checklist

### DIFF
--- a/tests/test_theseus_layer.py
+++ b/tests/test_theseus_layer.py
@@ -26,12 +26,12 @@ device = "cuda:0" if torch.cuda.is_available() else "cpu"
 
 
 def model(x, b):
-    return (b[..., :1] * x) ** 2 + b[..., 1:]
+    return b[..., :1] * x**2 + b[..., 1:]
 
 
 def model_grad(x, b):
-    g1 = 2 * (b[..., :1] * x) * x
-    g2 = b[..., 1:] * torch.ones_like(x)
+    g1 = x**2
+    g2 = torch.ones_like(x)
     return g1, g2
 
 


dishank-b:
@mhmukadam @luisenp, made changes in the `test_theseus_layer`. They were causing bug for DCEM convergence. 

Also, in the  comments in the file, it say the model we want to have is `f(x) = Ax^2+b` whereas the one implemented was `f(x) = (Ax)^2+b`. I corrected it now. Though verify the change in the grad once. I feel the grad in previous version also wrong for the model we had.

tests/test_theseus_layer.py
```diff
@@ -26,12 +26,12 @@
...
 def model(x, b):
-    return (b[..., :1] * x) ** 2 + b[..., 1:]
+    return b[..., :1] * x**2 + b[..., 1:]
 
 
 def model_grad(x, b):
-    g1 = 2 * (b[..., :1] * x) * x
-    g2 = b[..., 1:] * torch.ones_like(x)
+    g1 = x**2
+    g2 = torch.ones_like(x)
```
mhmukadam, line 34: Should the size here be related to `b[..., 1:]`?

mhmukadam (approved):
LGTM, thanks for the fix. Just one comment.

tests/test_theseus_layer.py
luisenp: Thanks for catching this, not sure what I was thinking when I added that `b` factor. The size is correct, because of the following:

- `b` (optim_var) is shape `(batch_size, 2)`
- `x` (data, constant) is shape `(batch_size, num_points)`
- error is shape `(batch_size, num_points)`

We want the jacobian with respect to `b` be of shape `(batch_size, num_points, 2)`, and it's computed below as the stack of `[g1, g2]` on `axis=2`, so we need for `g2` to be of shape `(batch_size, num_points)`, which is the shape of `x`.

luisenp (approved):
Thanks for the fix!

dishank-b:
Thanks. Can we merge this then?  Then I will rebase the DCEM PR on top of this.

PR #478, opened 2023-03-09
Fixed broken compute graph in LieTensor.new()

luisenp:


tests/labs/lie/test_lie_tensor.py
```diff
@@ -118,3 +118,20 @@ def _to_torch(t):
         # Check class-version
         jac_c, out_c = _to_functional_fmt(getattr(c, f"j{op_name}")(*c_inputs))
         torch.testing.assert_close([jac1, out], [jac_c, out_c])
+
+
+@run_if_labs()
+def test_backward_works():
```
mhmukadam, line 124: Should we loop over all Lie group types beyond SE3?

mhmukadam (commented):
LGTM!

tests/labs/lie/test_lie_tensor.py
luisenp: Since, I'll make a new PR to clean the one with the Lie group example I can just throw a look with SO3 in there, it's pretty easy.

PR #494, opened 2023-04-05
Lie Labs API v3

luisenp:
Addresses comments from #476.

### DIFF
--- a/theseus/labs/lie/functional/lie_group.py
+++ b/theseus/labs/lie/functional/lie_group.py
@@ -169,6 +169,16 @@ class _RandFnType(Protocol):
         pass
 
 
+class _IdentityFnType(Protocol):
+    def __call__(
+        *size: int,
+        dtype: Optional[torch.dtype] = None,
+        device: DeviceType = None,
+        requires_grad: bool = False,
+    ) -> torch.Tensor:
+        pass
+
+
 # Namespace to facilitate type-checking downstream
 class LieGroupFns:
     def __init__(self, module):
@@ -199,3 +209,4 @@ class LieGroupFns:
         self.check_left_project_matrix: _CheckFnType = module.check_left_project_matrix
         self.rand: _RandFnType = module.rand
         self.randn: _RandFnType = module.randn
+        self.identity: _IdentityFnType = module.identity
--- a/theseus/labs/lie/functional/se3_impl.py
+++ b/theseus/labs/lie/functional/se3_impl.py
@@ -81,19 +81,9 @@ def rand(
 ) -> torch.Tensor:
     if len(size) != 1:
         raise ValueError("The size should be 1D.")
-    rotation = so3.rand(
-        size[0],
-        generator=generator,
-        dtype=dtype,
-        device=device,
-    )
+    rotation = so3.rand(size[0], generator=generator, dtype=dtype, device=device)
     translation = torch.rand(
-        size[0],
-        3,
-        1,
-        generator=generator,
-        dtype=dtype,
-        device=device,
+        size[0], 3, 1, generator=generator, dtype=dtype, device=device
     )
     ret = torch.cat((rotation, translation), dim=2)
     ret.requires_grad_(requires_grad)
@@ -112,21 +102,31 @@ def randn(
 ) -> torch.Tensor:
     if len(size) != 1:
         raise ValueError("The size should be 1D.")
-    rotation = so3.randn(
-        size[0],
-        generator=generator,
-        dtype=dtype,
-        device=device,
-    )
+    rotation = so3.randn(size[0], generator=generator, dtype=dtype, device=device)
     translation = torch.randn(
-        size[0],
-        3,
-        1,
-        generator=generator,
-        dtype=dtype,
-        device=device,
+        size[0], 3, 1, generator=generator, dtype=dtype, device=device
     )
     ret = torch.cat((rotation, translation), dim=2)
+    ret.requires_grad_(requires_grad)
+    return ret
+
+
+# -----------------------------------------------------------------------------
+# Identity
+# -----------------------------------------------------------------------------
+def identity(
+    *size: int,
+    dtype: Optional[torch.dtype] = None,
+    device: constants.DeviceType = None,
+    requires_grad: bool = False,
+) -> torch.Tensor:
+    if len(size) != 1:
+        raise ValueError("The size should be 1D.")
+    ret = torch.tensor(
+        [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]],
+        dtype=dtype,
+        device=device,
+    ).repeat(size[0], 1, 1)
     ret.requires_grad_(requires_grad)
     return ret
 
--- a/theseus/labs/lie/functional/se3_impl.py
+++ b/theseus/labs/lie/functional/se3_impl.py
@@ -114,6 +114,9 @@ def randn(
 # -----------------------------------------------------------------------------
 # Identity
 # -----------------------------------------------------------------------------
+_BASE_IDENTITY_SE3 = [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]]
+
+
 def identity(
     *size: int,
     dtype: Optional[torch.dtype] = None,
@@ -122,11 +125,9 @@ def identity(
 ) -> torch.Tensor:
     if len(size) != 1:
         raise ValueError("The size should be 1D.")
-    ret = torch.tensor(
-        [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]],
-        dtype=dtype,
-        device=device,
-    ).repeat(size[0], 1, 1)
+    ret = torch.tensor(_BASE_IDENTITY_SE3, dtype=dtype, device=device).repeat(
+        size[0], 1, 1
+    )
     ret.requires_grad_(requires_grad)
     return ret
 
--- a/theseus/labs/lie/functional/se3_impl.py
+++ b/theseus/labs/lie/functional/se3_impl.py
@@ -7,7 +7,7 @@ import torch
 from typing import cast, List, Tuple, Optional
 
 from . import constants
-from . import lie_group, so3_impl as so3
+from . import lie_group, so3_impl as SO3
 from .utils import get_module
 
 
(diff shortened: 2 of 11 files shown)


theseus/labs/lie/functional/se3_impl.py
```diff
@@ -81,19 +81,9 @@ def rand(
 ) -> torch.Tensor:
     if len(size) != 1:
         raise ValueError("The size should be 1D.")
-    rotation = so3.rand(
-        size[0],
-        generator=generator,
-        dtype=dtype,
-        device=device,
-    )
+    rotation = so3.rand(size[0], generator=generator, dtype=dtype, device=device)
```
mhmukadam, line 84: For consistency with the new user facing `lieF.SE3/SO3` should the imports internally also update to SE3/SO3 (line 10 in this file)?

theseus/labs/lie/functional/se3_impl.py
```diff
@@ -112,25 +102,35 @@ def randn(
...
+def identity(
+    *size: int,
+    dtype: Optional[torch.dtype] = None,
+    device: constants.DeviceType = None,
+    requires_grad: bool = False,
+) -> torch.Tensor:
+    if len(size) != 1:
+        raise ValueError("The size should be 1D.")
+    ret = torch.tensor(
+        [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]],
```
mhmukadam, line 126: Call into SO3.identity or keep it independent?

theseus/labs/lie/lie_tensor.py
```diff
@@ -55,15 +56,30 @@ def euclidean_enabled() -> bool:
 
 
 class _LieTensorBase(torch.Tensor):
-    def __new__(cls, data: torch.Tensor, ltype: _ltype, requires_grad=False):
-        return cls._make_subclass(cls, data, requires_grad)  # type: ignore
+    def __new__(
```
mhmukadam, line 59: Do any unit tests need updating?

theseus/labs/lie/lie_tensor.py
```diff
@@ -55,15 +56,30 @@ def euclidean_enabled() -> bool:
...
+        cls,
+        data: torch.Tensor,
+        ltype: _ltype,
+        requires_grad: bool = False,
+        _shared_memory: bool = False,
```
mhmukadam, line 64: Do any Lie group unit tests need updating? In particular to test new features like the shared memory?

theseus/labs/lie/lie_tensor.py
```diff
@@ -143,7 +159,8 @@ class LieTensor(_LieTensorBase):
         torch.Tensor.layout.__get__,  # type: ignore
         torch.Tensor.requires_grad.__get__,  # type: ignore
         torch.Tensor.retains_grad.__get__,  # type: ignore
-        torch.Tensor.shape.__get__,  # type: ignore
+        torch.Tensor.storage,  # type: ignore
```
mhmukadam, line 162: Couldn't figure out where this is used.

theseus/labs/lie/lie_tensor.py
```diff
@@ -388,28 +427,39 @@ def as_lietensor(
     data: Any,
     ltype: Optional[_ltype] = None,
     dtype: torch.dtype = torch.float,
-    device: Union[torch.device, str] = None,
+    device: DeviceType = None,
 ) -> _LieTensorBase:
     if isinstance(data, LieTensor):
-        return data
+        data.to()
```
mhmukadam, line 433: If passed data is on GPU does calling default `.to()` move it to CPU and then back to GPU below in line 436?

mhmukadam (commented):
Looks great! Thanks for adding all the updates from feedback on alpha. Left some minor comments.

theseus/labs/lie/lie_tensor.py
```diff
@@ -55,15 +56,30 @@ def euclidean_enabled() -> bool:
...
     def _t(self) -> torch.Tensor:
         return self.as_subclass(torch.Tensor)
 
-    def __init__(self, data: Any, ltype: _ltype, requires_grad=None):
+    def __init__(
+        self,
+        data: Any,
+        ltype: _ltype,
+        requires_grad: Optional[bool] = None,
+        _shared_memory: bool = False,
```
fantaosha, line 79: So `_shared_memory` has not been used in the constructor? Also, I guess `_shared_memory` tries to use the initialized tensor?

theseus/labs/lie/functional/se3_impl.py
```diff
@@ -112,25 +102,35 @@ def randn(
...
+        dtype=dtype,
+        device=device,
+    ).repeat(size[0], 1, 1)
```
fantaosha, line 129: Looks like `expand()` will be more memory efficient. However, not sure if there are any potential issues.

fantaosha (commented):
Left some comments. LGTM overall.

theseus/labs/lie/lie_tensor.py
mhmukadam, line 59: Flagging @rmurai0610's comment here so we can maybe address it in this PR.

> Thank you Luis, it looks great!
> 
> I have a minor comment.
> For `LieTensor.new()`, should it share storage? 
> For example, `tensor.new_tensor` always copies the data, and I think the term new sort of implies that, and I think it would be nicer if we kept consistent.
> 
> Also, I'm not sure how exactly I should read this line.
> `x.new(data)` isn't very intuitive, maybe something like `x.new_lietensor` is more clear (similar to new_tensor)

theseus/labs/lie/functional/se3_impl.py
mhmukadam: I don't think we have `SO3.eye`, we can alias although I have not seen this commonly used.

theseus/labs/lie/functional/se3_impl.py
luisenp: Sounds good, will fix.

theseus/labs/lie/functional/se3_impl.py
luisenp: This was actually proposed by @fantaosha and seems like a more straightforward solution to me (it requires one less function call).

I actually thought about naming it `SO3.eye` as well, but I wasn't sure. I do like it though, it's very concise.

theseus/labs/lie/lie_tensor.py
luisenp: When any of [these](https://github.com/facebookresearch/theseus/pull/494/files/a4e93e7d03f065d06f2849c4a678ae6190b49ad0#diff-640b323e767b7377419a10b17fa0b39e8bf056486fa0921a44f986eca932a8bfR151) functions is called by `torch`, our class simply runs the default method provided by `torch.Tensor` class. See [here](https://github.com/facebookresearch/theseus/pull/494/files/a4e93e7d03f065d06f2849c4a678ae6190b49ad0#diff-640b323e767b7377419a10b17fa0b39e8bf056486fa0921a44f986eca932a8bfR212).

theseus/labs/lie/lie_tensor.py
```diff
@@ -185,12 +215,21 @@ def _torch_function_impl_lie(cls, func, types, args=(), kwargs=None):
...
+                assert ret.ndim in [2, 3]
+                if ret.ndim == 2:
+                    ret = ret._t.unsqueeze(0)
+            return tree_map_only(torch.Tensor, lambda x: from_tensor(x, ltype), ret)
+        return NotImplemented
+        # raise NotImplementedError(
+        #     "Tried to call a torch function not supported by LieTensor. "
+        #     "If trying to operate on the raw tensor data, please use group._t, "
+        #     "or run inside the context lie.as_euclidean()."
+        # )
```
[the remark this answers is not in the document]
luisenp: Actually, this was only a temporary change while debugging. Thanks for catching!

theseus/labs/lie/lie_tensor.py
luisenp: Oops, this shouldn't be here, I wrote it so I could click into the function code with the IDE lol. Sorry about that.

theseus/labs/lie/lie_tensor.py
luisenp: `_shared_memory` is used in `__new__`, which is called by Python before `__init__`, and it's used to allocate the memory for the new object.

theseus/labs/lie/lie_tensor.py
luisenp: Fixed. cc @rmurai0610

theseus/labs/lie/lie_tensor.py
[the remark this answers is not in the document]
luisenp: Fixed.

theseus/labs/lie/functional/se3_impl.py
mhmukadam: Based on offline discussion we will only keep identity for now and add eye if user feedback surfaces the need.

theseus/labs/lie/lie_tensor.py
mhmukadam: Sounds good.

theseus/labs/lie/lie_tensor.py
mhmukadam: No worries, thanks for updating.

theseus/labs/lie/functional/lie_group.py
```diff
@@ -169,6 +169,16 @@ def __call__(
         pass
 
 
+class _IdentityFnType(Protocol):
```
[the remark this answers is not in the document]
mhmukadam: Offline discussion: won't add eye for now since it is confusing in the context of groups (especially when storage is quaternion as pointed out by Riku).

theseus/labs/lie/lie_tensor.py
```diff
@@ -245,14 +245,13 @@ def _check_ltype(self, other: "LieTensor", op_name: str):
...
-    def new(  # type: ignore
+    def new_lietensor(  # type: ignore
         self, args: Any, device: Device = None, dtype: Optional[torch.dtype] = None
     ) -> "LieTensor":
         if isinstance(args, LieTensor):
             warnings.warn(
-                "Calling new() on a LieTensor results in shared data storage. "
                 "To copy construct from a LieTensor, it is recommended to use "
-                "lie_tensor.clone().",
+                "t.clone().detach() instead of lie_tensor.new_lietensor(t)",
```
mhmukadam, line 253: Assuming `t` is a `LieTensor`?

luisenp: Oops, that should be `"...lie_tensor.clone.detach()..."`. Good catch.

PR #498, opened 2023-04-11
Made the threshold for converting occupancy -> SDF configurable.

luisenp:


theseus/embodied/collision/signed_distance_field.py
```diff
@@ -22,14 +22,17 @@ def __init__(
         cell_size: Union[float, torch.Tensor, Variable],
         sdf_data: Optional[Union[torch.Tensor, Variable]] = None,
         occupancy_map: Optional[Union[torch.Tensor, Variable]] = None,
+        occupancy_threshold: float = 0.75,
```
mhmukadam, line 25: Any need to make this consistent with others as `Union[float, torch.Tensor, Variable]` or overkill?

luisenp: This is used in a non-differentiable part of the code, so I think anything other than a `float` would be overkill.

PR #502, opened 2023-04-27
Update robust loss

xphsean12:
## Motivation and Context
Current Robust Loss cannot deal with outlinear existing in a optimization problem

## Types of changes

## Checklist


theseus/core/robust_cost_function.py
```diff
@@ -80,20 +82,33 @@ def error(self) -> torch.Tensor:
 
     def weighted_error(self) -> torch.Tensor:
         weighted_error = self.cost_function.weighted_error()
-        squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
-        error_loss = self.loss.evaluate(squared_norm, self.log_loss_radius.tensor)
-
+        
+        if self.RobustSum:
```
luisenp, line 86: I think this can be done much more simply. If I'm not mistaken, you can keep all of the old code and just add the following lines at the beginning of `weighted_error()`
```python
if self.flatten_dims:
    weighted_error = weighted_error.view(-1, 1)
```
and then reshape the output back to `(-1, self.dim())` before returning. 

A similar procedure can be followed for `weighted_jacobians`.

theseus/core/robust_cost_function.py
```diff
@@ -80,20 +82,33 @@ def error(self) -> torch.Tensor:
...
+            torch.ones_like(weighted_error)
+            * (error_loss / self.dim() + RobustCostFunction._EPS).sqrt()
+            )
+        else:
+            squared_norm = weighted_error**2
+            error_loss = self.loss.evaluate(squared_norm, self.log_loss_radius.tensor)
+            return (
+              (error_loss  + RobustCostFunction._EPS).sqrt()
+            )
+        # Inside will compare squared_norm with exp(radius)
```
luisenp, line 99: Please remove any debug comments.

theseus/core/robust_loss.py
```diff
@@ -11,13 +11,14 @@
 
 
 class RobustLoss(abc.ABC):
+    # .suqre() make it easier to compare with residual.square(), which is x
     @classmethod
     def evaluate(cls, x: torch.Tensor, log_radius: torch.Tensor) -> torch.Tensor:
-        return cls._evaluate_impl(x, log_radius.exp())
-
+        return cls._evaluate_impl(x, log_radius.square())
```
luisenp, line 17: We should keep this as `exp()`.

theseus/core/robust_loss.py
```diff
@@ -49,4 +50,4 @@ def _evaluate_impl(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
 
     @staticmethod
     def _linearize_impl(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
-        return torch.sqrt(radius / torch.max(x, radius) + _LOSS_EPS)
+        return torch.sqrt(radius / torch.max(x, radius) + _LOSS_EPS)
```
luisenp, line 53: Missing line at end of file.

tutorials/test_RobustLoss.ipynb
```diff
@@ -0,0 +1,731 @@
+{
```
luisenp, line 1: For the unit test, the right place to do it is [this](https://github.com/facebookresearch/theseus/blob/main/tests/core/test_robust_cost.py) file. There is no need to write a full example. We only need two versions of a small objective and compare that the both versions lead to the same error vectors. We can remove this notebook.

theseus/core/robust_cost_function.py
```diff
@@ -107,15 +122,43 @@ def weighted_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
...
-        squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
-        rescale = (
-            self.loss.linearize(squared_norm, self.log_loss_radius.tensor)
-            + RobustCostFunction._EPS
-        ).sqrt()
-
-        return [
-            rescale.view(-1, 1, 1) * jacobian for jacobian in weighted_jacobians
-        ], rescale * weighted_error
+        #squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
```
luisenp, line 125: See comment above on a suggestion for implementing this more easily.

luisenp (changes requested):
Thanks a lot for your contribution! I left some comments that need to be addressed. 

Also CI is failing due to some lint errors. Can you please install `pip install -r requirements/dev.txt` and also [pre-commit hooks](https://pre-commit.com/#usage), so that they take care of lint errors automatically. 

Thanks!

theseus/core/robust_cost_function.py
xphsean12: I am not sure what you mean. And why I need to reshape it. Because what I want is given any shape of error, like (Batch, Dim) and the corresponding radius with the shape of (Batch, 1), they will all be passed to the evaluation function self.loss.evaluate and return the scaled loss. 

image

theseus/core/robust_loss.py
xphsean12: Sure. Could I know the initiative behind it?

tutorials/test_RobustLoss.ipynb
xphsean12: I use notebook to test to avoid the old version will mess up with the new one. Because I have installed Theseus 0.1.4 and do not want to make any changes to it. And any of the function that call th.xx will call the original Theseus function. Do you have any good ideas? Or I can just replace my new script file with the one in the environment site-package?

theseus/core/robust_cost_function.py
luisenp: It's a bit hard to explain, but it works, I just tested it. I'll put another PR soon.

luisenp: Please take a look at #503. The code for the jacobians part was a bit tricker than I suggested above, but I was still able to mostly use all of the old computation. Also look at the unit test I added and let me know if covers the expected behavior. Thanks!

theseus/core/robust_loss.py
fantaosha: Since square() is not monotonous, I personally prefer 'exp()', which is more useful when we try to learn a robust loss radius.

fantaosha (commented):
Add some more comments in #503.

theseus/core/robust_loss.py
xphsean12: > 

Oh! That's great! Because at a very early time, I tried to let the system automatically learn the Radius, and I failed at that time (performance was not good). But with the new version, the training outcome may change.

theseus/core/robust_cost_function.py
xphsean12: > It's a bit hard to explain, but it works, I just tested it. I'll put another PR soon.

Oh! I see what you mean! You treat every element of the original cost as the original "square norm"

PR #503, opened 2023-04-28
Added an option to flatten_dims in RobustCostFunction.

### LINKED ISSUE #497: Huber Loss Not Correct
## 🐛 Bug

I add the huber loss to the current loss function. However, I found the value comes from the huber loss may with some problem 
## Steps to Reproduce

First I will check the cost without Huber, the result is as below:
image
And then I add the huber with the a sufficiently large radius (100). According to the logic of Huber loss, the cost should maintain the same, or at least keeps a similar trend as the original cost.
image
However, the result is as below.
image
Besides, below is the cost with radius equal to 1.
image

I add the Huber loss similar to the below example
image

Thanks.

## System Info

- OS (e.g., Linux):
- Python version:
- GPU models and configuration:
- CUDA version:
- pip/conda dependencies packages versions:
- Any other relevant information:

luisenp:
Closes #497

### DIFF
--- a/tests/core/test_robust_cost.py
+++ b/tests/core/test_robust_cost.py
@@ -136,3 +136,53 @@ def test_mask_jacobians(loss_cls):
     torch.testing.assert_close(err, err_expected)
     for j1, j2 in zip(jac, jac_expected):
         torch.testing.assert_close(j1, j2)
+
+
+@pytest.mark.parametrize("batch_size", [1, 4])
+def test_flatten_dims(batch_size):
+    n = 10
+    w = th.ScaleCostWeight(0.5)
+    log_loss_radius = th.as_variable(0.5)
+
+    # First create an objective with individual cost functions per error term
+    xs = [th.Vector(1, name=f"x{i}") for i in range(n)]
+    ts = [th.Vector(1, name=f"t{i}") for i in range(n)]
+    obj_unrolled = th.Objective()
+    for i in range(n):
+        obj_unrolled.add(
+            th.RobustCostFunction(
+                th.Difference(xs[i], ts[i], w, name=f"cf{i}"),
+                th.HuberLoss,
+                log_loss_radius,
+                name=f"rcf{i}",
+            )
+        )
+    lin_unrolled = th.DenseLinearization(obj_unrolled)
+    th_inputs = {f"x{i}": torch.ones(batch_size, 1) * (i + 1) for i in range(n)}
+    obj_unrolled.update(th_inputs)
+    lin_unrolled.linearize()
+
+    # Now one with a single vectorized cost function, and flatten_dims=True
+    xb = th.Vector(n, name="xb")
+    tb = th.Vector(n, name="tb")
+    obj_flattened = th.Objective()
+    obj_flattened.add(
+        th.RobustCostFunction(
+            th.Difference(xb, tb, w, name="cfb"),
+            th.HuberLoss,
+            log_loss_radius,
+            name="rcf",
+            flatten_dims=True,
+        )
+    )
+    lin_flattened = th.DenseLinearization(obj_flattened)
+    th_inputs = {
+        "xb": torch.cat([torch.ones(batch_size, 1) * (i + 1) for i in range(n)], dim=1)
+    }
+    obj_flattened.update(th_inputs)
+    lin_flattened.linearize()
+
+    # Both objectives should result in the same error and linearizations
+    torch.testing.assert_close(obj_unrolled.error(), obj_flattened.error())
+    torch.testing.assert_close(lin_unrolled.b, lin_flattened.b)
+    torch.testing.assert_close(lin_unrolled.AtA, lin_flattened.AtA)
--- a/tests/core/test_robust_cost.py
+++ b/tests/core/test_robust_cost.py
@@ -138,37 +138,75 @@ def test_mask_jacobians(loss_cls):
         torch.testing.assert_close(j1, j2)
 
 
+def _data_model(a, b, x):
+    return a * x.square() + b
+
+
+def _generate_data(num_points=100, a=1, b=0.5, noise_factor=0.01):
+    data_x = torch.rand((1, num_points))
+    noise = torch.randn((1, num_points)) * noise_factor
+    return data_x, _data_model(a, b, data_x) + noise
+
+
 @pytest.mark.parametrize("batch_size", [1, 4])
 def test_flatten_dims(batch_size):
+    # This creates two objectives for a regression problem
+    # and compares their linearization
+    #   - Obj1: N 1d robust costs functions each evaluating one residual term
+    #   - Obj2: 1 Nd robust cost function that evaluates all residuals at once
+    # Data for regression problem y ~ Normal(Ax^2 + B, sigma)
     n = 10
+    data_x, data_y = _generate_data(num_points=n)
+    data_y[:, :5] = 1000  # include some extreme outliers for robust cost
+
+    # optimization variables are of type Vector with 1 degree of freedom (dof)
+    a = th.Vector(1, name="a")
+    b = th.Vector(1, name="b")
+
+    def residual_fn(optim_vars, aux_vars):
+        a, b = optim_vars
+        x, y = aux_vars
+        return y.tensor - (a.tensor * x.tensor.square() + b.tensor)
+
     w = th.ScaleCostWeight(0.5)
     log_loss_radius = th.as_variable(0.5)
 
     # First create an objective with individual cost functions per error term
     xs = [th.Vector(1, name=f"x{i}") for i in range(n)]
-    ts = [th.Vector(1, name=f"t{i}") for i in range(n)]
+    ys = [th.Vector(1, name=f"y{i}") for i in range(n)]
     obj_unrolled = th.Objective()
     for i in range(n):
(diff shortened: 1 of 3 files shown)


theseus/core/robust_cost_function.py
```diff
@@ -80,9 +82,13 @@ def error(self) -> torch.Tensor:
 
     def weighted_error(self) -> torch.Tensor:
         weighted_error = self.cost_function.weighted_error()
+        if self.flatten_dims:
+            weighted_error = weighted_error.reshape(-1, 1)
         squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
         error_loss = self.loss.evaluate(squared_norm, self.log_loss_radius.tensor)
 
+        if self.flatten_dims:
+            return (error_loss.reshape(-1, self.dim()) + RobustCostFunction._EPS).sqrt()
```
fantaosha, line 91: Not sure if something like `(error_loss.reshape(-1, self.dim()).sum(dim=-1, keepdims=True) + RobustCostFunction._EPS).sqrt()` is better and more consistent.

theseus/core/robust_cost_function.py
```diff
@@ -107,15 +113,25 @@ def weighted_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
...
+                weighted_jacobians[i] = wj.diagonal(dim1=1, dim2=2).reshape(-1, 1, 1)
         squared_norm = torch.sum(weighted_error**2, dim=1, keepdim=True)
         rescale = (
             self.loss.linearize(squared_norm, self.log_loss_radius.tensor)
             + RobustCostFunction._EPS
         ).sqrt()
 
-        return [
+        rescaled_jacobians = [
             rescale.view(-1, 1, 1) * jacobian for jacobian in weighted_jacobians
```
fantaosha, line 127: Following my previous comments, this will be something like `torch.einsum("ni, nij->nij", rescale.view(-1, self.dims()), jacobians)`

theseus/core/robust_cost_function.py
```diff
@@ -107,15 +113,25 @@ def weighted_jacobians_error(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
...
-        ], rescale * weighted_error
+        ]
+        rescaled_error = rescale * weighted_error
```
fantaosha, line 129: This line might also need changes as well

fantaosha (commented):
LGTM. Some minor comments on how to deal with rescaling.

theseus/core/robust_cost_function.py
luisenp: Wouldn't this change the result? The test I added checks that the current behavior is the same as `flatten_dims=False` for separate cost functions, which is what we want.

theseus/core/robust_cost_function.py
luisenp: Similar comment as above.

xphsean12: With .diagonal, it presumes that the Jacobian matrix is a square matrix. And for a jacobian, it would be dE/dX, which determines its shape is (dim(E), dim(x)). My question is, will it be necessarily equal for dim(E) and dim(X)?

luisenp: Good observation! I forgot to mention that this implementation assumes that we want to model N robust cost functions of dim=1, each with $k$ variables of dof=1, with a single robust cost function of dimension N and $k$ variables of dof=N. Basically, this assumes you want to use a single `RobustCost` object to model a problem of the form

$$J =  \sum_{i=1:N} \rho(f_i(x_i, y_i, ...)^2)$$

Supporting cost function of dim > 1 and other combinations of variables dof gets messier, and I'm not sure it's worth the effort.

Note that this feature is mostly for programming convenience. Creating individual cost functions objects doesn't hurt performance, because `TheseusLayer` by defaults vectorizes cost functions with the same pattern so they are evaluated in a single call.

xphsean12: But it seems it's not feasible for the case like in tutorial 4, motion model (First figure). In that case, the Error is the combination of 2 vectors, which could be usual and convenient in many case. Therefore, in my original code, I obtain the dimension of the input X through the Jacobian matrix. Because the error itself only contains the batch size and the dimension of the error. And I believe based on the chain rule the dRobustLoss/dOriLoss should be able to be multiplied by the original Jocobian directly.
image
image
image

luisenp: A more appropriate example is the pose graph optimization case where each error term is dim=6. You can see that we model each error term with its [own separate `RobustCost`](https://github.com/facebookresearch/theseus/blob/main/examples/pose_graph/pose_graph_synthetic.py#L161), rather than having one large `RobustCost` that computes all error terms at once. There is no need for a `flatten_dims` argument or anything similar. The loss penalizes only those edges in the graph that might be the result of very noise relative pose measurements, it doesn't matter that the cost has dim > 1.

luisenp: Of course, we can always construct cost functions that are not supported by our implementation. For example, if you combine measurements from two dim > 1 sources in a single cost function, then this wouldn't do what you want. But I would argue that in that case rewriting as separate cost terms would be cleaner and less error prone.

xphsean12: Yes. You are right. Currently, I do not see an optimization problem that would result in where dim(E) != dim(X) or cannot be decoupled into multiple cost functions that fulfill dim(E) = dim(X). We can further discuss it if one encounters those problems. Thank you! I have learned a lot from your project!

luisenp: No problem, and thank you for all your useful feedback!

tests/core/test_robust_cost.py
```diff
@@ -136,3 +136,53 @@ def test_mask_jacobians(loss_cls):
     torch.testing.assert_close(err, err_expected)
     for j1, j2 in zip(jac, jac_expected):
         torch.testing.assert_close(j1, j2)
+
+
+@pytest.mark.parametrize("batch_size", [1, 4])
+def test_flatten_dims(batch_size):
```
mhmukadam, line 142: Assuming we don't need any additional test for flatten_dims to verify compatibility with our internal vectorization?

luisenp: I think it should work fine unless one has a mix of similar cost functions but some have `flatten_dims=True` and others not. Then vectorization code will copy whichever flag value happens to find first, and apply it to all of them. And even in that case, I think the output value should be correct. 

However, this is actually a general issue that I found a while ago and I had forgotten about it. I'll open a Github bug report so I don't forget. But I'm not sure there is an easy work around for this. Basically, if your error depends on python variables not declared as optim or aux vars, and they are not shared across all cost functions, then vectorization might result in incorrect results.

mhmukadam: > error depends on python variables not declared as optim or aux vars
Hmm by definition shouldn't such variables be included in aux vars?

luisenp: If they are `theseus` variables then yes, this is natural. But this is kinda awkward for things like flags. Check out the example in #511. I could see this coming up even with neural nets in some corner cases.

mhmukadam: Ah okay makes sense. I wasn't thinking about flags.

theseus/core/robust_cost_function.py
luisenp: @xphsean12 I have to apologize, because turns out I was wrong in some of the above! We definitely want to support the case were dim(error) != dim(X). For example, in the regression problem, dim(error) = N_points, but dim(X) = 1 both for A and B. In the above I was mixing up optim vars with aux vars. 

Luckily, the fix is still pretty simple, and I have rewritten the test to check for the case above, using the same regression example. You can look at the last few commits if you are curious. 

The code still assumes that `flatten_dims` acts as if it was dim(err) independent costs of dim=1.

tests/core/test_robust_cost.py
luisenp: In the end I added a custom fix for `RobustCostFunction` so there shouldn't be any issues with vectorize either (see 648ba7842d413282f548e36abe9fc5d16d53bab4).

theseus/core/robust_cost_function.py
xphsean12: Thanks for your feedback! By the way, for the radius,  (maybe) I found a new issue. Since I do not have very concrete evidence, I am not going to raise an issue so far. In my previous commit, I change the radius = radius.exp() into .square(). @fantaosha mention that exp() is better for differentiating. However, the easiest approach should be just let radius = radius. And we just need to let the user input an radius >= 0.

The main difference is that (1) based on my observation, .exp() is much slower than radius = radius when I try to train the Huber radius (2) training an value, which will be further pass to the exp() function may be very difficult. Because when changing the trained value from 0 to 5, the actual radius being used is changed from 1 to 148. The exponential growth may cause training an accurate radius difficult. Because the radius will be enlarged by an exponential function. I am not sure whether my second concern is correct, just kind of thinking.

tests/core/test_robust_cost.py
mhmukadam: Great, this is a clean fix and sufficient for now.

mhmukadam (approved):

PR #504, opened 2023-04-28
Add `labs.lie` support for `functorch` and `vmap` pt. 1

luisenp:
Adds `vmap` support for the functional package of `labs.lie`. I have only tested that `vmap + jacrev` works correctly on `inv`, `exp`, `log`, `compose`.

Requires PyTorch >= 2.0, but backward compatibility is maintained. 

TODO:
- [ ] test other operators (maybe? need to figure out what a good test for those is, @fantaosha)
- [x] remove remaining jacobians cache after confirming they are not necessary with `jacrev`
- [x] incorporate `transform_from` fixes in #505

### DIFF
--- a/tests/labs/lie/functional/common.py
+++ b/tests/labs/lie/functional/common.py
@@ -126,3 +126,60 @@ def left_project_func(module, group):
         return module._left_project_autograd_fn(group, matrix[sels, ..., sels, :, :])
 
     return func
+
+
+def check_jacrev_unary(group_fns, dim, batch_size, name):
+    if not hasattr(torch, "vmap"):
+        return
+
+    test_fn = getattr(group_fns, name)
+
+    fn_input = (
+        torch.randn(batch_size, dim) if name == "exp" else group_fns.rand(batch_size)
+    )
+
+    def f(t):
+        return group_fns.log(test_fn(t.unsqueeze(0))).squeeze(0)
+
+    j = torch.vmap(torch.func.jacrev(f))(fn_input)
+    jac_vmap = group_fns.left_project(fn_input, j) if name != "exp" else j
+
+    jlog, jtest = [], []
+    group_fns.log(test_fn(fn_input, jacobians=jtest), jacobians=jlog)
+    jac_analytic = jlog[0] @ jtest[0]
+
+    torch.testing.assert_close(jac_vmap, jac_analytic)
+
+
+def check_jacrev_binary(group_fns, batch_size, name):
+    if not hasattr(torch, "vmap"):
+        return
+
+    test_fn = getattr(group_fns, name)
+
+    fn_inputs = (
+        (group_fns.rand(batch_size), torch.randn(batch_size, 3))
+        if name == "transform_from"
+        else (group_fns.rand(batch_size), group_fns.rand(batch_size))
+    )
+
+    def f(t1, t2):
+        op_out = test_fn(t1.unsqueeze(0), t2.unsqueeze(0))
+        if name == "compose":
+            op_out = group_fns.log(op_out)
+        return op_out.squeeze(0)
+
+    jacs_vmap = []
+    for i in range(2):
+        j = torch.vmap(torch.func.jacrev(f, i))(fn_inputs[0], fn_inputs[1])
+        if fn_inputs[i].ndim == 3:  # group input
+            j = group_fns.left_project(fn_inputs[i], j)
+        jacs_vmap.append(j)
+
+    jlog, jtest = [], []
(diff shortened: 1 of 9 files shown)


tests/labs/lie/functional/common.py
```diff
@@ -126,3 +126,60 @@ def func(matrix: torch.Tensor):
         return module._left_project_autograd_fn(group, matrix[sels, ..., sels, :, :])
 
     return func
+
+
+def check_jacrev_unary(group_fns, dim, batch_size, name):
+    if not hasattr(torch, "vmap"):
```
fantaosha, line 132: So this is for `torch<2.0`?

tests/labs/lie/functional/common.py
```diff
@@ -126,3 +126,60 @@ def func(matrix: torch.Tensor):
...
+        return
+
+    test_fn = getattr(group_fns, name)
+
+    fn_input = (
+        torch.randn(batch_size, dim) if name == "exp" else group_fns.rand(batch_size)
+    )
+
+    def f(t):
+        return group_fns.log(test_fn(t.unsqueeze(0))).squeeze(0)
```
fantaosha, line 142: How do you handle `log()` itself, whose output is on the Euclidean space?

tests/labs/lie/functional/common.py
```diff
@@ -126,3 +126,60 @@ def func(matrix: torch.Tensor):
...
+        (group_fns.rand(batch_size), torch.randn(batch_size, 3))
+        if name == "transform_from"
+        else (group_fns.rand(batch_size), group_fns.rand(batch_size))
+    )
+
+    def f(t1, t2):
+        op_out = test_fn(t1.unsqueeze(0), t2.unsqueeze(0))
+        if name == "compose":
+            op_out = group_fns.log(op_out)
+        return op_out.squeeze(0)
```
fantaosha, line 170: I don't think this applies to `transform_from` whose output is a vector? Maybe we can classify these operators into several categories depending on the types of input and output.

tests/labs/lie/functional/common.py
```diff
@@ -126,3 +126,60 @@ def func(matrix: torch.Tensor):
...
+
+    jacs_vmap = []
+    for i in range(2):
+        j = torch.vmap(torch.func.jacrev(f, i))(fn_inputs[0], fn_inputs[1])
+        if fn_inputs[i].ndim == 3:  # group input
+            j = group_fns.left_project(fn_inputs[i], j)
```
fantaosha, line 176: We don't need `left_project`  for the 2nd argument of `transform_from`.

tests/labs/lie/functional/common.py
```diff
@@ -126,3 +126,60 @@ def func(matrix: torch.Tensor):
         return module._left_project_autograd_fn(group, matrix[sels, ..., sels, :, :])
 
     return func
+
+
+def check_jacrev_unary(group_fns, dim, batch_size, name):
```
fantaosha, line 131: I think we add another two arguments to specify the types (Euclidean/Riemannian) of the input and output. In that case, we could set the tests automatically.

fantaosha (commented):
First version of review about the tests. It might be better to add additional arguments to specify the types (Eucidlean/Riemannian) of inputs and outputs so that we know if we should apply `log()` and `left_project` accordingly.

tests/labs/lie/functional/common.py
luisenp: Yes, so the test doesn't fail for older torch.

tests/labs/lie/functional/common.py
luisenp: Are you talking about the `log` function call? If so, that's why I have an if catching it's only used for `compose`.

tests/labs/lie/functional/common.py
luisenp: The second argument of `transform_from` is `ndim == 2` , so `left_project` is not called in that case.

tests/labs/lie/functional/common.py
luisenp: In the old test code I added [this](https://github.com/facebookresearch/theseus/blob/main/tests/labs/lie/functional/common.py#L13), so the same test code can be used for all operators. I didn't add something like this here because (for now) we are only testing 4 operators.

fantaosha: That makes sense. All the tests look good to me.

> In the old test code I added [this](https://github.com/facebookresearch/theseus/blob/main/tests/labs/lie/functional/common.py#L13), so the same test code can be used for all operators. I didn't add something like this here because (for now) we are only testing 4 operators.

PR #510, opened 2023-05-03
Fixed convergence check error in DCEM with negative error metrics.

luisenp:


bamos (approved):

mhmukadam (approved):
Thanks for the quick fix!

PR #522, opened 2023-05-29
Multi-batch unit tests for Lie groups

luisenp:
Missing unit test for gradient ops

### DIFF
--- a/tests/labs/lie/functional/common.py
+++ b/tests/labs/lie/functional/common.py
@@ -88,7 +88,10 @@ def sample_inputs(input_types, batch_size, dtype, rng):
     return tuple(_sample(type_str) for type_str in input_types)
 
 
-# Run the test for a Lie group operator
+# Run some unit tests for a Lie group operator:
+# checks:
+#   - jacobian of default torch autograd consistent with custom backward implementation
+#   - multi-batch output consistent with single-batch output
 def run_test_op(op_name, batch_size, dtype, rng, dim, data_shape, module):
     is_multi_batch = not isinstance(batch_size, int)
     if is_multi_batch and op_name == "log":
@@ -103,9 +106,6 @@ def run_test_op(op_name, batch_size, dtype, rng, dim, data_shape, module):
             else None
         )
 
-        # checks:
-        #   - analytic backward for the operator
-        #   - multi-batch output consistent with single-batch output
         check_lie_group_function(
             module,
             op_name,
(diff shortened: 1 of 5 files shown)


tests/labs/lie/functional/common.py
```diff
@@ -56,57 +57,81 @@ def get_test_cfg(op_name, dtype, dim, data_shape, module=None):
...
+#   - multi-batch output consistent with single-batch output
 def run_test_op(op_name, batch_size, dtype, rng, dim, data_shape, module):
+    is_multi_batch = not isinstance(batch_size, int)
+    bs = len(batch_size) if is_multi_batch else 1
     all_input_types, atol = get_test_cfg(op_name, dtype, dim, data_shape, module=module)
     for input_types in all_input_types:
         inputs = sample_inputs(input_types, batch_size, dtype, rng)
         funcs = (
-            tuple(left_project_func(module, x) for x in inputs)
+            tuple(left_project_func(module, x, bs) for x in inputs)
```
fantaosha, line 103: Wondering if we need to check `bs=None` or `dim_out=None`, which assumes no broadcasting and is most common user-case.

tests/labs/lie/functional/common.py
```diff
@@ -122,17 +147,52 @@ def check_lie_group_function(module, op_name: str, atol: float, inputs, funcs=No
...
+        # Convert to single-batch-dim sparse gradient format
+        batch_size = matrix.shape[:batch_dim]
+        if batch_dim > 0:
+            d = reduce(lambda x, y: x * y, batch_size)
+            matrix = matrix.reshape(d, -1, d, *group.shape[-2:])
+            sels = range(matrix.shape[0])
+            matrix = matrix[sels, ..., sels, :, :]
+            g = group.reshape(d, *group.shape[-2:])
+        # Compute projected gradient matrix
+        ret = module._left_project_autograd_fn(g, matrix)
```
fantaosha, line 179: Do we assume the same dimension of inputs and outputs?

fantaosha (approved):
LGTM. Only minor comments.

tests/labs/lie/functional/common.py
luisenp: I thought the case `tuple()` covered this for the no batch size (see line 10). Or do you mean something else? 

The tests for different `dim_out` are in #527.

tests/labs/lie/functional/common.py
luisenp: Yes, I think so. There should be no broadcasting in the test where this function is used.

PR #523, opened 2023-05-29
Move `torchlie` out of labs

luisenp:


tests/lie_tests/functional/test_so3.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 import torch
 
-from tests.decorators import run_if_labs
+from tests.theseus_tests.decorators import run_if_labs
```
fantaosha, line 10: Do we still need `run_if_labs` if `torchlie` has been moved out of labs?

examples/lie_labs.py
```diff
@@ -1,7 +1,7 @@
 import torch
```
fantaosha, line 1: Do need rename this file? Maybe `trochlie` instead of `lie_labs`?

fantaosha (approved):
LGTM

tests/lie_tests/functional/test_so3.py
luisenp: We don't need it, but I had missed this at first. I removed them in later PR, [here](https://github.com/facebookresearch/theseus/pull/527/commits).

examples/lie_labs.py
luisenp: Yes, but after we merge all of this stuff, I have to fix the library import to `torchlie` in multiple places, so we can do that then.

PR #525, opened 2023-05-31
Deprecated torch < 2.0.0.

luisenp:
Bumps version to `0.2.0.dev0`

### DIFF
--- a/theseus/_version.py
+++ b/theseus/_version.py
@@ -11,7 +11,7 @@ def get_version() -> str:
             " in future releases.",
             FutureWarning,
         )
-    return "0.2.0.dev0"
+    return "0.2.0.dev0"  # __THESEUS_VERSION__
 
 
 __version__ = get_version()
--- a/theseus/_version.py
+++ b/theseus/_version.py
@@ -3,15 +3,11 @@ import warnings
 from torch import __version__ as _torch_version
 from semantic_version import Version
 
+if Version(_torch_version) < Version("2.0.0"):
+    warnings.warn(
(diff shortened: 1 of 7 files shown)


theseus/_version.py
```diff
@@ -0,0 +1,13 @@
+import warnings
+
+from torch import __version__ as _torch_version
+from semantic_version import Version
+
+if Version(_torch_version) < Version("2.0.0"):
+    warnings.warn(
+        "Using torch < 2.0 is deprecated and support will be discontinued "
```
mhmukadam, line 8: will be discontinued -> is discontinued?

theseus/core/cost_function.py
```diff
@@ -11,7 +11,11 @@
 
 import torch
 import torch.autograd.functional as autogradF
-from functorch import jacrev, vmap
+
+try:
+    from torch.func import jacrev, vmap
+except ModuleNotFoundError:
+    from functorch import jacrev, vmap  # type: ignore
```
mhmukadam, line 18: Is this in case torch < 2.0 is installed and functorch is separate install+import?

mhmukadam (commented):
LGTM

theseus/core/cost_function.py
luisenp: Yes!

theseus/_version.py
luisenp: Changed the message here 103219c2df6db8c924fcc16649faa0e5380b5402.

Issue #521, opened 2023-05-23
Revisit Lie group consistency checks


luisenp:
We have observed some 64-bit SE2 tensors with cosines slightly larger than ones sneaking through after application of retraction maps. We should consider making the Lie group checks stronger to prevent issues like this. Another option would be to normalize all `exp_map` outputs (e.g., [here](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/se2.py#L259)).

mhmukadam:
Is this happening in the new Lie groups too?

mhmukadam:
Flagging unresolved comments from #528 that need to be resolved with this issue.
- need greater eps for the log and exp derivative computation
- We might need to add `so3_derivative_near_zero_eps_floatxx`

luisenp:
Closed by #544

PR #528, opened 2023-06-02
Added configurable global options for `torchlie` and `theseus`

luisenp:


fantaosha (commented):
I'm only concerned with additional constants for the `dexp` and `dlog`. Maybe we can file an issue on this.

lie/lie/functional/se3_impl.py
```diff
@@ -414,7 +414,7 @@ def _jlog_impl_helper(
     ret_lin = tangent_vector[..., :3]
     ret_ang = tangent_vector[..., 3:]
     size = get_tangent_vector_size(tangent_vector)
-    near_zero = theta < constants._SO3_NEAR_ZERO_EPS[tangent_vector.dtype]
+    near_zero = theta < LIE_OPTS.get_eps("so3", "near_zero", tangent_vector.dtype)
```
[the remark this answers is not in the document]
luisenp: Add this when solving #521

lie/lie/options.py
```diff
@@ -0,0 +1,68 @@
...
+class _TorchLieOptions:
+    so3_near_pi_eps_float32: float = 0
+    so3_near_zero_eps_float32: float = 0
+    so3_matrix_eps_float32: float = 0
+    so3_quat_eps_float32: float = 0
+    so3_hat_eps_float32: float = 0
+    se3_hat_eps_float32: float = 0
+    so3_near_pi_eps_float64: float = 0
+    so3_near_zero_eps_float64: float = 0
+    so3_matrix_eps_float64: float = 0
```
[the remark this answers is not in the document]
luisenp: Add this when solving #521

lie/lie/options.py
```diff
@@ -0,0 +1,68 @@
...
+def _CHECK_DTYPE_SUPPORTED(dtype):
+    if dtype not in [torch.float32, torch.float64]:
+        raise ValueError(
+            f"Unsupported data type {dtype}. "
+            "Theseus only supports 32- and 64-bit tensors."
+        )
+
+
+@dataclass
+class _TorchLieOptions:
```
mhmukadam, line 20: Minor: wondering if 'options' is the least ambiguous name for the file/class/functions. It sounds more like selecting between some alternatives for some workflow.

Any reason these need to be outside of `constant.py`? If not, we could just use for instance `class _TorchLieConstants`, `set_global_constants`, etc and host everything in that same file.

theseus/options.py
```diff
@@ -0,0 +1,74 @@
...
+def _CHECK_DTYPE_SUPPORTED(dtype):
+    if dtype not in [torch.float32, torch.float64]:
+        raise ValueError(
+            f"Unsupported data type {dtype}. "
+            "Theseus only supports 32- and 64-bit tensors."
+        )
+
+
+@dataclass
+class _TheseusGlobalOptions:
```
mhmukadam, line 21: Is this separate/duplicate setup for Theseus and torchlie mostly because we don't have SO2/SE2 in the torchlie yet? It would make sense to consolidate later.

mhmukadam (commented):
LGTM

lie/lie/options.py
luisenp: My thinking was that constants should refer to things that we really never want to change, unless need to make an update to the library (e.g., the value of PI, EPS for most tests, etc.), while "options" are things users can change depending on their preferences. 

I guess these Lie group tolerances fall somewhat close to being "constants" according to the definition above, but I figured having flexibility for this wouldn't hurt. And the functionality is now in place for other things.

theseus/options.py
luisenp: No, mainly because these will be two separate repos at some point and I didn't want the Theseus options structure depend on the other one. Ideally I would have made this follow some common interface, but I wasn't sure where to put it that it wouldn't be an issue when moving the repo out.

luisenp: But maybe we should remove the support for doing something like `th.set_global_options({f"torchlie.so3_near_pi_eps_float32": 100.0})`, and just let users do things separately.

lie/lie/options.py
mhmukadam: That makes sense to separate out the true 'constants'. Maybe tolerance / eps / params are clearer terms than options?

theseus/options.py
mhmukadam: What I am worried about is having two say values of eps around in the code. Could we set this up as a wrapper on Theseus side calling into torchlie similar to the functional api? Then when torchlie moves to a standalone repo users still have the possibility of changing eps for the internally imported torchlie via the theseus level function. That way there is only one true value of an eps in any script/program.

mhmukadam: Maybe
```python
class _TheseusGlobalOptions(_TorchLieOptions):
    ...
    se2/so2 eps # remove after se2/so2 added to torchlie
    ....
```

lie/lie/options.py
luisenp: Maybe `params` is better. To be clear, I'm intending for these to be more general than just defining EPS, it's just happens to be the only configuration settings that we have at the moment.

theseus/options.py
luisenp: I'll take a look at what kind of unification we can do that's better than what we have. But the type of subclassing in your last example is something I was trying deliberately to avoid. I don't think we want for  `theseus` configuration settings to be a super set of `torchlie` configuration settings. I think they should remain separate entities.

lie/lie/options.py
mhmukadam: Sounds good. Let's go with `params` then if  it's worth the hassle to make the changes.

theseus/options.py
luisenp: Fixed in #547

PR #538, opened 2023-06-20
Add inverse kinematics example

fantaosha:
## Types of changes

## Checklist


theseus/labs/embodied/robot/forward_kinematics.py
```diff
@@ -123,15 +109,72 @@ def backward(cls, ctx, *grad_outputs):
...
+        jposes: torch.Tensor = backward_helper(poses)
+
+        rets = tuple(poses[id] for id in link_ids)
+        jacs: List[torch.Tensor] = []
+
+        for link_id in link_ids:
+            pose = poses[link_id]
+            jac = jposes.new_zeros(angles.shape[0], 6, robot.dof)
+            sel = robot.links[link_id].ancestor_non_fixed_joint_ids
+            jac[:, :, sel] = SE3.adj(SE3.inv(pose)) @ jposes[:, :, sel]
```
luisenp, line 152: The body of `forward_kinematics_body_jacobian` is almost the same as that of `forward_kinematics_spatial_jacobian`; only one line changes between the two.  

Can we please stop duplicating code? :'(

luisenp (changes requested):
Thanks for working on this! Left some comments. 

Besides addressing this, I think we should include comments to guide people throughout the example. The scripts in this folder are meant for people to learn how to use `theseus`.

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,54 @@
...
+selected_links = ["panda_virtual_ee_link"]
+fk, jfk_b, jfk_s = get_forward_kinematics(robot, selected_links)
+
+print("---------------------------------------------------")
+print("Body Jacobian")
+print("---------------------------------------------------")
+theta = torch.rand(10, robot.dof, dtype=dtype)
+targeted_poses_ee: torch.Tensor = fk(theta)[0]
+theta = torch.zeros_like(theta)
+for iter in range(50):
```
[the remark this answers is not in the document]
luisenp: As discussed offline, let's keep the current version and focus on adding more explanations and references. We can remove the `CostFunction` for now, but keep the code around somewhere, in case we have time to add it and clean it up properly soon.

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,54 @@
...
+urdf_path = os.path.join(os.path.dirname(__file__), URDF_REL_PATH)
+
+robot = Robot.from_urdf_file(urdf_path, dtype)
+selected_links = ["panda_virtual_ee_link"]
+fk, jfk_b, jfk_s = get_forward_kinematics(robot, selected_links)
+
+print("---------------------------------------------------")
+print("Body Jacobian")
+print("---------------------------------------------------")
+theta = torch.rand(10, robot.dof, dtype=dtype)
```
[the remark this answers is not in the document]
fantaosha: Fixed. Thanks!

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,54 @@
...
+targeted_poses_ee: torch.Tensor = fk(theta)[0]
+theta = torch.zeros_like(theta)
+for iter in range(50):
+    jac_b, poses = jfk_b(theta)
+    error = SE3.log(SE3.compose(SE3.inv(poses[-1]), targeted_poses_ee)).view(-1, 6, 1)
```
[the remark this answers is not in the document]
fantaosha: I will add `local()` to `lie.functional`. See #542

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,54 @@
...
+    print(error.norm())
+    if error.norm() < 1e-4:
+        break
+    theta = theta + 0.5 * (jac_b[-1].pinverse() @ error).view(-1, robot.dof)
+
+print("---------------------------------------------------")
+print("Spatial")
+print("---------------------------------------------------")
+theta = torch.rand(10, robot.dof, dtype=dtype)
```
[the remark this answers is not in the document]
fantaosha: Implemented. Thanks!

theseus/labs/embodied/robot/forward_kinematics.py
fantaosha: Implemented.

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,150 @@
...
+# *********************************************
+print("---------------------------------------------------")
+print("Theseus Optimizer")
+print("---------------------------------------------------")
+
+
+# IK can also be solved as an optimization problem:
+#      min \|log(pose^-1 * targeted_pose)\|^2
+# as the following
+def targeted_pose_error(optim_vars, aux_vars):
```
mhmukadam, line 116: Would this be similar to the [Local](https://github.com/facebookresearch/theseus/blob/main/theseus/embodied/misc/local_cost_fn.py#L14)/Difference CostFunction on SE3?

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,150 @@
...
+    (theta,) = optim_vars
+    (targeted_pose,) = aux_vars
+    pose = th.SE3(tensor=fk(theta.tensor)[0])
+    return (pose.inverse().compose(targeted_pose)).log_map()
```
mhmukadam, line 120: we can call `targeted_pose.local(pose)`?

mhmukadam (approved):
LGTM

examples/labs/inverse_kinematics.py
```diff
@@ -0,0 +1,154 @@
...
+urdf_path = os.path.join(os.path.dirname(__file__), URDF_REL_PATH)
+robot = Robot.from_urdf_file(urdf_path, dtype)
+
+# We can get differentiable forward kinematics functions for specific links
+# by using `get_forward_kinematics_fns`. This function creates three differentiable
+# functions for evaluating forward kinematics, body jacobian and spatial jacobian of
+# the selected links, in that order. The return types of these functions are as
+# follows:
+#
+# - fk: return poses of selected links
```
luisenp, line 43: I think we should be more explicit here, this description doesn't help the reader understand the API. 

For example, `fk` returns N + 1 outputs, where N is the number of selected links. The first N correspond to poses (tensors, I think?). And I think the last output is a list containing poses for all links? It would be really useful to include this level of detail.

Same goes for the jacobians.

examples/labs/inverse_kinematics.py
luisenp: Is it `pose.local(targeted_pose)`? In either case, agree that we should call `local()`.

luisenp (changes requested):
Code looks good, but left some minor comments. I think the most important is to use this example to clearly explain the API, because we don't have any documentation for this code.

examples/labs/inverse_kinematics.py
mhmukadam: That's what we have in the `Local` cost function in /embodied/misc. In the longer term we should actually make a unified version of this cost function that can optionally call in to FK and apply local on pose/point/rot of a specific joint.

luisenp: I think the right way to do this would be [this](https://github.com/facebookresearch/theseus/pull/279#discussion_r955357896), which unfortunately we don't have time to implement, but it would be very nice to have.

examples/labs/inverse_kinematics.py
fantaosha: No, `fk` will only return the tuple of selected link poses.

luisenp: Ah, my bad, I missed [this](https://github.com/facebookresearch/theseus/blob/main/theseus/labs/embodied/robot/forward_kinematics.py#L143) slicing for `fk` function. Sorry about that. You should still mention that the return is a tuple of pose tensors, one for each selected link, in the same order as the link names given as input. 

Also, can you reword the jacobians part so it says "returns a tuple where the first element is a list with the body jacobians of selected links, and the second element is a tuple with the corresponding poses."?

PR #547, opened 2023-06-24
Rename GlobalOptions as Params and remove duplication of some of them with `torchlie`

luisenp:


theseus/global_params.py
```diff
@@ -19,57 +21,51 @@ def _CHECK_DTYPE_SUPPORTED(dtype):
...
-        attr_name = f"{ltype}_{attr}_eps_{str(dtype)[6:]}"
-        return getattr(self, attr_name)
+        try:
+            return _LIE_GP.get_eps(ltype, attr, dtype)
+        except AttributeError:
+            _CHECK_DTYPE_SUPPORTED(dtype)
+            attr_name = f"{ltype}_{attr}_eps_{str(dtype)[6:]}"
+            return getattr(self, attr_name)
 
     def reset(self) -> None:
```
mhmukadam, line 43: Should this also call into params reset for torchlie?

mhmukadam (commented):
LGTM! Thanks for adding these updates.

theseus/global_params.py
mhmukadam: Luis via chat:
> in the new versions users are not allowed to modify torch lie params via theseus, so I think reset should not call the other reset

PR #557, opened 2023-06-27
update torchlie readme

mhmukadam:


torchlie/README.md
```diff
@@ -33,7 +35,38 @@
...
+    We currently provide wheels compiled using Python 3.10.
+    For other versions, consider installing from source.
+
+- #### **From source**
+    The simplest way to install torchlie from source is by running the following
+    ```bash
+    git clone https://github.com/facebookresearch/theseus.git && cd theseus/torchlie
+    pip install -e .
+    ```
+    If you are interested in contributing to torchlie, instead install
```
luisenp, line 61: `pip install -e ".[dev]"` doesn't work in this folder, but we can tell them to do 

```bash
cd .. # theseus root folder
pip install -r requirements/dev.txt
```

luisenp (changes requested):
Thanks for adding this! A couple of small comments, but overall LGTM.

torchlie/README.md
mhmukadam: If you run this in the root folder then it would try to install all theseus, torchlie, and trochkin from source right? What if you just want to install torchlie?

mhmukadam: Do you mean
```
git clone https://github.com/facebookresearch/theseus.git && cd theseus/torchlie
pip install -r ../requirements/dev.txt
```

luisenp: Ah yes, that's better. But you sill need `pip install -e .` for the contributing, so  let's do like:

regular source

```
git clone https://github.com/facebookresearch/theseus.git && cd theseus/torchlie
pip install -e .
```
and for contributing, also include
```python
pip install -r ../requirements/dev.txt
pre-commit install
```

Can you also include the pre-commit install command in our main README? I think it would be useful to have it there as well.

luisenp (approved):
LGTM.

Issue #554, opened 2023-06-26
Potential bug in the build script


JingyuQian:
## 🐛 Bug

The build script does not actually use the created theseus conda environment

## Steps to Reproduce

in the `build_wheel.sh`, change the contents of dockerfile from:

```dockerfile
# --- Install torch
ENV CUDA_HOME /usr/local/cuda-cpu
RUN pip install "torch==1.13" --extra-index-url https://download.pytorch.org/whl/cpu
```
to:
```dockerfile
# --- Install torch
ENV CUDA_HOME /usr/local/cuda-cpu
RUN which pip && ls /opt/conda/envs/theseus/bin/pip && pip install "torch==1.13" --extra-index-url https://download.pytorch.org/whl/cpu
```
And run the build script. 

## Expected behavior

There will be 2 additional lines of output:

```
/opt/conda/bin/pip
/opt/conda/envs/theseus/bin/pip
```

Which means the pip used in the building stage is not the one from the "theseus" conda env.

## Additional context

The problem lies in this line:
```dockerfile
RUN source activate theseus
```
It does activate the theseus env, but then Docker creates a layer from this and proceeds to the next command and won't use the theseus env from this layer. The 3.10 actually comes from the conda base env.

### Bit of context
Was trying to build theseus with a lower version of PyTorch and python3.9. The wheel builds successfully but keeps tagging python3.10, which lead me to find this problem.

luisenp:
Hi @JingyuQian, thanks for reporting this! I noticed this behavior a couple weeks ago but didn't look too deeply into what the cause was, this is very helpful! 

Would you be interested in submitting a pull request with a fix?

JingyuQian:
Hi @luisenp , I've submitted a PR to fix this issue. Feel free to verify, thanks!

PR #561, opened 2023-06-27
Update build_wheel.sh to properly use created conda environment

JingyuQian:
## Motivation and Context

This PR tries to fix a bug in `build_wheel.sh` that causes the theseus conda environment not to be properly used during the build stage

* **Original bug report**: #554 
* **Fix**: Updated PATH env so both `pip` and `python3` come from the theseus conda environment.

## How Has This Been Tested

This is a small fix. Use `which pip` and `which python3` during the build stage to verify. These executables should have `/opt/conda/envs/theseus/bin` as a path prefix.

## Types of changes

## Checklist

### DIFF
--- a/build_scripts/build_wheel.sh
+++ b/build_scripts/build_wheel.sh
@@ -124,7 +124,7 @@ for PYTHON_VERSION in 3.10; do
         /bin/bash ~/miniconda.sh -b -p /opt/conda
     ENV PATH \$CONDA_DIR/bin:\$PATH
     RUN conda create --name theseus python=${PYTHON_VERSION}
-    RUN source activate theseus
+    ENV PATH \$CONDA_DIR/envs/theseus/bin:\$PATH
 
     # --- Install torch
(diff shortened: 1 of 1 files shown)


JingyuQian:
It looks like one of the ci tests failed. The local build is OK so I'm not sure why.

mhmukadam:
> It looks like one of the ci tests failed. The local build is OK so I'm not sure why.

Thanks for pointing out the bug and for making a PR with the fix. The CI issue is happening on other PRs as well and @luisenp is looking into it.

luisenp:
The wheel test always fails on external PRs because it's written to tell the build script to use the latest commit as the target to compile, but the commit doesn't exist in our repo (it's part of the fork where the PR comes from). 

Usually we can ignore, but since this is modifying the build wheel script, I'm also going to test this locally.

Thanks for the contribution @JingyuQian!

luisenp (approved):
Works great, thanks for adding this. 

Closes #554

PR #562, opened 2023-06-28
Temporarily fixed a bug for vmap

fantaosha:
## Types of changes

## Checklist

### DIFF
--- a/tests/torchkin_tests/test_forward_kinematics.py
+++ b/tests/torchkin_tests/test_forward_kinematics.py
@@ -58,18 +58,27 @@ def test_jacobian(batch_size: int, dtype: torch.dtype):
     selected_links = ["panda_link2", "panda_link5", "panda_virtual_ee_link"]
     fk, jfk_b, _ = get_forward_kinematics_fns(robot, selected_links)
 
+    def fk_vmap(t):
+        return tuple(pose.squeeze(0) for pose in fk(t.unsqueeze(0)))
+
     rng = torch.Generator()
     rng.manual_seed(0)
     angles = torch.rand(batch_size, robot.dof, generator=rng, dtype=dtype)
 
     jacs_actual, poses = jfk_b(angles)
 
-    sels = range(batch_size)
     jacs_dense = torch.autograd.functional.jacobian(fk, angles, vectorize=True)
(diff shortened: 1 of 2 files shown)


luisenp:
It does seem that there is some bug in `torch` when autograd generates a zero gradient for an operation. I passed a simple repro to @zou3519 to double check this. 

That being said, it seems that the reason we get a zero gradient is because the jacobian autograd functions return poses for selected links + ancestors (see [here](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L26) and [here](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L138)), but only the poses of links with parents [are used](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L63) (all but `poses[0]`, if I understand correctly). 

Is there a reason why we can't make `selected_links` only contain the links whose poses will be used by jacobian computation? (in line 138). As far as I can tell, the first element of [selected_poses](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L146) will never be used (it's not used by the backward_helper and it's not used by the loop in line 155). Is this correct?

fantaosha:
> It does seem that there is some bug in `torch` when autograd generates a zero gradient for an operation. I passed a simple repro to @zou3519 to double check this.
> 
> That being said, it seems that the reason we get a zero gradient is because the jacobian autograd functions return poses for selected links + ancestors (see [here](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L26) and [here](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L138)), but only the poses of links with parents [are used](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L63) (all but `poses[0]`, if I understand correctly).
> 
> Is there a reason why we can't make `selected_links` only contain the links whose poses will be used by jacobian computation? (in line 138). As far as I can tell, the first element of [selected_poses](https://github.com/facebookresearch/theseus/blob/main/torchkin/torchkin/forward_kinematics.py#L146) will never be used (it's not used by the backward_helper and it's not used by the loop in line 155). Is this correct?

luisenp (approved):

PR #570, opened 2023-07-06
Prevent update variables with device or dtype != Objective's

### LINKED ISSUE #569: `Objective.update()` allows passing tensors not consistent with `self.device`
`Objective.update()` allows passing tensors not consistent with `self.device`. If `update()` is used to update all the variables in the objective, NLLS optimizers will generate `delta` on the objective's original device, and then get a runtime error when trying to retract variables that are in a different device.

luisenp:
Closes #569


fantaosha (approved):
LGTM.

PR #573, opened 2023-07-07
Copied semantic_version to third_party.

luisenp:


fantaosha (approved):
LGTM. But please be careful about these CI failures.

PR #574, opened 2023-07-07
Removed semantic version dependency.

luisenp:


tests/theseus_tests/test_misc.py
```diff
@@ -0,0 +1,15 @@
+import pytest
```
mhmukadam, line 1: license header missing

Maybe we can add the flake8-copyright thing to Theseus?

mhmukadam (dismissed):

tests/theseus_tests/test_misc.py
luisenp: Oh yeah, I forgot about that, I'll add it.

mhmukadam (approved):

PR #581, opened 2023-07-17
Yipuz.more robust loss

YipuZhao:
## Motivation and Context

Adding two robust loss implementations to extend the types of inner optimization in theseus:
- graduated non-convexity (GNC) loss, based on GemanMcClure
- Hinge loss 

## How Has This Been Tested

Hinge loss is added to existing robust loss test script;
GNC loss is covered by a new test script adapted upon existing robust loss unit test.

## Types of changes

## Checklist


tests/theseus_tests/core/test_gnc_cost.py
```diff
@@ -0,0 +1,160 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import pytest
+
+import theseus as th
+
+import torch
+from tests.core.common import BATCH_SIZES_TO_TEST
```
luisenp, line 10: This should be `from tests.theseus_tests.core.common import BATCH_SIZES_TO_TEST`.

theseus/core/robust_cost_function.py
```diff
@@ -165,3 +165,65 @@ def _supports_masking(self) -> bool:
     def _supports_masking(self, val: bool):
         self.cost_function._supports_masking = val
         self.__supports_masking__ = val
+
+class GNCRobustCostFunction(RobustCostFunction):
```
luisenp, line 169: Looks like the only difference between this class and base `RobustCostFunction` is that we pass the GNC control parameter to the loss' methods. Is this correct?

If so, it would be preferable to reuse the code of the base class and only change how the loss function is called. For example:

* Add a dictionary member `self.loss_kwargs = {}` in the base class, and add the extra params in GNC class.
* Add an abstract method like `self._evaluate_loss()` and override in GNC class. 

Any thoughts?

theseus/core/robust_cost_function.py
```diff
@@ -165,3 +165,65 @@ def _supports_masking(self) -> bool:
...
+    def __init__(
+        self,
+        cost_function: CostFunction,
+        loss_cls: Type[GNCRobustLoss],
+        log_loss_radius: Variable,
+        gnc_control_val: Variable,
+        name: Optional[str] = None,
+    ):
+        super().__init__(cost_function, loss_cls, log_loss_radius, name=name)
```
luisenp, line 178: We should check that `self.loss` is a subclass of `GNCRobustLoss` and throw `ValueError` with an explanation otherwise.

theseus/core/robust_loss.py
```diff
@@ -50,3 +50,60 @@ def _evaluate_impl(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
...
+        radius: scale of quadratic bowl (also squared)
+        mu: gnc control parameter ranging from 1 (full Geman McClure) to positive infinity (full Quadratic)
+        """
+        return mu * radius * x / (mu * radius + x + _LOSS_EPS)
+
+    @staticmethod
+    def _linearize_impl(
+        x: torch.Tensor, radius: torch.Tensor, mu: torch.Tensor
+    ) -> torch.Tensor:
+        return (mu * radius) ** 2 / ((mu * radius + x) ** 2 + _LOSS_EPS)
```
luisenp, line 109: Missing line at end of file.

tests/theseus_tests/core/test_gnc_cost.py
```diff
@@ -0,0 +1,160 @@
...
+    batch_size = 512
+    rng = torch.Generator()
+    rng.manual_seed(0)
+    gnc_robust_cf = _new_gnc_robust_cf(batch_size, loss_cls, rng, masked_weight=True)
+    jac_expected, err_expected = gnc_robust_cf.weighted_jacobians_error()
+    gnc_robust_cf._supports_masking = True
+    jac, err = gnc_robust_cf.weighted_jacobians_error()
+    torch.testing.assert_close(err, err_expected)
+    for j1, j2 in zip(jac, jac_expected):
+        torch.testing.assert_close(j1, j2)
```
luisenp, line 160: Missing line at end of file.

tests/theseus_tests/core/test_gnc_cost.py
```diff
@@ -0,0 +1,160 @@
...
+# LICENSE file in the root directory of this source tree.
+import pytest
+
+import theseus as th
+
+import torch
+from tests.core.common import BATCH_SIZES_TO_TEST
+
+
+def _new_gnc_robust_cf(
```
luisenp, line 13: Is this test suite the same as the one for the base class? If so, I think it's better to reuse the original test code and just add some control flow that uses `GNCRobustCostFunction` and any extra args when the loss is of type `GNCRobustLoss`.

luisenp (changes requested):
Thanks a lot for this contribution @YipuZhao! Mostly code structure changes for now (haven't checked the  math). 

Can you also add a comment with a reference link to where these losses are described (like a paper or book citation/link).

theseus/core/robust_cost_function.py
YipuZhao: The additional gnc control parameter is used in `loss.evaluate` (weighted_error) and `loss.linearize` (weighted_jacobians_error); are we OK with only overriding these 2 in `GNCRobustCostFunction` class?

luisenp: I like overriding in subclass more than adding a branch in base class.

YipuZhao: Sounds good! I will push a revision in 1 hour.

tests/theseus_tests/core/test_gnc_cost.py
YipuZhao: Yeah we no longer should have a separate test for gnc.

theseus/core/robust_cost_function.py
```diff
@@ -54,9 +55,10 @@ class RobustCostFunction(CostFunction):
     def __init__(
         self,
         cost_function: CostFunction,
-        loss_cls: Type[RobustLoss],
+        loss_cls: Type[RobustLoss | GNCRobustLoss],
```
YipuZhao, line 58: @luisenp the inherent here won't pass pre-commit hook although pytest is OK. 
Problem is we have 2 robust_loss class that don't inherent to one other, while trying to reusing the same `__init__` here between 2 inheriting cost function classes.

YipuZhao:
> Thanks a lot for this contribution @YipuZhao! Mostly code structure changes for now (haven't checked the math).
> 
> Can you also add a comment with a reference link to where these losses are described (like a paper or book citation/link).

Thanks for the review and revised my change. Stuck at precommit hook and curious about your take.

theseus/__init__.py
```diff
@@ -16,11 +16,15 @@
...
     RobustCostFunction,
+    GNCRobustCostFunction,
+    GNCRobustLoss,
+    HingeLoss,
     RobustLoss,
     ScaleCostWeight,
     Variable,
     Vectorize,
     WelschLoss,
+    GemanMcClureLoss,
```
mhmukadam, line 27: Has isort not run here?

theseus/core/robust_cost_function.py
```diff
@@ -165,3 +168,52 @@ def _supports_masking(self) -> bool:
...
+    def __init__(
+        self,
+        cost_function: CostFunction,
+        loss_cls: Type[GNCRobustLoss],
+        log_loss_radius: Variable,
+        gnc_control_val: Variable,
+        flatten_dims: bool = False,
+        name: Optional[str] = None,
+    ):
+        assert issubclass(loss_cls, GNCRobustLoss)
```
mhmukadam, line 186: Probably replace with error?

theseus/core/robust_loss.py
```diff
@@ -50,3 +50,64 @@ def _evaluate_impl(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
...
+    @staticmethod
+    def _evaluate_impl(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
+        return torch.where(x > radius, torch.sqrt(x) - torch.sqrt(radius), _LOSS_EPS)
+
+    @staticmethod
+    def _linearize_impl(x: torch.Tensor, radius: torch.Tensor) -> torch.Tensor:
+        return torch.where(x > radius, 1.0 / (2 * torch.sqrt(x) + _LOSS_EPS), 0.0)
+
+
+class GNCRobustLoss(RobustLoss):
```
mhmukadam, line 65: `class GNCRobustLoss(RobustLoss, abc.ABC)` since this also has abstract methods and the class cannot be instantiated.

theseus/__init__.py
luisenp: Good eye lol. We have exceptions for this file due to errors with internal repos (the `# usort: skip` a few lines above). @YipuZhao alphabetical order would still be nice.

theseus/core/robust_cost_function.py
luisenp: +1

tests/theseus_tests/core/test_robust_cost.py
```diff
@@ -25,44 +29,96 @@ def _new_robust_cf(
...
+
+
 def test_robust_cost_weighted_error():
     generator = torch.Generator()
     generator.manual_seed(0)
     for _ in range(10):
         for batch_size in BATCH_SIZES_TO_TEST:
-            for loss_cls in [th.WelschLoss, th.HuberLoss]:
-                robust_cf = _new_robust_cf(batch_size, loss_cls, generator)
+            for loss_cls in [
```
mhmukadam, line 77: The two for loops could be moved to `@pytest.mark.parametrize`.

mhmukadam (commented):
Thanks for adding this @YipuZhao!

YipuZhao:
> Thanks for adding this @YipuZhao!

Thanks for the comments! Revised and rebased :)

mhmukadam (approved):

luisenp (approved):
LGTM @YipuZhao, thanks so much for the contribution and for iterating on this!

PR #582, opened 2023-07-18
Fix broken `sksparse` install in CI

luisenp:


fantaosha (approved):
LGTM

mhmukadam (commented):
LGTM! I guess we'll switch to the 0.4.11 scikit-sparse available on pypi now.

PR #612, opened 2023-10-11
Modify Objective::_erase_function_variables(...) to also delete varia…

Hesam-Vayu:
…ble name from _all_variables.

Not deleting a variable name in _all_variables that is no longer used causes crashes when you try to add a CostFunction that includes the variable name again.

## Motivation and Context

If you delete enough CostFunctions where the underlying variables are deleted and then add another CostFunction that uses the same variable, the [following check passes](https://github.com/facebookresearch/theseus/blob/9a117fd02867c5007c6686e342630f110e488c65/theseus/core/objective.py#L186), causing [this statement](https://github.com/facebookresearch/theseus/blob/9a117fd02867c5007c6686e342630f110e488c65/theseus/core/objective.py#L206C12-L206C12) to crash due to [this statement](https://github.com/facebookresearch/theseus/blob/9a117fd02867c5007c6686e342630f110e488c65/theseus/core/objective.py#L196) not being executed.

## How Has This Been Tested

Ran tests and optimizations in own code.

## Types of changes

## Checklist

### DIFF
--- a/tests/theseus_tests/core/test_objective.py
+++ b/tests/theseus_tests/core/test_objective.py
@@ -547,3 +547,18 @@ def test_to_dtype():
                 assert var.dtype == dtype
             for aux in cf.aux_vars:
                 assert aux.dtype == dtype
+
+def test_cost_delete_and_add():
+    x = th.Variable(torch.zeros(1), name="x")
+    y = th.Variable(torch.zeros(1), name="y")
+
+    def error_fn(optim_vars, aux_vars):
+        return 0
+
+    objective = th.Objective()
+    cost_function = th.AutoDiffCostFunction([x], error_fn, 1, aux_vars=[y], cost_weight=th.ScaleCostWeight(1.0))
+    
+    # Add a cost function, erase it, and add it again to make sure we don't have any bugs.
+    objective.add(cost_function)
+    objective.erase(cost_function.name)
+    objective.add(cost_function)
--- a/tests/theseus_tests/core/test_objective.py
+++ b/tests/theseus_tests/core/test_objective.py
@@ -549,16 +549,30 @@ def test_to_dtype():
                 assert aux.dtype == dtype
 
 def test_cost_delete_and_add():
-    x = th.Variable(torch.zeros(1), name="x")
-    y = th.Variable(torch.zeros(1), name="y")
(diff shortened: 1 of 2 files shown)


mhmukadam:
Thanks for catching this and the fix. Might be good to update the unit test so this delete behavior also gets checked.

Hesam-Vayu:
Done, I added the test.

tests/theseus_tests/core/test_objective.py
```diff
@@ -547,3 +547,18 @@ def test_to_dtype():
...
+    def error_fn(optim_vars, aux_vars):
+        return 0
+
+    objective = th.Objective()
+    cost_function = th.AutoDiffCostFunction([x], error_fn, 1, aux_vars=[y], cost_weight=th.ScaleCostWeight(1.0))
+    
+    # Add a cost function, erase it, and add it again to make sure we don't have any bugs.
+    objective.add(cost_function)
+    objective.erase(cost_function.name)
+    objective.add(cost_function)
```
mhmukadam, line 564: Thanks for adding the unit test. Might need some asserts to check if it does the right thing?

Any thoughts @luisenp?

Hesam-Vayu: Test failure is caused by throwing an exception. If the last statement doesn't throw an exception, the test passes.

luisenp: We could still check if `objective.optim_vars`, `objective.aux_vars`, and `objective.cost_functions` returns the right thing after adding, erasing and adding again.

luisenp: @Hesam-Vayu let me know if you have any questions about the suggestions above. Thanks!

Hesam-Vayu: I added some checks. I used the 'x.name in optim_vars' pattern as there's no guarantee that other variables will not be created during add cost function operations and enforcing the constraint that only passed in variables should be added in this test would severely restrict the design space of the system.

Hesam-Vayu: @luisenp Any thoughts on the tests?

luisenp (approved):
LGTM! Thanks for adding this fix!

Hesam-Vayu:
My pleasure :)

PR #627, opened 2023-11-17
Faster log map implementation

luisenp:
The current autograd operators in `torchlie.functional` result in duplicated computation for function calls such as 
```python
res = torchlie.functional.SO3.log(g, jacobians=jlist)
```
because these are computed as follows:

```python
jac, _= _jlog_impl(group)
res = _log_autograd_fn(group) # repeats all the computation done in `_op_impl`
return res, jac
``` 

This PR adds a wrapper that reuses the computation, while preserving the custom backward implementation, as follows

```python
jac, res_value = _jlog_impl(group)
res_tensor = _log_passthrough_fn(group, res_value, jac[0])
return res_tensor, jac
```

Benchmarking with [this script](https://github.com/facebookresearch/theseus/blob/lep.faster_log_map/evaluations/time_local_cost_backward.py) shows improvements of 15-23% runtime when backpropagating through local cost functions in Theseus.  I haven't written any dedicated unit tests, but both versions result in the same loss history in the script above, so things look correct.

### DIFF
--- a/theseus/embodied/misc/local_cost_fn.py
+++ b/theseus/embodied/misc/local_cost_fn.py
@@ -40,7 +40,7 @@ class Local(CostFunction):
         return self.target.local(self.var)
 
     def jacobians(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
-        if _THESEUS_GLOBAL_PARAMS.fast_approx_log_maps:
+        if _THESEUS_GLOBAL_PARAMS.fast_approx_local_jacobians:
             if (
                 self._jac_cache is not None
                 and self._jac_cache.shape[0] == self.var.shape[0]
--- a/theseus/embodied/misc/local_cost_fn.py
+++ b/theseus/embodied/misc/local_cost_fn.py
@@ -9,6 +9,7 @@ import torch
 
 from theseus import CostFunction, CostWeight
 from theseus.geometry import LieGroup
+from theseus.global_params import _THESEUS_GLOBAL_PARAMS
 
 
 class Local(CostFunction):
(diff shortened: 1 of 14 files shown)


torchlie/torchlie/functional/so3_impl.py
```diff
@@ -355,15 +355,41 @@ def backward(cls, ctx, grad_output):
 _jexp_autograd_fn = _jexp_impl
 
 
+_UPPER_IDX_3x3_CUDA: Dict[str, torch.Tensor] = None
+if torch.cuda.is_available():
+    _UPPER_IDX_3x3_CUDA = {
```
fantaosha, line 360: Maybe we can create both `_UPPER_IDX_3x3_CUDA` and `_LOWER_IDX_3x3_CUDA`?

theseus/embodied/misc/local_cost_fn.py
```diff
@@ -33,13 +34,31 @@ def __init__(
...
+        self._jac_cache: torch.Tensor = None
+
     def error(self) -> torch.Tensor:
         return self.target.local(self.var)
 
     def jacobians(self) -> Tuple[List[torch.Tensor], torch.Tensor]:
-        Jlist: List[torch.Tensor] = []
-        error = self.target.local(self.var, jacobians=Jlist)
-        return [Jlist[1]], error
+        if _THESEUS_GLOBAL_PARAMS.fast_approx_log_maps:
```
fantaosha, line 43: If I remember correctly, is this the version used in MiniSAM?

fantaosha: FYI, this one seems reasonable only if the `target` and `var` are close to each other. Not sure if we should implement it here.

fantaosha (approved):
Overall LGTM. Only some minor comments.

fantaosha (approved):
LGTM.

PR #628, opened 2023-11-20
Improves to Timer and Profiler utils

luisenp:
The timer now has the option of labeling timing runs, and collect history of times per label. Can also be called with `start()` and `end()` methods, without using as a context manager.

The profiler has now methods to to compile the statistics and dump results to a file.

### DIFF
--- a/theseus/utils/utils.py
+++ b/theseus/utils/utils.py
@@ -6,7 +6,8 @@ import cProfile
 import io
 import pstats
 import time
-from typing import Any, Callable, List, Optional, Type
+from collections import defaultdict
+from typing import Any, Callable, Dict, List, Optional, Type
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


theseus/utils/utils.py
```diff
@@ -200,27 +201,67 @@ def autograd_fn(*optim_var_tensors):
...
             self.elapsed_time = (time.perf_counter_ns() - self._start_time) / 1e9
+        if self._caller is not None:
+            self._stats[self._caller].append(self.elapsed_time)
+
+    def __call__(self, caller: Optional[str] = None) -> "Timer":
+        self.start(caller=caller)
+        return self
+
+    def __enter__(self) -> "Timer":
+        return self
```
mhmukadam, line 258: Should this be `self.start()`? otherwise does it the break the older behavior?

mhmukadam (approved):

theseus/utils/utils.py
luisenp: Yeah, I fixed it in [b80f791](https://github.com/facebookresearch/theseus/pull/628/commits/b80f7915584d333cfef9ba306ac14c2b80956809). Good eye!

mhmukadam: Thanks for adding the unit tests as well!

mhmukadam (approved):

PR #629, opened 2023-11-30
Faster log map implementation

luisenp:
Copy of #627, which was incorrectly merged to another branch. This PR sets the target directly to main. No other changes.


fantaosha (approved):
LGTM

PR #631, opened 2023-12-08
Fix broken backward graph when using `LieTensor.exp()`

### LINKED ISSUE #626: Optimization on SE(3)
## ❓ Questions and Help

Thank you for reading and sorry if this is a trivial question.

As an example, I am trying to find a homogenous transform such that `p_prime = HT @ p`. Please find below a script

```python
import theseus as th
import torch
import torchlie as lie

# a translation target transform
Th_target = lie.SE3.exp(torch.tensor([0.4, 0.5, 0.6, 0.0, 0.0, 0.0]))

# some point in R^3
p = torch.tensor([1.0, 0.0, 0.0]).unsqueeze(0)
p_prime = Th_target @ p

# setup optimization variables
p_var = th.Variable(p, name="p")
p_prime_var = th.Variable(p_prime, name="p_prime")
Th_vec_var = th.Vector(dof=6, name="Th_vec")


# specify cost function
def cost_fn(optim_vars, aux_vars):
    Th_vec = optim_vars[0]
    p, p_prime = aux_vars
    Th = lie.SE3.exp(Th_vec.tensor)

    p_prime_hat = Th @ p.tensor
    err = p_prime_hat - p_prime.tensor
    return err


# setup objective
objective = th.Objective()
cost_fn = th.AutoDiffCostFunction(
    optim_vars=[Th_vec_var],
    err_fn=cost_fn,
    dim=3,
    aux_vars=[p_var, p_prime_var],
    cost_weight=th.ScaleCostWeight(1.0),
    name="cost_fn",
)
objective.add(cost_fn)
optimizer = th.LevenbergMarquardt(objective, max_iteration=15, step_size=1.0)

layer = th.TheseusLayer(optimizer=optimizer)

with torch.no_grad():
    output, info = layer.forward(
        input_tensors={"Th_vec": torch.zeros(6).unsqueeze(0)},  # zeros as initial guess
        optimizer_kwargs={"track_best_solution": True, "verbose": True},
    )

print("best solution: ", info.best_solution)
print("target solution: ", Th_target.log())
print(output)
```

Somehow, the program doesn't find an optimal solution here. I believe, the Jacobians are zero, so maybe the `cost_fn` doesn't propagate the gradient. 

Any idea what goes wrong?

Thank you so much for the help!

luisenp:
Closes #626

### DIFF
--- a/tests/torchlie_tests/test_lie_tensor.py
+++ b/tests/torchlie_tests/test_lie_tensor.py
@@ -13,7 +13,7 @@ from .functional.common import get_test_cfg, sample_inputs
 
 @pytest.fixture
 def rng():
-    rng_ = torch.Generator()
+    rng_ = torch.Generator(device="cuda:0" if torch.cuda.is_available() else "cpu")
     rng_.manual_seed(0)
     return rng_
 
--- a/tests/torchlie_tests/test_lie_tensor.py
+++ b/tests/torchlie_tests/test_lie_tensor.py
@@ -124,14 +124,25 @@ def test_op(op_name, ltype_str, batch_size, rng):
 
 
 def test_backward_works():
-    # Run optimization to check that the compute graph is not broken
+    # Runs optimization to check that the compute graph is not broken
+    def _check(opt_tensor, target_tensor, tensor_fn):
+        opt = torch.optim.Adam([opt_tensor])
+        losses = []
+        for i in range(2):
+            opt.zero_grad()
+            d = tensor_fn(opt_tensor).local(target_tensor)
+            loss = torch.sum(d**2)
+            losses.append(loss.detach().clone())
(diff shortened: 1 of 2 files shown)


fantaosha (approved):
LGTM. However, I did not follow why the backward graph was broken.

luisenp:
@fantaosha The reason is that the standard `LieTensor` constructor is not a differentiable operation. If you want to make a new LieTensor in a differentiable way (e.g., as the result of an exp map), you need a tensor creation function that has a backward implementation. This is already available in [`FromTensor`](https://github.com/facebookresearch/theseus/blob/main/torchlie/torchlie/lie_tensor.py#L465) class, I just forgot to use it inside the exp map function.

PR #661, opened 2024-09-11
Add EPS for sqrt in SO3 to_quaternion

### LINKED ISSUE #654: Memory explosion problem
## ❓ Questions and Help
Hi, I'm using Theseus to do pose graph optimization, and I I encountered two problems:
1. Every time I try to run optimization, system memory usage sharply exceeds 50 Gb, although all the variables I pass into the objective only takes a few GB of memory
2. I want to have a look at the error_metric every time I add a new edge into the pose graph, and I notice that all error_metric become 'nan' since the last error_metric exceeds 1000 
Could you please give me some hints what can cause these problems?

luisenp:
Closes #654

### DIFF
--- a/theseus/_version.py
+++ b/theseus/_version.py
@@ -33,4 +33,4 @@ if lt_version(torch.__version__, "2.0.0"):
         FutureWarning,
     )
 
-__version__ = "0.2.2.dev0"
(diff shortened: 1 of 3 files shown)


fantaosha (approved):
LGTM.

Issue #486, opened 2023-03-20
Lie group checks should be silent for internal operations


luisenp:
When the outcome of something like `g1.compose(g2)` needs to be normalized, the current code throws a warning. However, this is overly informative for internal ops, and might result in an excessive/intrusive number of warnings. Better to keep these warnings only for when normalizing a user-given tensor at construction time.

jacoblubecki:
I'd like to take a look at this, but from some cursory code-searching, it isn't immediately clear where the warning is occurring. Could you point me in the right direction and I can throw something together?

luisenp:
Hi @jacoblubecki. Thanks for the offer to help!

These checks happen [here](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/manifold.py#L116). Which is called in the base `Manifold` class [here](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/manifold.py#L51). 

The first option that comes to mind is to extend the [`_LieGroupCheckContext`](https://github.com/facebookresearch/theseus/blob/main/theseus/geometry/lie_group_check.py#L10) class to include something like `silent_normalization`, which can then be passed in the call to `Manifold._check_tensor()`. Then in [`TheseusLayer.forward()`](https://github.com/facebookresearch/theseus/blob/main/theseus/theseus_layer.py#L170) you can disable/enable the silent option on enter/exit, w/o changing what the current value of `_LieGroupCheckContext.check_lie_group`  is. 

Happy to answer any questions.

jacoblubecki:
Thank you, this is perfect. I'll start taking a look! In the meantime, can you assign the ticket to me?

luisenp:
Assigned. Thanks!

jacoblubecki:
I created a draft pull request -- the `CONTRIBUTING.md` instructions say to add myself as the assignee and to assign some specific reviewers, but it looks like I don't have permission to set assignees or reviewers on the draft PR...

PR #669, opened 2024-11-09
Add _LieGroupCheckContext setting for disabling internal warnings

jacoblubecki:
## Motivation and Context

Normalization in Lie group checks can be overly verbose in logging, it was identified that it could be an improvement if there was a way to selectively disable these checks in certain contexts (e.g. when implicit normalization is desired).

https://github.com/facebookresearch/theseus/issues/486

## How Has This Been Tested

Existing unit tests + manual testing in python REPL.

Example REPL test:
```
>>> non_se3 = torch.Tensor([[100, 0, 0, 0], [100, 100, 0, 0], [0, 0, 100, 0]]).unsqueeze(0)
>>> with enable_checks(), set_lie_group_check_enabled(True):
        th.SE3(tensor=non_se3)

~/Projects/theseus/theseus/theseus/geometry/manifold.py:141: UserWarning: The input tensor is not valid for SE3 and has been normalized.
  warnings.warn(
SE3(tensor=tensor([[[ 0.8944, -0.4472,  0.0000,  0.0000],
         [ 0.4472,  0.8944,  0.0000,  0.0000],
         [ 0.0000,  0.0000,  1.0000,  0.0000]]]), name=SE3__1)
>>> with enable_checks(), set_lie_group_check_enabled(True, silence_internal_warnings=True):
        th.SE3(tensor=non_se3)

SE3(tensor=tensor([[[ 0.8944, -0.4472,  0.0000,  0.0000],
         [ 0.4472,  0.8944,  0.0000,  0.0000],
         [ 0.0000,  0.0000,  1.0000,  0.0000]]]), name=SE3__2)
```

## Types of changes

## Checklist

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes (no existing tests for this specific context manager, and code is covered in other existing tests)
- [x] All new and existing tests passed (***except CUDA tests which I could not run locally)


theseus/geometry/lie_group_check.py
```diff
@@ -15,51 +15,68 @@ def get_context(cls):
...
-        self.prev = super().get_context()
-        _LieGroupCheckContext.set_context(False, self._silent)
-        return self
-
-    def __exit__(self, typ, value, traceback):
-        _LieGroupCheckContext.set_context(*self.prev)
+        cls.contexts.silence_internal_warnings = silence_internal_warnings
+
+
+@contextmanager
```
jacoblubecki, line 39: I thought using the `@contextmanager` decorator might help simplify the code a little bit -- I believe the functionality should be the same since no internal code was using any of the subclasses directly, and since this API is internal, downstream code shouldn't break, but if you'd prefer I leave the class-based approach, I am happy to revert back to that.

luisenp:
Hi @jacoblubecki, thanks for adding this! Sorry for the delay, I had this in my TODO stack but somehow two weeks passed!

I haven't tested the code yet, but the code and the example lgtm. What we want now is to turn on this new option under the hood when running an optimizer, so that intermediate Lie group operations don't throw this warning if there are numerical precision errors that are not necessarily the users' fault. The best place to do this should be [this line](https://github.com/facebookresearch/theseus/blob/main/theseus/optimizer/optimizer.py#L51). Let me know if you have any questions. x

jacoblubecki:
Added in the location specified, let me know if the code looks okay!

luisenp (approved):
Sorry for the delay. LGMT! Thanks for adding this feature :)

